Issue #6, opened 2016-04-05
img-uses-alt does not allow empty string or possible empty string


lencioni:
This rule considers the following JSX bad:

```jsx
function Foo() {
  return <img alt={foo || ''} />;
}
```

this is also considered bad:

```jsx
function Foo() {
  return <img alt="" />;
}
```

but this is okay:

```jsx
function Foo() {
  return <img alt={foo} />;
}
```

However, I think that all three should be okay. Empty strings can be used appropriately for alt text on images that are decorative.

evcohen:
I agree on the first one, since we can't determine the value of foo until runtime. However, not sure about the second - You can place a space in between the quotation marks (`<img alt=" " />`) and it should pass and still semantically represent the same thing to a screen reader. Also, I can implement case where `alt=""` (or any other form of undefined value) passes lint rule if role=presentation is present.

lencioni:
I think the `role=presentation` bit makes sense. According to the spec, it should probably enforce `alt=""` if it has a presentation role.

> Authors SHOULD NOT provide meaningful alternative text (for example, use alt="" in HTML4) when the presentation role is applied to an image.

https://www.w3.org/TR/wai-aria/roles#presentation

evcohen:
Fixing first example in 0.5.3 - will upgrade minor on `role=presentation` enhancement. 0.5.3 should be done within the hour.

lencioni:
Awesome! Thanks again!

evcohen:
0.5.3 published - should fix first use case + other bugs that are closed!

lencioni:
I think this is still broken for cases like

```jsx
function Foo() {
  return <img alt={foo.bar || ''} />;
}
```

and

```jsx
function Foo() {
  return <img alt={bar() || ''} />;
}
```

and

```jsx
function Foo() {
  return <img alt={foo.bar() || ''} />;
}
```

evcohen:
Added test cases for those and fixed in v0.5.4 - still may be other edge cases, working on resolving cases to handle each type specified in [spec](https://github.com/facebook/jsx/blob/master/AST.md)

lencioni:
Wonderful!

lencioni:
@evcohen do you have an ETA on the `role="presentation"` change? No rush--I'm just wondering if I should roll with `alt=" "` or wait it out.

evcohen:
[...]

PR #11, opened 2016-04-06
Improve img-uses-alt error messages

lencioni:
This rule was giving the same error message for two different types of
errors. This could lead to confusion, so I decided to make the error
messages more descriptive.

When #6 is resolved, the empty alt value message should probably be
expanded to provide guidance about using `role="presentation"`.


evcohen:
👍 awesome - working on `role="presentation"` now.

PR #22, opened 2016-04-08
Allow alt="" or role="presentation"

lencioni:
As we discussed on #6, it is okay to use `alt=""` without
`role="presentation"`. It is possible that we want to enforce no alt
prop if `role="presentation"` but I'll leave that for another time and
possibly a different rule.

### DIFF
--- a/docs/rules/img-uses-alt.md
+++ b/docs/rules/img-uses-alt.md
@@ -41,7 +41,7 @@ To tell this plugin to also check your `Image` element, specify this in your `.e
 }
 ```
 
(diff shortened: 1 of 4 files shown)


evcohen:
Awesome, thanks! I think it makes more sense to return the **actual** attribute value when it's a literal, since `""` isn't a truthy value, anyway.

@lencioni Can you merge latest into your branch?

lencioni:
Rebased!

evcohen:
I'm going to merge and patch - nicer to hammer down rule details and not continually bump minors before v1.0.0 since npm doesn't play well with ^0.y.z

PR #24, opened 2016-04-12
Remove TODO

lencioni:
We have done all of the items on this list, and I think we should be
using issues for this type of thing anyway.


evcohen:
Can you rebase? Will merge right after 😄

lencioni:
Rebased!

Issue #17, opened 2016-04-07
Rule proposal: AX_ARIA_10


lencioni:
https://github.com/GoogleChrome/accessibility-developer-tools/wiki/Audit-Rules#ax_aria_10

This element has an unsupported ARIA attribute

Many ARIA attributes (states and properties) can only be used on elements with particular roles.

For example `aria-posinset` is only supported by `listitem`, `option`, `menuitemradio`, `radio` and `treeitem`.

Similar to #12

evcohen:
Built out [AX_ARIA_11](https://github.com/GoogleChrome/accessibility-developer-tools/wiki/Audit-Rules#ax_aria_11) today. I like all of these proposals. I'm thinking about maybe implementing all of them within the week and publishing v1.0.0

evcohen:
Does this rule mean that role has to be defined for these aria-* properties to be valid? I.E. can you use `aria-posinset` without role being defined?

```jsx
<li aria-posinset="16" />
```

lencioni:
My hunch was that your example should not warn, but I found conflicting answers on stack overflow:

- http://stackoverflow.com/a/11131648/18986
- http://stackoverflow.com/a/11355793/18986

Based on [this information](https://www.w3.org/TR/aria-in-html/#aria-adds-nothing-to-default-semantics-of-most-html-elements), I'd say that `li` has a default role and this rule should not warn. 

Furthermore, it looks like the code for the rule I linked in my OP, [calls a `getRoles` utility method](https://github.com/GoogleChrome/accessibility-developer-tools/blob/e70dd633a25e8a8c659a587f7ce58b923db1eef2/src/audits/UnsupportedAriaAttribute.js#L45) with an argument that [allows implicit semantics to be considered if there is no role attribute](https://github.com/GoogleChrome/accessibility-developer-tools/blob/e70dd633a25e8a8c659a587f7ce58b923db1eef2/src/js/AccessibilityUtils.js#L618-L653).

This seems like enough evidence to me to allow implicit semantics in the eslint rule.

evcohen:
Ah, great. Google's implicit role definition is [here](https://github.com/GoogleChrome/accessibility-developer-tools/blob/master/src/js/Constants.js#L1087). 

So the logic goes

1. Arrive at aria-* attribute node
2. Check the element that contains this attribute for an explicit role, then implicit role.
    - If role is defined, enforce rule accordingly.
[...]

PR #25, opened 2016-04-12
[breaking] - Implement multiple rules and refactor getAttributeValue

### LINKED ISSUE #12: Rule proposal: AX_ARIA_03
I think that this would make a really good linter rule. https://github.com/GoogleChrome/accessibility-developer-tools/wiki/Audit-Rules#ax_aria_03

Elements with ARIA roles must have all required attributes for that role

Good:

```html
<!-- Good: the checkbox role requires the aria-checked state -->
<span role="checkbox" aria-checked="false" aria-labelledby="foo" tabindex="0"></span>
```

Bad:

```html
<!-- Bad: the checkbox role requires the aria-checked state -->
<span role="checkbox" aria-labelledby="foo" tabindex="0"></span>
```

The [WAI-ARIA spec](http://www.w3.org/TR/wai-aria/roles#requiredState) has more information on "Required States and Properties"

### LINKED ISSUE #13: Rule proposal: AX_ARIA_04
I think that this would make a really good linter rule. https://github.com/GoogleChrome/accessibility-developer-tools/wiki/Audit-Rules#ax_aria_04

ARIA state and property values must be valid

The values of ARIA states and properties must be of the correct type.

Bad:

```html
<!-- Bad: the aria-hidden state is of type true/false -->
<span aria-hidden="yes">foo</span>
```

Good:

```html
<!-- Good: the aria-hidden state is of type true/false -->
<span aria-hidden="true">foo</span>
```

### LINKED ISSUE #14: Rule proposal: AX_ARIA_12
I think that this would make a really good linter rule. https://github.com/GoogleChrome/accessibility-developer-tools/wiki/Audit-Rules#ax_aria_12

This element does not support ARIA roles, states and properties

Some HTML elements should not be given ARIA attributes. This is often because they are not visible, for example meta, html, script, style.

Bad:

```html
<!-- Bad: the meta element should not be given any ARIA attributes -->
<meta charset="UTF-8" aria-hidden="false">
```

Good:

```html
<!-- Good: the meta element should not be given any ARIA attributes -->
<meta charset="UTF-8">
```

evcohen:
@lencioni I know this is a huge PR, but all tests are passing and I want to get this code into master so further PR's can be more explicit and modular.

I wanted to refactor `getAttributeValue` so that the extractor functions for different AST attribute value types can be composed together for different use cases (i.e. `getAttributeLiteralValue` will not try and infer values from types that are variables, functions, and other expressions). Those `switch` statements were ugly and harder to debug.

Fixes #12 
Fixes #13 
Fixes #14 
Lays groundwork for #17


src/index.js
```diff
@@ -10,7 +10,8 @@ module.exports = {
     'no-access-key': require('./rules/no-access-key'),
     'label-uses-for': require('./rules/label-uses-for'),
     'no-hash-href': require('./rules/no-hash-href'),
-    'valid-aria-role': require('./rules/valid-aria-role')
+    'valid-aria-role': require('./rules/valid-aria-role'),
+    'no-invalid-aria': require('./rules/no-invalid-aria')
```
lencioni: FWIW, this diff would be a little cleaner if you used a trailing comma ;)

src/rules/no-invalid-aria.js
```diff
@@ -0,0 +1,38 @@
...
+ */
+'use strict';
+
+// ----------------------------------------------------------------------------
+// Rule Definition
+// ----------------------------------------------------------------------------
+
+import validAriaProperties from '../util/validAriaProperties';
+
+const errorMessage = name => `${name}: This attribute is an invalid ARIA attribute.`;
```
lencioni: As a possible future enhancement, it would be really sweet to suggest attributes that are similar to what they had. I imagine this would really help with typos.

src/util/validAriaProperties.js
```diff
@@ -0,0 +1,22 @@
+// Taken from https://www.w3.org/TR/wai-aria/states_and_properties#state_prop_def
```
lencioni: Thanks for including comments like this. Links to more information are usually pretty helpful.

src/util/validAriaProperties.js
```diff
@@ -0,0 +1,22 @@
...
+  'ARIA-LABELLEDBY', 'ARIA-LEVEL',
+  'ARIA-LIVE', 'ARIA-MULTILINE',
+  'ARIA-MULTISELECTABLE', 'ARIA-ORIENTATION',
+  'ARIA-OWNS', 'ARIA-POSINSET',
+  'ARIA-PRESSED', 'ARIA-READONLY',
+  'ARIA-RELEVANT', 'ARIA-REQUIRED',
+  'ARIA-SELECTED', 'ARIA-SETSIZE',
+  'ARIA-SORT', 'ARIA-VALUEMAX',
+  'ARIA-VALUEMIN', 'ARIA-VALUENOW',
+  'ARIA-VALUETEXT'
```
lencioni: nit: I would put each value on its own line. That would make finding something you are looking for a bit easier, and make future modifications to this list a bit cleaner.

evcohen: This gets refactored into object on a later commit. Sorry this PR is a mammoth and hard to follow 🙈 Feel free to reach out if you have any question and checkout the branch and take a look at the code minus the diff.

tests/src/rules/no-invalid-aria.js
```diff
@@ -0,0 +1,55 @@
...
+  parserOptions
+}));
+
+ruleTester.run('no-invalid-aria', rule, {
+  valid: [
+    // Variables should pass, as we are only testing literals.
+    { code: '<div />', parserOptions },
+    { code: '<div></div>', parserOptions },
+    { code: '<div aria="wee"></div>', parserOptions }, // Needs aria-*
+    { code: '<div abcARIAdef="true"></div>', parserOptions },
```
lencioni: Should probably include one that has "aria-" but doesn't start with "aria-", like `fooaria-foo`.

src/util/values/expressions/TemplateLiteral.js
```diff
@@ -1,5 +1,30 @@
-import buildTemplateLiteral from '../../buildTemplateLiteral';
+/**
+ * Returns the string value of a template literal object.
+ * Tries to build it as best as it can based on the passed
+ * prop. For instance `This is a ${prop}` will return 'This is a {prop}'.
+ *
+ * If the template literal builds to undefined (`${undefined}`), then
+ * this should return "".
+ */
+const extract = value => {
```
lencioni: You don't really get much out of the arrow functions here. I think this would be better as `export default function extract(value) {`

lencioni: Also, ideally these function names match their filenames for maximum readability. I suggest also renaming the filename to be something more like `extractTemplateLiteral`.

src/util/values/expressions/UnaryExpression.js
```diff
@@ -0,0 +1,23 @@
+import getValue from './index';
+
+const extract = value => {
+  const { operator, argument } = value;
+
+  switch (operator) {
+    case '-':
+      return -getValue(argument);
+    case '+':
+      return Number(getValue(argument));
```
lencioni: Any reason why you chose `Number(getValue(argument))` over `+getValue(argument)`?

src/util/values/Literal.js
```diff
@@ -1,3 +1,13 @@
-const extract = value => value.value;
+const extract = value => {
+  const { value: extractedValue } = value;
+
+  if (extractedValue === "true") {
```
lencioni: nit: single quotes

src/util/values/index.js
```diff
@@ -1,15 +1,21 @@
...
   JSXExpressionContainer
 };
 
-const getValue = value => TYPES[value.type](value);
+const LITERAL_TYPES = Object.assign({}, TYPES, {
+  JSXElement: () => null,
+  JSXExpressionContainer: extractLiteral
+});
+
+export const getValue = value => TYPES[value.type](value);
```
lencioni: Similar to my comment on a different commit, I think these would be better off without the arrow functions. In this case `export default function getValue(value) {`

src/rules/valid-aria-proptypes.js
```diff
@@ -0,0 +1,75 @@
...
+// ----------------------------------------------------------------------------
+// Rule Definition
+// ----------------------------------------------------------------------------
+
+import ariaAttributes from '../util/ariaAttributes';
+import { getLiteralAttributeValue } from '../util/getAttributeValue';
+
+const errorMessage = (name, type) => `${name} must be of type ${type}.`;
+
+const validityCheck = (value, expectedType, tokens) => {
```
lencioni: Since this returns a boolean, it would be nice if it were named something like `isValid()`.

src/util/values/Literal.js
```diff
@@ -0,0 +1,13 @@
...
```
lencioni: nit: singles

src/util/values/Literal.js
```diff
@@ -0,0 +1,13 @@
...
+    return true;
+  } else if (extractedValue === "false") {
```
lencioni: I don't have a lot of context here--does this distinguish between `true`/`false` and `'true'`/`'false'`?

tests/src/rules/valid-aria-proptypes.js
```diff
@@ -25,15 +25,33 @@ const parserOptions = {
...
+    case 'tokenlist':
+      return `The value for ${name} must be a list of one or more tokens from the following: ${permittedValues}.`;
+    case 'boolean':
+    case 'string':
+    case 'integer':
+    case 'number':
+    default:
+      return `The value for ${name} must be a ${type}.`;
+  }
+};
```
lencioni: Maybe the rule should export this instead of repeating it in the tests?

src/util/values/Literal.js
evcohen: Yes for the use case `<div aria-hidden="true" />` we actually want that to resolve to a boolean.

src/rules/valid-aria-role.js
```diff
@@ -9,50 +9,36 @@
...
+    // If value is undefined, then the role attribute will be dropped in the DOM.
+    // If value is null, then getLiteralAttributeValue is telling us that the value isn't in the form of a literal.
+    if (value === undefined || value === null) {
       return;
     }
 
-    // If value is a literal.
-    const normalizedValues = attribute.value.value.toUpperCase().split(" ");
-    const isValid = normalizedValues.every(value => validRoleTypes.indexOf(value) > -1);
+    const normalizedValues = `${value}`.toUpperCase().split(" ");
```
lencioni: nit: single quotes

Also, what do you think about using `String(value)` instead of the template literal?

src/util/values/expressions/index.js
```diff
@@ -24,6 +24,13 @@ const TYPES = {
 const noop = () => null;
 
 const LITERAL_TYPES = assign({}, TYPES, {
+  Literal: value => {
+    const extractedVal = TYPES.Literal(value);
```
lencioni: I think I find this a little opaque. Perhaps it would be better if `TYPES.Literal` was a verb?

tests/src/rules/valid-aria-proptypes.js
evcohen: I was thinking about extracting all error messages into a single file that just exports an object where key is some unique error code/identifier and value is function/string. This way we only have to change in one place.

src/rules/role-requires-aria.js
```diff
@@ -0,0 +1,57 @@
...
+// Rule Definition
+// ----------------------------------------------------------------------------
+
+import validRoleTypes from '../util/attributes/role';
+import { getLiteralAttributeValue } from '../util/getAttributeValue';
+import hasAttribute from '../util/hasAttribute';
+
+const errorMessage = (role, requiredProps) =>
+  `Elements with the ARIA role "${role}" must have the following ` +
+  `attributes defined: ${requiredProps.toString().toLowerCase()}`;
```
lencioni: `String(requiredProps)` is preferred to `requiredProps.toString()` in case `requiredProps` gets a `toString` property on it for some reason.

tests/src/rules/role-requires-aria.js
```diff
@@ -0,0 +1,97 @@
...
+    { code: '<div />', parserOptions },
+    { code: '<div></div>', parserOptions },
+    { code: '<div role={role} />', parserOptions },
+    { code: '<div role={role || "button"} />', parserOptions },
+    { code: '<div role={role || "foobar"} />', parserOptions },
+    { code: '<div role="tabpanel row" />', parserOptions },
+    { code: '<span role="checkbox" aria-checked="false" aria-labelledby="foo" tabindex="0"></span>', parserOptions },
+    { code: '<Bar baz />', parserOptions }
+  ].concat(basicValidityTests),
+  invalid: [
```
lencioni: Scanning this file, it is a bit hard to spot where valid switches over to invalid. It might be nice to add an empty line above this line.

src/util/values/expressions/UnaryExpression.js
evcohen: Had a `no-implicit-coercion` lint error with `+getValue(argument)`

src/rules/role-requires-aria.js
lencioni: I think this error message would be much better if it just told you the ones you are missing instead of listing all of the ones that are required.

src/util/values/Literal.js
evcohen: Just for learning sake - why do you prefer singles over doubles?

docs/rules/no-unsupported-elements-use-aria.md
```diff
@@ -0,0 +1,23 @@
+# no-unsupported-elements-use-aria
+
+Certain reserved DOM elements do not support ARIA roles, states and properties. This is often because they are not visible, for example `meta`, `html`, `script`, `style`. This rule enforces that these DOM elements do not contain the role and/or aria-* props.
```
lencioni: I think you could use backticks around "role" and "aria-*".

docs/rules/no-unsupported-elements-use-aria.md
```diff
@@ -0,0 +1,23 @@
...
+1. [AX_ARIA_12](https://github.com/GoogleChrome/accessibility-developer-tools/wiki/Audit-Rules#ax_aria_12)
+
+## Rule details
+
+This rule takes no arguments.
+
+### Succeed
+```jsx
+<!-- Good: the meta element should not be given any ARIA attributes -->
+<meta charset="UTF-8">
```
lencioni: This should probably be a self-closing tag because JSX

src/rules/no-unsupported-elements-use-aria.js
```diff
@@ -0,0 +1,44 @@
...
+    // If it's not reserved, then it can have ARIA-* roles, states, and properties
+    if (isReservedNodeType === false) {
+      return;
+    }
+
+    // Check if it has role attribute;
+    const hasRole = hasAttribute(node.attributes, 'role');
+    const hasAria = Object.keys(ARIA).some(prop => hasAttribute(node.attributes, prop.toLowerCase()));
+
+    if (hasRole || hasAria) {
```
lencioni: Not sure if it matters, but this would be more efficient if you reported and returned after checking for role, and then checked for aria attributes.

src/rules/no-unsupported-elements-use-aria.js
```diff
@@ -0,0 +1,44 @@
...
+    const isReservedNodeType = nodeAttrs && nodeAttrs.reserved || false;
+
+    // If it's not reserved, then it can have ARIA-* roles, states, and properties
+    if (isReservedNodeType === false) {
+      return;
+    }
+
+    // Check if it has role attribute;
+    const hasRole = hasAttribute(node.attributes, 'role');
+    const hasAria = Object.keys(ARIA).some(prop => hasAttribute(node.attributes, prop.toLowerCase()));
```
lencioni: I imagine that there were usually be fewer props than possible aria attributes, so it might be better to invert this loop.

tests/src/rules/valid-aria-proptypes.js
lencioni: I think it is still valuable for the error messages to be in the rules. It would be useful to put them all in one file if you find yourself working on multiple error messages at the same time, but that seems less likely than iterating on individual rules.

lencioni:
LGTM Thanks for all your work on this!

src/util/values/Literal.js
lencioni: I mostly prefer consistency, and I've noticed more singles than doubles here.

In general, I like single quotes because in many languages single quotes don't have interpolation, whereas double quotes do. Although there can be a micro micro micro performance benefit in those languages, that's not really important. This gives you a stronger signal of developer intention, which I think makes code more readable. This isn't the case in JS, but it is nice to be consistent for folks who work in multiple languages. And, in any case, they require fewer keypresses.

src/util/values/expressions/UnaryExpression.js
lencioni: Ah interesting. I think it might be worth disabling that rule in this particular case, since you really want to emulate the original code which used the implicit coercion as closely as possible.

lencioni: BTW, you can disable it for just this line by appending `// eslint-disable-line no-implicit-coercion`

tests/src/rules/valid-aria-proptypes.js
evcohen: That makes sense. Agreed. Will just export the message/message functions for easier access in tests. We just want to make sure it's reporting the _right_ error message given the test case.

src/rules/valid-aria-role.js
evcohen: Makes sense. Intention is clearer

.eslintrc.js
```diff
@@ -143,6 +143,7 @@ module.exports = {
     "prefer-const": 2,
     "prefer-spread": 2,
     "prefer-template": 2,
+    'quotes': [2, 'single', 'avoid-escape'],
```
lencioni: :+1: but it's really funny to me to see this line amongst a sea of double quotes. :)

src/rules/no-unsupported-elements-use-aria.js
```diff
@@ -26,11 +26,10 @@ module.exports = context => ({
       return;
     }
 
-    // Check if it has role attribute;
-    const hasRole = hasAttribute(node.attributes, 'role');
-    const hasAria = Object.keys(ARIA).some(prop => hasAttribute(node.attributes, prop.toLowerCase()));
+    const invalidAttributes = Object.keys(ARIA).concat('ROLE');
+    const hasAria = hasAttribute(node.attributes, ...invalidAttributes);
```
lencioni: I forget--does hasAttribute do the case conversion? It must

.eslintrc.js
evcohen: haha just added it for your sanity 😄 
crushing through it right now

src/util/hasAttribute.js
```diff
@@ -7,17 +7,20 @@
  * as the purpose of this linter is to do hard checks of explicit JSX props.
  *
  */
-const hasAttribute = (attributes, attribute) => {
+const hasAttribute = (attributesOnNode, ...attributesToCheck) => {
```
lencioni: Do you know what version of Node added `...`?

.eslintrc.js
lencioni: Thankfully, eslint can autofix this for you. Just run `eslint . --fix` (or something like that)

src/util/hasAttribute.js
evcohen: We transpile as build step, should be good here - the travis config runs on all versions and it just passed

lencioni: Ah, that's right. :+1:

package.json
```diff
@@ -20,7 +20,7 @@
     "build": "rimraf lib && babel src --out-dir lib",
     "prepublish": "npm run lint && npm run test && npm run build",
     "coveralls": "cat ./reports/coverage/lcov.info | coveralls",
-    "lint": "eslint  --config .eslintrc.js src tests",
+    "lint": "eslint  --config .eslintrc.js src/**/*.js tests/**/*.js",
```
lencioni: Instead of globbing like this, I think it would make more sense to use a .eslintignore, and then this becomes `eslint --config .eslintrc.js .`

src/util/getAttributeValue.js
```diff
@@ -1,8 +1,13 @@
 'use strict';
 
-import { getValue, getLiteralValue } from './values';
+import {
+  default as getValue,
```
lencioni: If you like, I think this can be written as:

```js
import getValue, { getLiteralValue } from './values';
```

evcohen: ah true, i was struggling with the import there. let me fix. I think good to merge after that

src/util/values/index.js
```diff
@@ -1,22 +1,54 @@
...
+ * @param value - AST Value object on a JSX Attribute.
+ */
+const getLiteralValue = function getLiteralValue(value) {
+  return LITERAL_TYPES[value.type](value);
+};
+
+export {
+  getValue as default,
+  getLiteralValue
+};
```
lencioni: I think you would be better off moving the exports to the function definitions and dropping the variable assignments. Even if you want to keep this here, you could drop the variable assignments since they are pretty redundant in this case.

```js
export function getValue(value) {
...
export default function getLiteralValue(value) {
```

evcohen: Right - this was my workaround to get the import right. Was struggling how to import in this case because getValue was being imported as undefined with this: 

```js
import { getValue, getLiteralValue } from './values';
```
But seeing your comment above about the import clarifies this

PR #26, opened 2016-04-14
[new] - Implement avoid-positive-tabindex

### LINKED ISSUE #16: Rule proposal: AX_FOCUS_03
https://github.com/GoogleChrome/accessibility-developer-tools/wiki/Audit-Rules#ax_focus_03

Avoid positive integer values for tabIndex

evcohen:
Fixes #16 

@lencioni Fairly simple one.

### DIFF
--- a/src/rules/avoid-positive-tabindex.js
+++ b/src/rules/avoid-positive-tabindex.js
@@ -1,0 +1,41 @@
+/**
+ * @fileoverview Enforce tabIndex value is not greater than zero.
+ * @author Ethan Cohen
+ */
+'use strict';
+
+// ----------------------------------------------------------------------------
+// Rule Definition
+// ----------------------------------------------------------------------------
+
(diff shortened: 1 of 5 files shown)


src/rules/avoid-positive-tabindex.js
```diff
@@ -0,0 +1,41 @@
...
+
+    context.report({
+      node: attribute,
+      message: errorMessage
+    });
+  }
+});
+
+module.exports.schema = [
+  { type: 'object' }
```
lencioni: I think this is an options schema, and since this rule takes no options, it should probably be empty.

lencioni:
LGTM

src/rules/avoid-positive-tabindex.js
evcohen: Yeah it is - all of the rules with no options have same schema, so I'll keep it for consistency right now. If we want to remove we can do it in one commit and separate PR.

PR #28, opened 2016-04-16
[new] - Implement onclick-has-focus rule.

### LINKED ISSUE #15: Rule proposal: AX_FOCUS_02
https://github.com/GoogleChrome/accessibility-developer-tools/wiki/Audit-Rules#ax_focus_02

Elements with onclick handlers must be focusable

evcohen:
Fixes #15

### DIFF
--- a/docs/rules/onclick-has-focus.md
+++ b/docs/rules/onclick-has-focus.md
@@ -4,8 +4,13 @@ Enforce that visible elements with onClick handlers must be focusable. Visible m
 
 To make an element focusable, you can either set the tabIndex property, or use an element type which is inherently focusable.
 
+Elements that are inherently focusable are as follows:
+- `<input>`, `<button>`, `<select>` and `<textarea>` elements which are not disabled
+- `<a>` or `<area>` elements with an `href` attribute.
+
 #### References
 1. [AX_FOCUS_02](https://github.com/GoogleChrome/accessibility-developer-tools/wiki/Audit-Rules#ax_focus_02)
+2. [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_button_role#Keyboard_and_focus)
 
 ## Rule details
 
@@ -16,11 +21,13 @@ This rule takes no arguments.
 <!-- Good: div with onClick attribute is hidden from screen reader -->
 <div aria-hidden onClick={() => void 0} />
 <!-- Good: span with onClick attribute is in the tab order -->
-<span onclick="doSomething();" tabindex="0">Click me!</span>
+<span onClick="doSomething();" tabIndex="0">Click me!</span>
 <!-- Good: span with onClick attribute may be focused programmatically -->
-<span onclick="doSomething();" tabindex="-1">Click me too!</span>
+<span onClick="doSomething();" tabIndex="-1">Click me too!</span>
 <!-- Good: anchor element with href is inherently focusable -->
-<a href="javascript:void(0);" onclick="doSomething();">Click ALL the things!</a>
+<a href="javascript:void(0);" onClick="doSomething();">Click ALL the things!</a>
+<!-- Good: buttons are inherently focusable -->
+<button onClick="doSomething();">Click the button :)</a>
 ```
 
 ### Fail
@@ -29,6 +36,4 @@ This rule takes no arguments.
 <span onclick="submitForm();">Submit</span>
 <!-- Bad: anchor element without href is not focusable -->
 <a onclick="showNextPage();">Next page</a>
-<!-- Bad: May not be hidden from screenreader ¯\_(ツ)_/¯ -->
-<input onClick={() => void 0} type="hidden" />
 ```
--- a/docs/rules/onclick-has-focus.md
+++ b/docs/rules/onclick-has-focus.md
@@ -1,0 +1,34 @@
+# onclick-has-focus
+
+Enforce that visible elements with onClick handlers must be focusable. Visible means that it is not hidden from a screen reader. Examples of non-interactive elements are `div`, `section`, and `a` elements without a href prop. Elements which have click handlers but are not focusable can not be used by keyboard-only users.
+
+To make an element focusable, you can either set the tabIndex property, or use an element type which is inherently focusable.
+
+#### References
+1. [AX_FOCUS_02](https://github.com/GoogleChrome/accessibility-developer-tools/wiki/Audit-Rules#ax_focus_02)
+
+## Rule details
+
+This rule takes no arguments.
+
+### Succeed
+```jsx
+<!-- Good: div with onClick attribute is hidden from screen reader -->
+<div aria-hidden onClick={() => void 0} />
+<!-- Good: span with onClick attribute is in the tab order -->
+<span onclick="doSomething();" tabindex="0">Click me!</span>
+<!-- Good: span with onClick attribute may be focused programmatically -->
+<span onclick="doSomething();" tabindex="-1">Click me too!</span>
+<!-- Good: anchor element with href is inherently focusable -->
+<a href="javascript:void(0);" onclick="doSomething();">Click ALL the things!</a>
+```
+
+### Fail
+```jsx
+<!-- Bad: span with onClick attribute has no tabindex -->
+<span onclick="submitForm();">Submit</span>
+<!-- Bad: anchor element without href is not focusable -->
+<a onclick="showNextPage();">Next page</a>
+<!-- Bad: May not be hidden from screenreader ¯\_(ツ)_/¯ -->
+<input onClick={() => void 0} type="hidden" />
+```
--- a/src/rules/avoid-positive-tabindex.js
+++ b/src/rules/avoid-positive-tabindex.js
@@ -1,0 +1,41 @@
+/**
+ * @fileoverview Enforce tabIndex value is not greater than zero.
+ * @author Ethan Cohen
+ */
+'use strict';
+
+// ----------------------------------------------------------------------------
+// Rule Definition
+// ----------------------------------------------------------------------------
+
+import { getLiteralAttributeValue } from '../util/getAttributeValue';
+
+const errorMessage = 'Avoid positive integer values for tabIndex.';
+
+module.exports = context => ({
+  JSXAttribute: attribute => {
+    const name = attribute.name.name;
+    const normalizedName = name.toUpperCase();
+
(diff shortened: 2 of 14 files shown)


docs/rules/onclick-has-focus.md
```diff
@@ -0,0 +1,34 @@
+# onclick-has-focus
+
+Enforce that visible elements with onClick handlers must be focusable. Visible means that it is not hidden from a screen reader. Examples of non-interactive elements are `div`, `section`, and `a` elements without a href prop. Elements which have click handlers but are not focusable can not be used by keyboard-only users.
+
+To make an element focusable, you can either set the tabIndex property, or use an element type which is inherently focusable.
```
lencioni: Would it be worth giving some examples of elements that are inherently focusable here?

docs/rules/onclick-has-focus.md
```diff
@@ -0,0 +1,34 @@
...
+### Succeed
+```jsx
+<!-- Good: div with onClick attribute is hidden from screen reader -->
+<div aria-hidden onClick={() => void 0} />
+<!-- Good: span with onClick attribute is in the tab order -->
+<span onclick="doSomething();" tabindex="0">Click me!</span>
+<!-- Good: span with onClick attribute may be focused programmatically -->
+<span onclick="doSomething();" tabindex="-1">Click me too!</span>
+<!-- Good: anchor element with href is inherently focusable -->
+<a href="javascript:void(0);" onclick="doSomething();">Click ALL the things!</a>
```
lencioni: Maybe add a `<button>` as well?

src/rules/onclick-has-focus.js
```diff
@@ -0,0 +1,50 @@
...
+import isInteractiveElement from '../util/isInteractiveElement';
+import hasAttribute from '../util/hasAttribute';
+import getNodeType from '../util/getNodeType';
+
+// ----------------------------------------------------------------------------
+// Rule Definition
+// ----------------------------------------------------------------------------
+
+const errorMessage = 'Elements with onClick handlers must be focusable. ' +
+  'You can either set the tabIndex property, or use an element type which is inherently focusable.';
```
lencioni: I wonder if this could be even more useful, e.g. suggest a value for tabIndex or offer inherently focusable elements as examples.

src/rules/onclick-has-focus.js
```diff
@@ -0,0 +1,50 @@
...
+
+module.exports = context => ({
+  JSXOpeningElement: node => {
+    const attributes = node.attributes;
+    if (hasAttribute(attributes, 'onclick') === false) {
+      return;
+    }
+
+    const isVisible = isHiddenFromScreenReader(attributes) === false;
+    if (isVisible === false) {
```
lencioni: Not a big deal, but I think this code would be a little better without the indirection of these variables. Since you have well-named functions, I don't think you sacrifice any readability.

```js
if (isHiddenFromScreenReader(attributes)) {
  return;
}
```

src/rules/onclick-has-focus.js
```diff
@@ -0,0 +1,50 @@
...
+// Rule Definition
+// ----------------------------------------------------------------------------
+
+const errorMessage = 'Elements with onClick handlers must be focusable. ' +
+  'You can either set the tabIndex property, or use an element type which is inherently focusable.';
+
+module.exports = context => ({
+  JSXOpeningElement: node => {
+    const attributes = node.attributes;
+    if (hasAttribute(attributes, 'onclick') === false) {
```
lencioni: nit: I noticed you used camelCase for tabIndex, but not onClick. IIRC the case doesn't matter in this function, but it would be nice to be consistent throughout.

src/util/isInteractiveElement.js
```diff
@@ -11,7 +11,12 @@ const interactiveMap = {
   a: attributes => {
     const hasHref = hasAttribute(attributes, 'href');
     const hasTabIndex = hasAttribute(attributes, 'tabIndex');
-    return (Boolean(hasHref) || !hasHref && Boolean(hasTabIndex));
+    return (Boolean(hasHref) || (!hasHref && Boolean(hasTabIndex)));
+  },
+  area: attributes => {
```
lencioni: jinkies, I haven't thought about area in ages!

src/util/isInteractiveElement.js
```diff
@@ -11,7 +11,12 @@ const interactiveMap = {
...
+    const hasHref = hasAttribute(attributes, 'href');
+    const hasTabIndex = hasAttribute(attributes, 'tabIndex');
+    return (Boolean(hasHref) || (!hasHref && Boolean(hasTabIndex)));
```
lencioni: Since this is identical to `a`, I wonder if there is an opportunity to DRY some things up in here.

tests/src/rules/onclick-has-focus.js
```diff
@@ -0,0 +1,74 @@
...
+const  expectedError  = {
+  message: 'Elements with onClick handlers must be focusable. ' +
+    'You can either set the tabIndex property, or use an element type which is inherently focusable.',
+  type: 'JSXOpeningElement'
+};
+
+ruleTester.run('onclick-has-focus', rule, {
+  valid: [
+    { code: '<div />', parserOptions },
+    { code: '<div aria-hidden onClick={() => void 0} />', parserOptions },
```
lencioni: I think `void 0` is fine, but it is pretty unusual in my experience. I usually see no ops defined like `() => {}` FWIW.

tests/src/rules/onclick-has-focus.js
```diff
@@ -0,0 +1,74 @@
...
+    { code: '<button onClick={() => void 0} className="foo" />', parserOptions },
+    { code: '<option onClick={() => void 0} className="foo" />', parserOptions },
+    { code: '<select onClick={() => void 0} className="foo" />', parserOptions },
+    { code: '<area href="#" onClick={() => void 0} className="foo" />', parserOptions },
+    { code: '<textarea onClick={() => void 0} className="foo" />', parserOptions },
+    { code: '<a tabIndex="0" onClick={() => void 0} />', parserOptions },
+    { code: '<a role="button" href="#" onClick={() => void 0} />', parserOptions },
+    { code: '<a onClick={() => void 0} href="http://x.y.z" />', parserOptions },
+    { code: '<a onClick={() => void 0} href="http://x.y.z" tabIndex="0" />', parserOptions },
+    { code: '<TestComponent onClick={doFoo} />', parserOptions },
```
lencioni: This is valid because it is a Component, right? I wonder if we should make this rule configurable to also look at specified custom components. I guess we can wait until someone requests that feature.

tests/src/rules/onclick-has-focus.js
```diff
@@ -0,0 +1,74 @@
...
+    { code: '<div onClick={() => void 0} />;', errors: [ expectedError ], parserOptions },
+    { code: '<div onClick={() => void 0} role={undefined} />;', errors: [ expectedError ], parserOptions },
+    { code: '<div onClick={() => void 0} {...props} />;', errors: [ expectedError ], parserOptions },
+    { code: '<section onClick={() => void 0} />;', errors: [ expectedError ], parserOptions },
+    { code: '<main onClick={() => void 0} />;', errors: [ expectedError ], parserOptions },
+    { code: '<article onClick={() => void 0} />;', errors: [ expectedError ], parserOptions },
+    { code: '<header onClick={() => void 0} />;', errors: [ expectedError ], parserOptions },
+    { code: '<footer onClick={() => void 0} />;', errors: [ expectedError ], parserOptions },
+    { code: '<div onClick={() => void 0} aria-hidden={false} />;', errors: [ expectedError ], parserOptions },
+    { code: '<input onClick={() => void 0} type="hidden" />;', errors: [ expectedError ], parserOptions },
```
lencioni: This is unexpected. I'd be surprised if hidden inputs are visible to screen readers, and I think that the onClick here won't do anything. We probably need to special case this in `isHiddenFromScreenReader`.

src/rules/avoid-positive-tabindex.js
```diff
@@ -0,0 +1,41 @@
...
+
+    // Check if tabIndex is the attribute
+    if (normalizedName !== 'TABINDEX') {
+      return;
+    }
+
+    // Only check literals because we can't infer values from certain expressions.
+    const value = Number(getLiteralAttributeValue(attribute));
+
+    if (isNaN(value) || value <= 0) {
```
lencioni: Perhaps this is the domain of a different rule or plugin, but I wonder if we should report a different error message on `isNaN` for `tabIndex`.

lencioni:
LGTM. Thanks for cranking these out! I noticed that you are using the old names here so I assume you'll be doing a mass rename later.

evcohen:
Yes, thanks for always doing thorough code reviews, they are super useful.

Yup want to do rename at very last minute so master's README doesn't misdirect people in case they are just picking package up.

tests/src/rules/onclick-has-focus.js
evcohen: Agree - did that for `img-has-alt` and a few others, but will hold off for now.

tests/src/rules/onclick-has-focus.js
evcohen: I'm having trouble finding any documentation for this in either direction 😞

lencioni: It would be pretty bonkers if screen readers announced hidden inputs because that would make many forms pretty unusable and would offer no value to the people using screen readers. This should definitely be considered as hidden from screen reader.

evcohen: haha ok ok I'm not sure where I made this up in my head. fixing now.

evcohen: No, no I think this is correct. You shouldn't put a click handler on a hidden input - that doesn't make any sense. That's the point: you shouldn't have click handlers on elements that can't achieve focus. I think with the following you can achieve access on a screenreader:

```jsx
<input type="hidden tabIndex="0" value="x" />
```

However, I don't think this will be read:
```jsx
<input type="hidden" aria-hidden value="x" tabIndex"0" />
```

I'm in a workshop right now, but I will do a POC with voiceover for mac before we make any changes to this.

evcohen: Turns out both cases were hidden voiceover on mac & Chrome. Will return `true` from `isHiddenFromScreenReader` when

```jsx
<input type="hidden" ...props />
```

PR #33, opened 2016-04-18
Clean up readme a little

lencioni:


evcohen:
Should we hold off on this until we change all the rule names? I think I can get v1.0.0 out by tonight.

lencioni:
I think we should merge this as is, and the re-sort when we rename.

evcohen:
LGTM then

Issue #27, opened 2016-04-14
Review: simplify and clarify rule names


evcohen:
Right now we have:

- [img-uses-alt](docs/rules/img-uses-alt.md): Enforce that img jsx elements use the alt attribute.
- [onclick-uses-role](docs/rules/onclick-uses-role.md): Enforce that non-interactive, visible elements (such as div) that have click handlers use the role attribute.
- [mouse-events-map-to-key-events](docs/rules/mouse-events-map-to-key-events.md): Enforce that onMouseOver/onMouseOut are accompanied by onFocus/onBlur for strictly keyboard users.
- [use-onblur-not-onchange](docs/rules/use-onblur-not-onchange.md): Enforce that onBlur is used instead of onChange.
- [no-access-key](docs/rules/no-access-key.md): Enforce that the accessKey prop is not used on any element to avoid complications with keyboard commands used by a screenreader.
- [label-uses-for](docs/rules/label-uses-for.md): Enforce that label elements have the htmlFor attribute
- [redundant-alt](docs/rules/redundant-alt.md): Enforce img alt attribute does not contain the word image, picture, or photo.
- [no-hash-href](docs/rules/no-hash-href.md): Enforce an anchor element's href prop value is not just #.
- [no-invalid-aria](docs/rules/no-invalid-aria.md): Enforce all aria-* properties are valid.
- [valid-aria-role](docs/rules/valid-aria-role.md): Enforce that elements with ARIA roles must use a valid, non-abstract ARIA role.
- [valid-aria-proptype](docs/rules/valid-aria-proptype.md): Enforce ARIA state and property values are valid.
- [role-requires-aria](docs/rules/role-requires-aria.md): Enforce that elements with ARIA roles must have all required attributes for that role.
- [no-unsupported-elements-use-aria](docs/rules/no-unsupported-elements-use-aria.md): Enforce that elements that do not support ARIA roles, states and properties do not have those attributes.
- [avoid-positive-tabindex](docs/rules/avoid-positive-tabindex.md): Enforce tabIndex value is not greater than zero.

Some of these names are verbose and unclear. Feel free to suggest more consistently clear and simple rule names based on descriptions.

Whatever results from this issue will be used for v1.0.0 release fairly soon.

lencioni:
I think `avoid-positive-tabindex` could be `no-positive-tabindex` to be more consistent with other `no-*` rules.

ljharb:
[...]

PR #36, opened 2016-04-19
[new] - Implement aria-role-supports-attribute

### LINKED ISSUE #17: Rule proposal: AX_ARIA_10

evcohen:
Fixes #17 

Thinking can just be renamed to `role-supports-attribute` or something of the sorts.

Still need to add README and doc file. Will do in the morning.

### DIFF
--- a/src/rules/aria-role-supports-attribute.js
+++ b/src/rules/aria-role-supports-attribute.js
@@ -1,5 +1,6 @@
 /**
- * @fileoverview Enforce that an element does not have an unsupported ARIA attribute.
+ * @fileoverview Enforce that elements with explicit or implicit roles defined contain only
+ * `aria-*` properties supported by that `role`.
  * @author Ethan Cohen
  */
 'use strict';
--- a/src/rules/aria-role-supports-attribute.js
+++ b/src/rules/aria-role-supports-attribute.js
@@ -1,0 +1,54 @@
+/**
+ * @fileoverview Enforce that an element does not have an unsupported ARIA attribute.
+ * @author Ethan Cohen
+ */
+'use strict';
+
+// ----------------------------------------------------------------------------
+// Rule Definition
+// ----------------------------------------------------------------------------
+
+import hasAttribute from '../util/hasAttribute';
+import { getLiteralAttributeValue } from '../util/getAttributeValue';
+import getNodeType from '../util/getNodeType';
+import ROLES from '../util/attributes/role';
+import ARIA from '../util/attributes/ARIA';
+import getImplicitRole from '../util/getImplicitRole';
+
+const errorMessage = (attr, role, tag) =>
+  `The attribute ${attr} is not supported by the role ${role}. This role may be implicit based on the tag ${tag}.`;
+
+module.exports = context => ({
+  JSXOpeningElement: node => {
+    // If role is not explicitly defined, then try and get its implicit role.
+    const type = getNodeType(node);
+    const hasRole = hasAttribute(node.attributes, 'role');
+    const role = hasRole ? getLiteralAttributeValue(hasRole) : getImplicitRole(type, node.attributes);
+
+    // If there is no explicit or implicit role, then assume that the element
+    // can handle the global set of aria-* properties.
+    // This actually isn't true - should fix in future release.
+    if (!role || ROLES[role.toUpperCase()] === undefined) {
+      return;
+    }
+
+    // Make sure it has no aria-* properties defined outside of its property set.
+    const propertySet = ROLES[role.toUpperCase()].props;
+    const invalidAriaPropsForRole = Object.keys(ARIA).filter(attribute => propertySet.indexOf(attribute) === -1);
+    const invalidAttr = hasAttribute(node.attributes, ...invalidAriaPropsForRole);
+
+    if (invalidAttr === false) {
+      return;
+    }
+
+    context.report({
+      node,
+      message: errorMessage(invalidAttr.name.name.toLowerCase(), role.toLowerCase(), type.toLowerCase())
+    });
+
+  }
+});
+
+module.exports.schema = [
+  { type: 'object' }
+];
--- a/src/rules/aria-role-supports-attribute.js
+++ b/src/rules/aria-role-supports-attribute.js
@@ -15,8 +15,13 @@ import ROLES from '../util/attributes/role';
 import ARIA from '../util/attributes/ARIA';
 import getImplicitRole from '../util/getImplicitRole';
 
-const errorMessage = (attr, role, tag) =>
-  `The attribute ${attr} is not supported by the role ${role}. This role may be implicit based on the tag ${tag}.`;
+const errorMessage = (attr, role, tag, isImplicit) => {
+  if (isImplicit) {
+    return `The attribute ${attr} is not supported by the role ${role}. This role is implicit on the element ${tag}.`;
+  }
+
+  return `The attribute ${attr} is not supported by the role ${role}.`;
+};
 
 module.exports = context => ({
   JSXOpeningElement: node => {
@@ -24,6 +29,7 @@ module.exports = context => ({
     const type = getNodeType(node);
     const hasRole = hasAttribute(node.attributes, 'role');
     const role = hasRole ? getLiteralAttributeValue(hasRole) : getImplicitRole(type, node.attributes);
+    const isImplicit = role && !hasRole;
 
     // If there is no explicit or implicit role, then assume that the element
     // can handle the global set of aria-* properties.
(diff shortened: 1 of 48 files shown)


src/rules/aria-role-supports-attribute.js
```diff
@@ -0,0 +1,54 @@
...
+
+import hasAttribute from '../util/hasAttribute';
+import { getLiteralAttributeValue } from '../util/getAttributeValue';
+import getNodeType from '../util/getNodeType';
+import ROLES from '../util/attributes/role';
+import ARIA from '../util/attributes/ARIA';
+import getImplicitRole from '../util/getImplicitRole';
+
+const errorMessage = (attr, role, tag) =>
+  `The attribute ${attr} is not supported by the role ${role}. This role may be implicit based on the tag ${tag}.`;
```
lencioni: We should be able to determine if the role is implicit or not, and tailor the error message appropriately.

src/rules/aria-role-supports-attribute.js
```diff
@@ -0,0 +1,54 @@
...
+    // can handle the global set of aria-* properties.
+    // This actually isn't true - should fix in future release.
+    if (!role || ROLES[role.toUpperCase()] === undefined) {
+      return;
+    }
+
+    // Make sure it has no aria-* properties defined outside of its property set.
+    const propertySet = ROLES[role.toUpperCase()].props;
+    const invalidAriaPropsForRole = Object.keys(ARIA).filter(attribute => propertySet.indexOf(attribute) === -1);
+    const invalidAttr = hasAttribute(node.attributes, ...invalidAriaPropsForRole);
```
lencioni: It is a little unintuitive that a function named `has*` would return anything other than a boolean.

src/rules/aria-role-supports-attribute.js
```diff
@@ -0,0 +1,54 @@
...
+
+    if (invalidAttr === false) {
+      return;
+    }
+
+    context.report({
+      node,
+      message: errorMessage(invalidAttr.name.name.toLowerCase(), role.toLowerCase(), type.toLowerCase())
```
lencioni: I'm a little concerned that the case conversions will result in error messages that don't match the original code.

src/util/getImplicitRole.js
```diff
@@ -0,0 +1,22 @@
...
+ * @param attributes - The collection of attributes on the node.
+ * @returns {String} - String representing the node's implicit role or '' if it doesn't exist.
+ */
+export default function getImplicitRole(type, attributes) {
+  const normalizedType = type.toUpperCase();
+
+  if (implicitRoles[normalizedType]) {
+    return implicitRoles[normalizedType](attributes);
+  } else if (DOM[normalizedType] && DOM[normalizedType].role) {
+    return DOM[normalizedType].role;
```
lencioni: It's a little strange that some implicit roles are defined in JSON and others are defined in JS switch statements. I think either is fine, but it would probably be best to not have two ways to do this.

src/util/implicitRoles/img.js
```diff
@@ -0,0 +1,16 @@
...
+ */
+export default function getImplicitRoleForImg(attributes) {
+  const hasAlt = hasAttribute(attributes, 'alt');
+
+  if (hasAlt && getLiteralAttributeValue(hasAlt) === '') {
+    return 'presentation';
+  }
+
+  return DOM.IMG.role;
+}
```
lencioni: I think this is where our earlier decision to allow img `alt=''` without `role="presentation"` is really solidified. I wonder if we should use the implicit role stuff in the img-has-alt rule?

src/util/implicitRoles/input.js
```diff
@@ -0,0 +1,38 @@
...
+      case 'RADIO':
+        return 'radio';
+      case 'RANGE':
+        return 'slider';
+      // These will resolve to `textbox` in default.
+      case 'EMAIL':
+      case 'PASSWORD':
+      case 'SEARCH': // with [list] selector it's combobox
+      case 'TEL': // with [list] selector it's combobox
+      case 'URL': // with [list] selector it's combobox
```
lencioni: I don't think I understand these comments.

src/util/getImplicitRole.js
evcohen: The JSON file has the base implicit role - meaning that the role listed there is the implicit role when no other props are defined. Sometimes that role changes for certain elements when props are used (i.e. `a` elements have no implicit role until an `href` attribute is used on the element).

src/util/implicitRoles/input.js
evcohen: Yeah, I ported this over from the google a11y constants file which said input[type="search"][list] selector mapped to combobox. not sure what the list selector is, but can investigate further. just left it there for now.

src/util/getImplicitRole.js
lencioni: Yeah, I totally understand that. I just thing that either we should go all in on the declarative JSON approach and find a way to express the implicit roles when certain props are used, or we should go all in on the other approach. That way, if I am trying to answer the question "what implicit roles might this element have" I can look in one place instead of two.

evcohen: That makes sense. Will move everything it into function approach so that it's consistent and we don't have to come up with some weird convention in JSON to represent this. The cost of running the code is minimal

src/rules/aria-role-supports-attribute.js
evcohen: I agree it's a little weird. Maybe `getAttribute` is better and it returns the attribute itself or undefined. We can check for truthiness there without a heavy cost of changing much of the other code.

lencioni: Yeah, that sounds better. Doesn't need to be part of this PR though.

src/rules/aria-role-supports-attribute.js
```diff
@@ -15,15 +15,21 @@ import ROLES from '../util/attributes/role';
...
 import getImplicitRole from '../util/getImplicitRole';
 
-const errorMessage = (attr, role, tag) =>
-  `The attribute ${attr} is not supported by the role ${role}. This role may be implicit based on the tag ${tag}.`;
+const errorMessage = (attr, role, tag, isImplicit) => {
+  if (isImplicit) {
+    return `The attribute ${attr} is not supported by the role ${role}. This role is implicit on the element ${tag}.`;
+  }
+
+  return `The attribute ${attr} is not supported by the role ${role}.`;
```
lencioni: Awesome! Do you think it makes sense to list supported attributes for the role?

lencioni:
👍

src/rules/aria-role-supports-attribute.js
```diff
@@ -0,0 +1,60 @@
...
+      node,
+      message: errorMessage(invalidAttr.name.name, role, type, isImplicit)
+    });
+
+  }
+});
+
+module.exports.schema = [
+  { type: 'object' }
+];
```
lencioni: This rule needs documentation.

src/rules/aria-role-supports-attribute.js
evcohen: 😄 ...eh - the lists can get lengthy and will make the error message verbose. I think it'd be better to just document it, for now. Next, I think we can try to flesh out warning messages to get smart and try and make 'intelligent' recommendations based on context.

If the user is willing to go through the code, it's also pretty clear in `src/util/attributes/role.json`.

evcohen:
@lencioni thoughts on name? Mentioned in initial comment that i think it should be just reduced to `role-supports-attribute`

src/rules/aria-role-supports-attribute.js
lencioni: Makes sense. I like the idea of "intelligent" recommendations--as long as we can get the implementation right. :)

lencioni:
Looking at #27, we have the following similar names:

- valid-aria-role -> aria-role
- valid-aria-proptype -> aria-proptypes
- role-requires-aria -> aria-props-for-role

I think this rule is most similar to `role-requires-aria` -> `aria-props-for-role`, but I'm noticing the name we chose to change that one to is ambiguous in the context of this rule. So, what if we call this rule `role-supports-aria-props` and the other one `role-has-aria-props` (or `role-has-required-aria-props`)?

evcohen:
Awesome, I like it. Going to merge this, do the rename and maybe some other cleanup (with `hasAttribute` => `getAttribute`) then publish v1.0.0 today 🎉

lencioni:
I think that `role-has-required-aria-props` is clearer than `role-has-aria-props` FWIW. Looking forward to 1.0.0! Thanks for all your hard work on this. :)

PR #45, opened 2016-05-06
Import all roles from DPUB-ARIA and test them

darobin:
ARIA `role` is meant to be extensible by other specifications. [Digital Publishing WAI-ARIA Module 1.0](https://www.w3.org/TR/dpub-aria-1.0/) is doing just that. This change adds the new roles that DPUB ARIA defines.


evcohen:
This is awesome, thank you! I didn't know that 👍 

[v1.2.0](https://github.com/evcohen/eslint-plugin-jsx-a11y/releases/tag/1.2.0) published.

PR #48, opened 2016-05-20
Fix checks involving the tabIndex attribute that do not account for integer literals

SohKai:
`tabIndex` values can be specified with both strings (ie. `tabIndex="0"`) or integers (ie. `tabIndex={0}`). Currently, some of the checks for `tabIndex` simply coerce the attribute's value into a boolean, so using `{0}` fails.

This change also enforces `tabIndex`s to have valid values in order for them to be considered present, rather than just checking for their existence.


evcohen:
dude you rock. thank you

evcohen:
[v1.2.2](https://github.com/evcohen/eslint-plugin-jsx-a11y/releases/tag/1.2.2) published with this fix.

PR #52, opened 2016-06-02
Spelling suggestions for aria-* props using levenshtein distance

evcohen:
Just a small POC for spelling suggestions for invalid `aria-*` properties. Can also be applied to role, but wanted to keep surface area small just to experiment. Can eventually use more advanced spell-checking algorithms, too.


src/util/getSuggestion.js
```diff
@@ -0,0 +1,27 @@
...
+// Max distance away from value to be considered a decent suggestion.
+const THRESHOLD = 3;
+// Number of suggestions to return.
+const LIMIT = 2;
+
+export default function getSuggestion(value, dictionary = DICTIONARY, threshold = THRESHOLD, limit = LIMIT) {
+  const normalizedValue = value.toUpperCase();
+
+  const distances = DICTIONARY.reduce((suggestions, word) => {
+    suggestions[word] = levenshtein(normalizedValue, word);
```
lencioni: There is an implicit dependency here that `word` is uppercase. You might want to convert it to uppercase here.

src/util/getSuggestion.js
```diff
@@ -0,0 +1,27 @@
...
+    return suggestions;
+  }, {});
+
+  return Object.keys(distances)
+    .filter(suggestion => distances[suggestion] <= threshold)
+    .sort((a, b) => distances[a] - distances[b]) // Sort by distance
+    .slice(0, limit)
+    .map(suggestion => suggestion.toLowerCase());
```
lencioni: I'm worried that converting it to lowercase here might cause some inaccurate suggestions.

src/util/suggestions/levenshtein.js
```diff
@@ -0,0 +1,46 @@
...
+
+        const cost = Math.min(substitution, Math.min(insertion, deletion));
+
+        matrix[i][j] = cost;
+      }
+    }
+  }
+
+  return matrix[second.length][first.length];
+}
```
lencioni: I personally think you'd be better off just bringing in a dependency instead of having this implemented here.

tests/src/rules/aria-props.js
```diff
@@ -25,10 +25,21 @@ const parserOptions = {
 
 const ruleTester = new RuleTester();
 
-const errorMessage = name => ({
-  message: `${name}: This attribute is an invalid ARIA attribute.`,
-  type: 'JSXAttribute'
-});
+import getSuggestion from '../../../src/util/getSuggestion';
```
lencioni: imports should generally be placed at the top. They are hoisted before other things, so having them in different places may result in confusing behavior.

src/util/getSuggestion.js
```diff
@@ -0,0 +1,27 @@
...
+}
```
lencioni: Might be nice to have a test for this directly.

lencioni:
Overall, this looks good to me. I think this will generate more helpful error messages which should help people do the right thing!

src/util/getSuggestion.js
evcohen: yeah, it should be whatever the dictionary provides. this was specific to the aria dictionary so good catch

src/util/suggestions/levenshtein.js
evcohen: 😝 just having some fun. added one in next commit

PR #75, opened 2016-08-10
Add support for custom words in img-redundant-alt

badtant:
Usefull for users who are writing apps in a different language.


evcohen:
this is really great, thanks again!

Issue #78, opened 2016-08-24
Rule to disallow non-native DOM element interactions


ljharb:
For example, `onClick` should only be valid on `<a>`, `<button>`, `<input>`, `<textarea>`, `<select>`… i'm sure I'm forgetting some, but there should not be an `onClick` on a `<div>`, for example.

The same should apply to keyboard events.

ljharb:
This is a followup from https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/77#issuecomment-241918479

evcohen:
Is this something we would want to error on or just warn with suggestion to use a natively-interactable DOM element? I know you've felt in the past that warning isn't entirely useful, but there are ways to make elements like `<div>` clickable as long as a proper `role` and a positive `tabIndex` are applied to the element.

ljharb:
How someone wishes to configure it would be up to them - and a helpful message is great - but I certainly wouldn't want anything to be configured as less than an error.

There are totally a myriad of workarounds to make elements like `<div>` clickable, but then you lose the semantic value of the tag name.

evcohen:
To be clear, I was talking about our recommended config 😛

ljharb:
aha, in that case it's up to you. It's certainly an opinionated rule, so a warning might make sense there.

sir-marc:
How do you handle onClick events on divs then?
For example a Modal. If i click on the Modal outside the popup itself, i want it to close. The Modalcontainer is a div for sure (or am i getting something wrong?). So how would i apply this use case then?

PR #79, opened 2016-08-25
New rules: no-static-element-interaction and click-events-have-key-events

### LINKED ISSUE #77: Rule proposal: require clickable non-button/non-link elements listen for space or enter
If you are making a non-button or non-link element clickable, you also need to wire up keyboard listeners for space and enter. While I don't think we will be able to ensure that this is all implemented correctly, we can likely at least add a warning for any DOM element that is not `a` or `button` that has `onClick` and not a keyboard event listener.

cc @backwardok

evcohen:
Fixes #77 and #78 
Willing to clarify the naming! If anyone can find any good a11y resources to add to the docs for this, that would be super helpful.

Thanks 😄


docs/rules/click-events-have-key-events.md
```diff
@@ -0,0 +1,19 @@
+# click-events-have-key-events
+
+Enforce `onclick` is accompanied by at least one of the following: `onkeyup`, `onkeydown`, `onkeypress`. Coding for the keyboard is important for users with physical disabilities who cannot use a mouse, AT compatibility, and screenreader users.
```
ljharb: should these handler names be camelCased?

docs/rules/click-events-have-key-events.md
```diff
@@ -0,0 +1,19 @@
...
+
+## Rule details
+
+This rule takes no arguments.
+
+### Succeed
+```jsx
+<div onClick={() => void 0} onKeyDown={_handleKeyDown}/>
```
ljharb: maybe `this.handleKeyDown`, if we're going to pick an example that doesn't have a definition?

Also, `() => {}` might be better than `() => void 0` for making it clear that it's a placeholder function.

ljharb:
Thanks, these look great! `click-events-have-key-events` seems like a good name; for the other, maybe `no-non-native-interaction`? I'm not really sure - your name is fine too ¯\\\_(ツ)_/¯

Issue #65, opened 2016-07-20
role-has-required-aria-props being applied to react components


majapw:
We are trying to build a React component that takes in aria-props and passes them down to a child `<input />` element. Thus, we have properties on our component like `ariaAutocomplete`, `ariaExpanded`, `ariaOwns`, etc.

However, we can't add a `role` prop because when we try to assign  `role="combobox"` to the React component to be passed down, we end up triggering this rule because we are using `ariaExpanded` and not `aria-expanded` at this level.

Would it be possible to limit this rule to DOM level elements?

lencioni:
Alternatively, what do you think about checking for kebab-case ARIA props on DOM components and camelCase ARIA props on non-DOM components?

ljharb:
I think that instead, the attributes on the wrapper should be `expanded`, not `ariaExpanded`, for example. The wrapper should be abstracting around the dom element - iow, the wrapper IS the dom element conceptually. So when i’m using the wrapper, I shouldn’t know or care what’s inside it - and i’m assigning aria props as if that’s the leaf node, and an "aria" prefix makes that abstraction leaky.

evcohen:
@ljharb i agree with encapsulating `aria` props to the low-level DOM component, however couldn't this cause collisions with props that have different semantic meaning than the `aria` prop itself (i.e. `owns` could be a prop that could mean something totally different than what `aria-owns` means in the context of the application)

We can just add a config setting for user to note his/her convention for wrapper components

ljharb:
It might make sense for all aria rules to only ever touch DOM components by default, and have config options both for conventions, and for which components (like `Input`, for example) the rules should apply to? It seems like some rules behave this way but some do not.

evcohen:
Yeah that makes sense, it just makes sharing configs shitty as people have different naming conventions for components. People who use `eslint-config-airbnb` would have to override all of these rules to apply the right component name to the config, which sucks.

It probably makes sense to have default mappings of _probable_ wrapper components (`input` => `Input`, `Image` => `img`) and just go through a series of checks:

1. First check for camelCase
2. Then check for prop without `aria`
3. If neither, then warn.

ljharb:
Perhaps it would be better to use global config "settings", rather than per-rule configs for this? That way, all aria rules could read from the same settings value. `eslint-plugin-import` uses this to configure module extension resolution, for example.

evcohen:
[...]

PR #81, opened 2016-08-28
Update schemas for all rules to objects

### LINKED ISSUE #76: Be able to specify component and props for rules like href-no-hash
We have a component that takes a couple of `href` props, like `hrefLeft` and `hrefRight`. These end up being used further down in the chain as `href` attributes, so it would be nice to be able to specify this in the configuration for this rule so it can be checked.

There may be other rules that could use a similar style of configuration.

cc @wyattdanger

evcohen:
Fixes #76 - can pass down `props` option to `href-no-hash` rule.

CC: @lencioni


ljharb:
This makes it semver-major - is there not a way to accept the previous config pattern along with the new one?

evcohen:
There is, but I am in the middle of creating that global settings suggested in #65 and was going to major bump with both of these

ljharb:
@evcohen would it make sense to take the opportunity to change all of the plugin option schemas to accept an object, so future option additions don't cause a major bump?

docs/rules/anchor-has-content.md
```diff
@@ -7,7 +7,19 @@ Enforce that anchors have content and that the content is accessible to screen r
...
+{
+    "rules": {
+        "jsx-a11y/anchor-has-content": [ 2, {
+            "components": [ "Anchor" ],
+          }],
+    }
+}
+```
+
+For the `components` option, these strings determine which JSX elements (**including** `<a>` and `<Link>` by default) should be checked for having content. This is a good use case when you have a wrapper component that simply renders an `a` element (like in React):
```
lencioni: Am I reading this correctly that `<Link>` is checked for content by default? I'm not so sure this is a good idea, especially if you are unable to remove this from the list.

docs/rules/href-no-hash.md
```diff
@@ -4,7 +4,20 @@ Enforce an anchor element's href prop value is not just #. You should use someth
...
 
-This rule takes one optional argument of type string or an array of strings. These strings determine which JSX elements should be checked for the `href` prop **including** `a` by default. This is a good use case when you have a wrapper component that simply renders an `a` element (like in React):
+This rule takes one optional object argument of type object:
+
+```json
+{
+    "rules": {
+        "jsx-a11y/href-no-hash": [ 2, {
+            "components": [ "Link" ],
+            "props": [ "hrefLeft", "hrefRight" ]
```
lencioni: It seems like the configuration of props is specific to specific components, so it should be connected to the components. e.g. `{ "SpecialLink": ["hrefLeft", "hrefRight"] }`

docs/rules/anchor-has-content.md
ljharb: This was already the case - why wouldn't it be a good idea?

lencioni: Since I can create any component called `Link`, I might have a different API or use-case than what this rule assumes. If I am unable to configure this rule to not warn on every usage of my `Link`, then I will likely disable this rule entirely.

ljharb: Either way, I'd be in favor of removing the `Link` default - not everyone uses react-router.

docs/rules/anchor-has-content.md
```diff
@@ -7,7 +7,19 @@ Enforce that anchors have content and that the content is accessible to screen r
...
 ## Rule details
 
-This rule takes one optional argument of type string or array of strings. These strings determine which JSX elements should be checked including `a` and `Link` (react-router) by default. This is a good use case when you have a wrapper component that simply renders a anchor element (like in React):
+This rule takes one optional object argument of type object:
+
+```json
+{
+    "rules": {
+        "jsx-a11y/anchor-has-content": [ 2, {
+            "components": [ "Anchor" ],
```
ljharb: Does this override `<a>` or supplement it? I assume it's additive.

evcohen: additive - it says below:
```
For the `components` option, these strings determine which JSX elements (**including** `<a>` by default)
```

ljharb: "By default" implies to me that if I supply something, it would override.

PR #92, opened 2016-09-12
Fix x-has-content rules to check for children prop

### LINKED ISSUE #87: anchor-has-content false positive for object spread
A stripped-down version of the component that `anchor-has-content` has an issue with:
```jsx
const Button = ({ className, type, link, title, children }) => {
  const commonProps = { className, title, children };
  if (link) {
    return <Link {...commonProps} to={link} />;
  }
  return <button {...commonProps} type={type} />;
};
```

evcohen:
Fixes #87


lencioni:
👍

PR #99, opened 2016-10-06
Add "switch" aria role

adamrneary:
See http://rawgit.com/w3c/aria/master/aria/aria.html#switch for more details.

I am not sure if the adjustment to the test is necessary, but it doesn't hurt!

### DIFF
--- a/src/util/attributes/role.json
+++ b/src/util/attributes/role.json
@@ -2446,6 +2446,31 @@
     "requiredProps": [],
     "abstract": true
   },
+  "SWITCH": {
+    "requiredProps": [
+      "ARIA-CHECKED"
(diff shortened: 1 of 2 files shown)


lencioni:
I agree that it would be nice to add those tests, but I don't think they need to hold up releasing this bugfix if @evcohen has time.

evcohen:
@ljharb we get the `aria-checked` required test for free with `role-has-required-aria-props` rule tests. maybe there is something we can do to add better messaging for `aria-proptypes` rule to add custom messages for each `aria-*` prop and how they treat certain values.

thanks for submitting this @adamrneary !! 💃

PR #102, opened 2016-10-08
fix(rule): img-has-alt - img must have an alt attribute

### LINKED ISSUE #100: `jsx-a11y/img-has-alt`
First of all, thank you for this plugin. This is awesome!

Currently `jsx-a11y/img-has-alt` warns with the following message
 ```
img elements must have an alt prop or use role="presentation"
```
IMHO, 
- it should be an `error` instead of `warning`.
- error message should say something like `img elements must have an alt prop, either with a meaningful text or empty string for decorative images`

Also avoid `aria` (role, states, property suggestion) if it can be achieved by/in HTML. 

You might want to refer http://webaim.org/techniques/alttext/#basics for further details.

Happy to raise PR!

### PR DESCRIPTION
**Pull request by sarbbottam**
closes #100 

Since the new error/warn message does not state about presentation, removed the corresponding code.

### DIFF
--- a/__tests__/src/rules/img-has-alt-test.js
+++ b/__tests__/src/rules/img-has-alt-test.js
@@ -24,13 +24,18 @@ const parserOptions = {
 const ruleTester = new RuleTester();
 
 const missingPropError = type => ({
-  message: `${type} elements must have an alt prop or use role="presentation".`,
+  message: `${type} elements must have an alt prop, either with meaningful text, or an empty string for decorative images.`,
   type: 'JSXOpeningElement',
 });
 
 const altValueError = type => ({
   message: `Invalid alt value for ${type}. \
-Use alt="" or role="presentation" for presentational images.`,
+Use alt="" for presentational images.`,
+  type: 'JSXOpeningElement',
+});
+
+const preferAltError = () => ({
+  message: 'Prefer alt="" over role="presentation". First rule of aria is to not use aria if it can be achieved via native HTML.',
   type: 'JSXOpeningElement',
 });
 
@@ -68,9 +73,6 @@ ruleTester.run('img-has-alt', rule, {
     { code: '<img alt="" role="presentation" />', parserOptions },
     { code: '<img alt="" role={`presentation`} />', parserOptions },
     { code: '<img alt="" role={"presentation"} />', parserOptions },
-    { code: '<img role="presentation" />;', parserOptions },
-    { code: '<img alt={undefined} role="presentation" />;', parserOptions },
-    { code: '<img alt role="presentation" />;', parserOptions },
     { code: '<img alt="this is lit..." role="presentation" />', parserOptions },
     { code: '<img alt={error ? "not working": "working"} />', parserOptions },
     { code: '<img alt={undefined ? "working": "not working"} />', parserOptions },
@@ -115,6 +117,9 @@ ruleTester.run('img-has-alt', rule, {
     { code: '<img role />', errors: [missingPropError('img')], parserOptions },
     { code: '<img {...this.props} />', errors: [missingPropError('img')], parserOptions },
     { code: '<img alt={false || false} />', errors: [altValueError('img')], parserOptions },
+    { code: '<img alt={undefined} role="presentation" />;', errors: [altValueError('img')], parserOptions },
+    { code: '<img alt role="presentation" />;', errors: [altValueError('img')], parserOptions },
+    { code: '<img role="presentation" />;', errors: [preferAltError()], parserOptions },
 
     // CUSTOM ELEMENT TESTS FOR ARRAY OPTION TESTS
     {
--- a/src/rules/img-has-alt.js
+++ b/src/rules/img-has-alt.js
@@ -35,17 +35,20 @@ module.exports = {
       const isPresentation = roleProp && typeof roleValue === 'string'
         && roleValue.toLowerCase() === 'presentation';
 
-      if (isPresentation) {
-        return;
-      }
-
       const altProp = getProp(node.attributes, 'alt');
 
       // Missing alt prop error.
       if (altProp === undefined) {
+        if (isPresentation) {
+          context.report({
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: src/rules/img-has-alt.js [hunk 1]
```diff
@@ -40,22 +40,13 @@ module.exports = {
         return;
       }
 
-      const roleProp = getProp(node.attributes, 'role');
-      const roleValue = getPropValue(roleProp);
-      const isPresentation = roleProp && typeof roleValue === 'string'
-        && roleValue.toLowerCase() === 'presentation';
-
-      if (isPresentation) {
-        return;
```
ljharb: removing this constitutes a breaking change. why does something with the presentation role also need an empty string alt prop?

src/rules/img-has-alt.js
```diff
@@ -40,22 +40,13 @@ module.exports = {
...
-      }
-
       const altProp = getProp(node.attributes, 'alt');
 
       // Missing alt prop error.
       if (altProp === undefined) {
         context.report({
           node,
-          message: `${nodeType} elements must have an alt prop or use role="presentation".`,
+          message: `${nodeType} elements must have an alt prop, either with a meaningful text or empty string for decorative images.`,
```
ljharb: "img elements must have an alt prop, either with meaningful text, or an empty string for decorative images"

src/rules/img-has-alt.js
sarbbottam: >why does something with the presentation role also need an empty string alt prop?

I guess the question should have been ...
>Why `role=presentation` needs to be used if it can be addressed using `alt=""`

Screen reader ignores a image with `alt=""` attribute. `alt=""` must be preferred over `role="presentation"`. [First rule of aria is not to use aria if it can be achieved via native HTML.](https://www.w3.org/TR/aria-in-html/#first-rule-of-aria-use)

>removing this constitutes a breaking change.

May be this PR should not be merged till a major version.

src/rules/img-has-alt.js
sarbbottam: - [x]

evcohen:
I think this is fair - maybe one thing we can do is add a eslint fix for this to remove `role="presentation"` and replace with `alt=""` if it doesn't already exist.

will hold off on merging until schema breaking change gets in and we can release a major.

src/rules/img-has-alt.js
evcohen: maybe we should still check for role="presentation" and inform to use `alt=""` instead with that exact reasoning:

```
Prefer alt="" over role="presentation". The first rule of aria is to not use aria if it can be achieved via native HTML.
```

This way, users can learn from the rule even if they are both logically correct and equivalent.

evcohen:
Just merged in updated schema rules which introduced breaking changes. Let me know when you've updated so I can publish all breaking changes in one release 🚀

src/rules/img-has-alt.js
sarbbottam: Are you suggesting to display different messages depending on the situation?

- no `alt` no `role="presentation"`: `img elements must have an alt prop, either with meaningful text, or an empty string for decorative images`
- no `alt` but `role="presentation"`: `prefer alt="" over role="presentation". The first rule of aria is to not use aria if it can be achieved via native HTML.`

evcohen: yes and maybe a third case when `alt` is present and `role="presentation"`:
```
Role="presentation" may or may not cause a screen reader to provide alternative text for this image. If this is a presentational image, set alt="" and remove role="presentation"; otherwise, remove role="presentation" and provide meaningful text for the alt prop.
```

or something simpler than that

evcohen: no rush, @sarbbottam, want to hear @ljharb and @lencioni's opinions first, anyway. thanks for contributing!

lencioni: If we have determined that `alt=""` is better than `role="presentation"` then I think it is worth enforcing it in this rule. It would be nice to include an autofix for this, to help folks who are upgrading and need to make this change.

evcohen: doing some research, i'm seeing some contradictory opinions regarding `role="presentation"` vs `alt=""`. From what I've gathered though, I think `alt=""` is always required and it is harmless to add `role="presentation"`. 

Therefore, I think the two error states mentioned by @sarbbottam [above](https://github.com/evcohen/eslint-plugin-jsx-a11y/pull/102#discussion_r82932886) are correct to me.
- no `alt` no `role="presentation"`: `img elements must have an alt prop, either with meaningful text, or an empty string for decorative images`
- no `alt` but `role="presentation"`: `prefer alt="" over role="presentation". The first rule of aria is to not use aria if it can be achieved via native HTML.`

Sources: 

1. https://www.w3.org/WAI/tutorials/images/decorative/#image-used-as-part-of-page-design
2. https://www.paciellogroup.com/blog/2011/04/html5-accessibility-chops-the-alt-decision/#d2

sarbbottam: Updated as per @evcohen last comment.

__tests__/src/rules/img-has-alt-test.js
```diff
@@ -68,9 +73,6 @@ ruleTester.run('img-has-alt', rule, {
     { code: '<img alt="" role="presentation" />', parserOptions },
     { code: '<img alt="" role={`presentation`} />', parserOptions },
     { code: '<img alt="" role={"presentation"} />', parserOptions },
```
sarbbottam: `role=presentation` is harmless, but of no use, redundant code. I will defer it to you all.

sarbbottam:
I am not familiar w/ auto-fixing, any pointer to a reference code?

sarbbottam:
@ljharb could you take another look?

sarbbottam:
> we can do a fast follow with the auto-fix. Not sure how to do it either so will take a look and we can collaborate on it

@evcohen sure.

Issue #91, opened 2016-09-12
[no-static-element-interactions] Suppress warning if specific role is specified?


tleunen:
Should the rule be auto-suppressed when there's a role attribute on the element?

``` jsx
<div role="button" onClick={sth}>Click me</div>
```

ljharb:
If it's a button, why isn't it a button?

tleunen:
Because of a CSS framework that style all buttons but for this specific case you don't want it to be styled as a "true" button? ^^

Note I agree if it has a click function, it should be a link or a button, but sometimes it's not possible and I thought the role was meant for that.

lencioni:
Role is important for that, but if you want this element to be accessible and usable, you also need to give it a tabIndex so it can be focused, and provide a keyboard handler to allow space or enter to activate it.

tleunen:
Is `<a onClick={sth}>` valid or does it require something more?

lencioni:
That needs an `href` that points to the relevant content (e.g. not `#`).

tleunen:
So the only valid element that can have `onClick` is the `<button>`?

lencioni:
Also interactive form elements like input, and links.

lencioni:
Oh, also `<area>`

mastertinner:
I'm experiencing a similar problem like @tleunen. I currently have a couple of `<div>`s with the `role="button"` and `onClick={sth}`. These are mainly chevrons to open collapsible areas which I don't want to be styled like a button at all and which don't point to a specific URL. So I can't apply either `<button>` (messes up all styles and has different style presets depending on the browser, which I would have to explicitly overwrite for each of these interactive elements) nor `<a>` (since my elements don't point to a new URL but only open a collapsible area).

Do you guys have any suggestions for these cases? According to http://uxmovement.com/buttons/when-to-use-a-button-or-link/ they should be links from a UX perspective but I don't see how I would add a `href` pointing to something in my case.

sir-marc:
More like a best practice question:
If I have a Modal component with a Modalcontainer Component. 
Both are divs, how do I handle an onClick event on the ModalContainer Component? I want then to close the Modal so I need an onClick handler on the div element? Do i get something terribly wrong or is this case acceptable?

shulcsm:
What about drawable canvas?

seansean11:
[...]

PR #105, opened 2016-10-11
[no-static-element-interactions] List valid interactive elements in docs

vdh:
Updated documentation for `no-static-element-interactions` to list the interactive elements defined in `isInteractiveElement`.

The conversations in #91 seems to indicate it would be good to have an explicit list of these elements?

### DIFF
--- a/docs/rules/no-static-element-interactions.md
+++ b/docs/rules/no-static-element-interactions.md
@@ -2,6 +2,8 @@
 
 Enforce non-interactive DOM elements have no interactive handlers. Static elements such as `<div>` and `<span>` should not have mouse/keyboard event listeners. Instead use something more semantic, such as a button or a link.
 
+Valid interactive elements are: `<a>`, `<area>`, `<button>`, `<input>`, `<option>`, `<select>`, or `<textarea>`.
+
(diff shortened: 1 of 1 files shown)


docs/rules/no-static-element-interactions.md
```diff
@@ -2,6 +2,8 @@
 
 Enforce non-interactive DOM elements have no interactive handlers. Static elements such as `<div>` and `<span>` should not have mouse/keyboard event listeners. Instead use something more semantic, such as a button or a link.
 
+Valid interactive elements are: `<a>`, `<area>`, `<button>`, `<input>`, `<option>`, `<select>`, or `<textarea>`.
```
evcohen: This is a great addition, thank you! Just a quick modification - `<a>` is not interactive unless it has either `href` or `tabIndex`, and `<input>` must not be `hidden`.

Logic for this lives [here](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/src/util/isInteractiveElement.js#L6)

vdh: I've updated it with a more clarified list

PR #108, opened 2016-10-15
[build] Reduce package size with additions to .npmignore

### LINKED ISSUE #107: Reduce size of npm package
Based on my analysis, v1.2.3 of `eslint-plugin-jsx-a11y` is 992k on disk! And this is because it includes `docs/`, `tests/`, and `src/` in addition to the `lib/` directory, the only one needed for the node module to run.

scottnonnenberg:
Fixes #107. Takes the package size down by something like 600k.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -4,3 +4,12 @@ npm-debug.log
 coverage
 .gitignore
 .DS_Store
+__tests__
+docs
+src
+.travis.yml
+.npmignore
+.eslintrc.js
+.eslintignore
+.babelrc
+yarn.lock


.npmignore
```diff
@@ -4,3 +4,12 @@ npm-debug.log
 coverage
 .gitignore
 .DS_Store
+__tests__
```
ljharb: Absolutely not. Tests should be shipped with a package.

.npmignore
```diff
@@ -4,3 +4,12 @@ npm-debug.log
...
+docs
+src
```
ljharb: as should the original source.

.npmignore
```diff
@@ -4,3 +4,12 @@ npm-debug.log
 coverage
 .gitignore
 .DS_Store
+__tests__
+docs
```
ljharb: Shipping docs with a package allows them to be accessed offline later.

.npmignore
```diff
@@ -4,3 +4,12 @@ npm-debug.log
...
+src
+.travis.yml
+.npmignore
+.eslintrc.js
```
ljharb: witholding these files means `npm explore foo && npm install && npm test` will not work.

scottnonnenberg:
`docs/` is relatively large at ~100k, but I suppose it's worth putting back to support offline documentation.

`src/` is the biggest impact since just three folders add up to 300k of data: `src/util/attributes`, `src/util/implicitRoles`, and `src/rules`

`__tests__/` isn't very big, but it would require `src/` to work. And it seems to me that the small number of users who try to `npm test` on an installed node module is vanishingly small compared to the number of users who are hit with a large install size.

The reason I'm here, why I requested these changes in the first place, is that `eslint-plugin-jsx-a11y` is by far the biggest ESlint plugin I've ever seen: install size of nearly a megabyte.

Lastly, I haven't seen many node modules with original source and test files included. Do you have any big examples? I just checked a couple: `react`, `express`, `lodash`, `chai`, `sinon`, `mocha`, `istanbul`, etc. None of them have tests, original source, or documentation included. And I only found it in one other out of nine ESLint plugins.

ljharb:
eslint plugins are a dev dep. Disk space is effectively infinite and free, and your local npm cache should mean it only has to be downloaded once.

Examples of packages doing it right: `tape`, `qs`, `object-keys`, `es5-shim`, `es6-shim`, `eslint-config-airbnb`, `enzyme`, (all of which I have a hand in, of course, but the point stands), and even `npm` itself.

ljharb:
Put another way - if the disease is that this package is large, then don't treat the symptoms, treat the disease: let's make the actual repo smaller.

scottnonnenberg:
> "Disk space is effectively infinite and free"

Sadly, this is likely why more and more people are complaining about `npm install` time. I certainly wouldn't want all module authors thinking this way.

evcohen:
@scottnonnenberg thanks for posting this, btw. honestly those files were bundled in the package just because of oversight- definitely learned from this discussion, though 👍

PR #122, opened 2016-11-21
update instructions for using recommended rules

willisplummer:
👋  I was having trouble using the recommended rule set after following the instructions in the README, and this is what ended up resolving the issue.


evcohen:
LGTM. thanks for fixing this!

PR #124, opened 2016-12-12
Explicitly distinguish interactive from non-interactive elements and roles

jessebeach:
## Motivation

A tabindex should only be required on elements with an interactive role:

```
[
  'button',
  'checkbox',
  'link',
  'menuitem',
  'menuitemcheckbox',
  'menuitemradio',
  'option',
  'radio',
  'spinbutton',
  'tab',
  'textbox',
]
```

There are legitimate use cases for placing an `onClick` on an element that should **not** have a tabIndex: a dialog screen or a container that catches bubbled clicks.

As I was codemodding violations of `onclick-has-focus` I found that I was adding `tabIndex={-1}` to hundreds of JSX fragments in order to satisfy the rule. These elements has roles like `presentation` or `listitem`.

Putting a tabindex on an element that shouldn't have it is just as bad as not having a tabindex on an element that should have it. I fear that the onclick-has-focus rule could incorrectly nudge developers to add tabindexes to elements more than they should. It will teach a bad habit. 

## Solution

For this reason, I'm proposing adding the `isInteractiveRole` module.

`onclick-has-focus` now makes the following distinctions:

### Valid cases

```
<div onClick={someHandler} />;
<a onClick="showNextPage();">Next page</a>
<header onClick={someHandler} />
```

### Invalid cases

```
<span role="button" onClick={someHandler} />
<div role="link" onClick={someHandler} />
<div role="textbox" onClick={someHandler} />
```

I modified `onclick-has-focus` to only recommend a tabindex on elements with an interactive role. The invalid cases above will trigger `onclick-has-role`. After adding an interactive role, they will then trigger `onclick-has-focus`.

### Before the change

I ran `onclick-has-role` and `onclick-has-focus` over a large codebase with 28750 React modules. This resulted in the following violations.

Rule | Violation count
------------ | -------------
`onclick-has-role` | 1286
`onclick-has-focus` | 1241

Almost every element that violates `onclick-has-role` also violates `onclick-has-focus`. But when I started writing a codemod to fix them, I began to notice that `onclick-has-role` was catching on elements that catch click bubbling. 

```
<div onClick={someHandler}>
  <button>
</div>
```

These elements should not have focus. The correct solution to this violation is:

```
<div role="presentation" onClick={someHandler}>
  <button>
</div>
```

not

```
<div role="presentation" tabIndex={-1} onClick={someHandler}>
  <button>
</div>
```

Because developers are going to incorrectly add `tabIndex` with a value of zero (as the warning recommends) and we'll have incoherent tab stops all over a page.

### After the change

I ran `onclick-has-role` and the updated `onclick-has-focus` over the same codebase of 28750 React modules. This resulted in the following violations.

Rule | Violation count
------------ | -------------
`onclick-has-role` | 1286
`onclick-has-focus` | 31

That's 1210 ambiguous violations of `onclick-has-focus` that have been eliminated. Yes, there are certainly more violations of `onclick-has-focus` within the violations of `onclick-has-role`, but I can confidently fix those and then deal with the `onclick-has-focus` violations after.

Here's an example of one of the 31 violations:

```
<div
  aria-label={'Increment value'}
  role="button"
  onClick={this._increment}>
```

**These are exactly the type of narrowly-targeted violations that can be easily remediated via codemod.** Additionally, I can now propose a rule that looks for elements with a `tabIndex` that should not have one -- non-interactive elements.

## Changes in this PR

1. Added `isInteractiveRole` util
1. Augmented the roles.json with an `interactive` field.
1. Augmented the DOM.json with an `interactive` field.
1. Added tests for `isInteractiveElement` and `isInteractiveRole`.
1. Added mocking for `JSXElement` and `JSXAttribute`.
1. Added `babel-plugin-transform-object-rest-spread` to `package.json`

### New error message for `onclick-has-focus`

'An non-interactive element with an onClick handler and an interactive role must be focusable. Either set the tabIndex property to a valid value (usually 0) or use an element type which is inherently focusable such as `button`.'


ljharb:
How often do you have a container implemented that catches bubbled clicks? In any app, I'd assume that number is always 1. Are there other use cases besides that to have `onClick` on an element that should not have a tabIndex?

jessebeach:
@ljharb, I've got a codebase with 26,000+ React components. `onclick-has-focus` and `onclick-has-role` triggered in about 5% of them, respectively. I'll get you some numbers before my changes and after.

lencioni:
>  In any app, I'd assume that number is always 1.

@ljharb why do you assume this?

ljharb:
@lencioni i'd assume that throughout an app, the "container that listens for clicks" would be a single component that's required/imported wherever it's used - ie, i don't understand why it'd ever be implemented more than once, even if it's used thousands of times.

jessebeach:
@lencioni There's theory and then there's practice :) I'll get some numbers and make my case for the change.

jessebeach:
@ljharb, @lencioni: I updated the summary with a summary, motivation and list of changes.

ljharb:
@jessebeach it sounds like your primary use case is `onClick={() => null}` to prevent click bubbling - can i ask why not, on the actual clickable elements, add `onClick={(e) => { e.stopPropagation(); }}`?

src/util/isInteractiveRole.js
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * isInteractiveRole is a Logical Disjunction:
+ * https://en.wikipedia.org/wiki/Logical_disjunction
+ * The JSX element does not have a tagName or it has a tagName and a role
+ * attribute with a value in the set of non-interactive roles.
+ */
+const isInteractiveRole = (tagName, attributes) => {
+  // Do not test higher level JSX components, as we do not know what
+  // low-level DOM element this maps to.
+  if (Object.keys(DOMElements).indexOf(tagName) === -1) {
```
evcohen: does the DOM element that contains the role attribute have any impact on the interactivity of the role? if not, i think this check can be removed.

src/util/isInteractiveElement.js
```diff
@@ -3,22 +3,27 @@ import getTabIndex from './getTabIndex';
 import DOMElements from './attributes/DOM.json';
 
 // Map of tagNames to functions that return whether that element is interactive or not.
-const interactiveMap = {
+
+const pureInteractiveElements = {};
+Object.keys(DOMElements)
+  .filter(name => DOMElements[name].interactive === true)
```
evcohen: nit- would prefer reduce here

```
const pureInteractiveElements = Object.keys(DOMElements)
  .filter(name => DOMElements[name].interactive === true)
  .reduce((interactiveElements, name) => {
    interactiveElements[name] = () => true;
    return interactiveElements;
  }, {})
```

package.json
```diff
@@ -29,7 +29,8 @@
     "babel-cli": "^6.14.0",
     "babel-core": "^6.14.0",
     "babel-eslint": "^7.0.0",
-    "babel-jest": "^18.0.0",
+    "babel-jest": "^17.0.1",
```
evcohen: i think you can bump this back up to 18, may have just missed in a merge/rebase

__mocks__/JSXElementMock.js
```diff
@@ -0,0 +1,13 @@
+export default function JSXElementMock (tagName, attributes) {
```
evcohen: thanks for adding these mocks! 😄

__mocks__/genInteractives.js
```diff
@@ -0,0 +1,80 @@
+import JSXAttributeMock from 'JSXAttributeMock';
+import JSXElementMock from 'JSXElementMock';
+import DOMElements from '../src/util/attributes/DOM.json';
+import roles from '../src/util/attributes/role.json';
+
+const pureInteractiveElements = {};
+Object.keys(DOMElements)
+  .filter(name => DOMElements[name].interactive === true)
+  .forEach(name => pureInteractiveElements[name] = []);
```
evcohen: same: re `reduce` over `forEach`.

__mocks__/genInteractives.js
```diff
@@ -0,0 +1,80 @@
...
+  ],
+  input: [
+    {prop: 'type', value: 'text'}
+  ],
+};
+
+const pureNonInteractiveElementsMap = {};
+Object.keys(DOMElements)
+  .filter(name => !DOMElements[name].interactive)
+  .forEach(name => pureNonInteractiveElementsMap[name] = []);
```
evcohen: `reduce`

__mocks__/genInteractives.js
```diff
@@ -0,0 +1,80 @@
...
+  role => roles[role].interactive === true
+);
+
+const nonInteractiveRoles = Object.keys(roles).filter(
+  role => roles[role].interactive === false
+);
+
+export function genInteractiveElements () {
+  const elements = [];
+  for (const name in interactiveElementsMap) {
```
evcohen: would prefer to just use `map` here instead of iterating and pushing result

jessebeach:
> @jessebeach it sounds like your primary use case is onClick={() => null} to prevent click bubbling - can i ask why not, on the actual clickable elements, add onClick={(e) => { e.stopPropagation(); }}

@ljharb derp derp, sorry, my toy examples suggested behavior that I did not intend to suggest. I'll update them in a moment.

In my case, catching clicks on bubble is just a pattern that devs use and it's not easy to dissuade them of that habit. I've seen it on list components where the list items are controlled by different teams. The list wants to perform an action on listitem click, but it wasn't architected in a way that passes down a handler through props. It's an anti-pattern, sure, but it's an existing pattern and I don't want to browbeat developers with a tabindex eslint rule into rearchitecting their product :) I especially don't want them putting `tabIndex={0}` on a list element and making it focusable in addition to the list items being focusable. 

I do like forcing the developer to put a role on these click catching elements. It's a signal that they're doing something outside normal practice. We can recommend in the `onclick-has-role` docs that if their element is catching bubbled clicks, they should apply `role="presentation none"`.

jessebeach:
@evcohen, thank you for the code review, I'll go through and make the updates.

> but do you have any other resources that suggest this? just for the sake of learning!

I had a conversation with @mcking65 about this rule before I ventured to start this PR. I'd been codemodding fixes for violations and I found myself putting in `tabIndex={-1}` a lot on elements that really didn't need it. @mcking65 and I discussed the relative merit and danger of nudging developers to add tabindex to their code. We concluded that incorrectly overapplying tabindex would lead to a detrimental experience. @mcking65 is an author of the [ARIA Authoring Practices Guide](https://github.com/w3c/aria-practices). A lot of the work we're currently undertaking is to reduce the number of tabstops on a page; to make tabstops an entry point into widgets so that a page can be traversed quickly through the major pieces. Arbitrary tab stops add clutter to be traversed through and they suggest operability where none may exist if the tabstop manifests from an incorrectly applied tabindex.

ljharb:
@jessebeach I think lint rules are exactly the right place to point out to developers when code is architected poorly.

I do agree that the rule should be adjusted so it doesn't recommend adding tabIndexes to places where none should exist - but I don't think it should allow anti-patterns nor make them easy - that's what inline override comments are for.

evcohen:
@jessebeach awesome, thanks for working on & caring about a11y. if this is the case, we should also add a warning when detecting `tabIndex={-1}` here and suggest using role properties instead

src/util/isInteractiveRole.js
jessebeach: It does not. The role attribute more or less overrides the inherent role of a semantic element. But I don't think we can express with certainty that a higher order JSX component with an interactive role won't have a tabindex declared in its composition chain. Removing this check introduces the possibility of false positive matches. Although I'm disappointed we can't be more opinionated, I agree with jsx-a11y approach of bailing on higher order components in most cases in order to provide a higher quality lint signal even if it means that signal might miss some cases. Noisy lint rules get turned off eventually :)

jessebeach:
> if this is the case, we should also add a warning when detecting tabIndex={-1} here and suggest using role properties instead

I'm going to propose a couple new rules dealing with `tabIndex` in the near future. Currently I'm going through the existing rules, running them against FBs codebase (I have reports every month going back to 10/15/2015), codemodding the violations and then turning them on.

src/util/isInteractiveRole.js
evcohen: that makes sense, and also we can't be sure that the role prop on a higher order component actually maps correctly to the role prop on the low level DOM element. sounds good to me!

src/util/isInteractiveRole.js
```diff
@@ -0,0 +1,39 @@
...
+
+  // If value is undefined, then the role attribute will be dropped in the DOM.
+  // If value is null, then getLiteralAttributeValue is telling us that the
+  // value isn't in the form of a literal
+  if (value === undefined || value === null) {
+    return false;
+  }
+
+  const normalizedValues = String(value).toUpperCase().split(' ');
+  const validRoles = Object.keys(roles).filter(role => roles[role].interactive === true);
```
evcohen: one last nit- let's bring this out of the function so it doesn't get computed on every pass.

```
const VALID_ROLES = ...

const isInteractiveRole = (tagName, attributes) => {
  ...
}
```

mcking65:
Sorry I am a little late to the party here.

Here are some relevant considerations. I haven't looked at the rule logic close enough to know if/how they can be accounted for in the logic of this rule.

`<span role="button" onClick={someHandler} />`
`<div role="link" onClick={someHandler} />`

These should not have tabindex if they are contained in an element, e.g., a toolbar, that uses activedescendant to manage focus. This is also the case for option elements in a listbox, menuitem elements in a menu, treeitem elements in a tree, gridcell elementss in a grid, etc.

However, if you have a link with an href, an HTML button, or an HTML input  in a toolbar that manages focus with aria-activedescendant, `tabindex="-1"` would be required.

For this rule, there may be more roles to consider from the set of 36 widget roles plus there is also the one structure that commonly manages focus and supports activedescendant -- toolbar. If you ask me, and many other working group members, toolbar should be a composite widget. While in the specification toolbar is a structure, in practice it is best if it manages focus like a composite. 

Following are all the [widget roles](http://w3c.github.io/aria/aria/aria.html#widget_roles) 
with my comments wrt the above.

Widget roles:

1. alert: Usually should not be focusable.
2. alertdialog: should never be focusable.
3. button: focusable with tabindex or activedescendant
4. checkbox: focusable with tabindex or activedescendant
5. dialog: should never be focusable
6. gridcell: focusable with tabindex or activedescendant
7. link: focusable with tabindex or activedescendant
8. log: Usually should not be focusable.
9. marquee: Usually should not be focusable.
10. menuitem: focusable with tabindex or activedescendant
11. menuitemcheckbox: focusable with tabindex or activedescendant
12. menuitemradio: focusable with tabindex or activedescendant
13. option: focusable with tabindex or activedescendant
14. progressbar: Focusability may be helpful in some contexts.
15. radio: focusable with tabindex or activedescendant
16. scrollbar: Focusability may be helpful in some contexts.
17. searchbox: Always requires DOM focus
18. slider: focusable with tabindex or activedescendant
19. spinbutton: Always requires DOM focus.
20. status: Usually should not be focusable.
21. switch: focusable with tabindex or activedescendant
22. tab: focusable with tabindex or activedescendant
23. tabpanel: Focusability may be helpful in some contexts.
24. textbox: Always requires DOM focus.
25. timer: Usually should not be focusable.
26. tooltip: Should never be focusable.
27. treeitem: focusable with tabindex or activedescendant

Elements with the widget roles that always require DOM focus require tabindex.

Elements with a widget role that can be focusable either via tabindex or aria-activedescendant require tabindex if any of the following is true:

1. The element is Not a descendant of a composite widget.
2. The element is a descendant of a composite widget that DOES NOT have aria-activedescendant.
3. `tabindex="-1"` is required if the element is natively focusable AND is a descendant of a composite widget that does have aria-activedescendant.

Said another way, tabindex is not required for elements with a widget role whose focus may be managed via aria-activedescendant if *BOTH* the following are true:

1. The element is not natively focusable like an HTML button, input, or link with an href.
2. The element is a a descendant of a composite widget or toolbar that has aria-activedescendant.

An Element with any of the following composite widget roles requires tabindex if it manages focus of descendants using aria-activedescendant. However, an element with one of these composite roles MUST NOT have tabindex if it does not use activedescendant. That is, if it uses a roving tabindex to manage focus of descendants, the composite widget itself should not be focusable.

Note, I have added toolbar, which is not 
listed as a composite in the spec, but it is best if it is treated as one.

1. combobox
2. grid
3. listbox
4. menu: If it is a submenu, does not need either activedescendant or tabindex
5. menubar
6. radiogroup
7. tablist
8. toolbar
9. tree
10. treegrid

That is, `<div role="grid">` should not have tabindex whereas `<div role="grid" aria-activedescendant="ID">` should have `tabindex="0"`.

jessebeach:
Thank you for the feedback @mcking65! 

@evcohen give me a little time to parse @mcking65's comments before committing. I think I can address his concerns with some documentation additions. We can't know through static analysis if a button will be rendered in a toolbar that is managing active descendants, but we can at least require a tabindex and point to good documentation. Maybe I should break out the cases like button and link into a separate rule that points to more specific documentation.

evcohen:
thank you @mcking65!! really helpful info here. 

haha @jessebeach, also going need some time to parse his comments so no rush. really, really appreciate all of the work your putting into making the web more accessible for all

jessebeach:
> These should not have tabindex if they are contained in an element, e.g., a toolbar, that uses activedescendant to manage focus. This is also the case for option elements in a listbox, menuitem elements in a menu, treeitem elements in a tree, gridcell elementss in a grid, etc.

@mcking65 Is `tabIndex="-1"` acceptable in this case? We can detail in documentation the cases where 0 should be applied and where -1 should be applied, but we won't be able to determine if either is correct -- a human will need to do that at this level of analysis.

> However, if you have a link with an href, an HTML button, or an HTML input in a toolbar that manages focus with aria-activedescendant, tabindex="-1" would be required.

This is not something we can determine through lint. This requires integration testing.

Regarding `aria-activedescendant`. The `onclick-has-focus` rule should never catch on an element whose focus is being managed by a parent composite, because the child elements should never actually be focused in the document and thus, an `onClick` handler would never catch. 

I opened #133 to address `aria-activedescendant`.

> tabindex="-1" is required if the element is natively focusable AND is a descendant of a composite widget that does have aria-activedescendant.

This is impossible to validate via lint. It's an integration testing concern.

mcking65:
@mcking65 wrote:

> These should not have tabindex if they are contained in an element, e.g., a toolbar, that uses activedescendant to manage focus. This is also the case for option elements in a listbox, menuitem elements in a menu, treeitem elements in a tree, gridcell elementss in a grid, etc.

@jessebeach asked:

> Is  tabIndex="-1"  acceptable in this case? We can detail in documentation the cases where 0 should be applied and where -1 should be applied, but we won't be able to determine if either is correct -- a human will need to do that at this level of analysis.

Yes, having tabindex="-1" on elements referenced by activedescendant is OK; it is just unnecessary. Of cource, tabindex of 0 or greater would be a problem.

jessebeach:
Keeping it fresh on master.

evcohen:
@jessebeach is this still WIP or ready to ship? want to get this in the next major release 😄

jessebeach:
Ready to ship, other than the yarn.lock conflict, which I'm fixing right now.

PR #126, opened 2016-12-14
Fixing aria-invalid=true and aria-invalid=false validation

### LINKED ISSUE #125: aria-invalid={true/false} incorrectly fails check
When specifying aria-invalid="false" or aria-invalid="true", the value gets converted into a boolean by getLiteralPropValue(), which fails the token check because it's looking for a "string".

fracmak:
Fixes #125

aria-invalid's prop value was being converted to a boolean and hence failing the token check which always forced all tokens to be strings. This relaxes that constraint allowing non-string values and replacing the "true" and "false" aria-invalid tokens with `true` and `false`

### DIFF
--- a/src/util/attributes/ARIA.json
+++ b/src/util/attributes/ARIA.json
@@ -47,7 +47,7 @@
   },
   "ARIA-INVALID": {
     "type": "token",
-    "values": [ "grammar", "false", "spelling", "true" ]
+    "values": [ "grammar", false, "spelling", true ]
   },
   "ARIA-LABEL": {
     "type": "string"
--- a/__tests__/src/rules/aria-proptypes-test.js
+++ b/__tests__/src/rules/aria-proptypes-test.js
@@ -123,6 +123,12 @@ ruleTester.run('aria-proptypes', rule, {
     { code: '<div aria-sort={`other`} />', parserOptions },
     { code: '<div aria-sort={foo} />', parserOptions },
     { code: '<div aria-sort={foo.bar} />', parserOptions },
+    { code: '<div aria-invalid={true} />', parserOptions },
+    { code: '<div aria-invalid="true" />', parserOptions },
+    { code: '<div aria-invalid={false} />', parserOptions },
+    { code: '<div aria-invalid="false" />', parserOptions },
+    { code: '<div aria-invalid="grammar" />', parserOptions },
+    { code: '<div aria-invalid="spelling" />', parserOptions },
 
     // TOKENLIST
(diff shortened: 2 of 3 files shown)


src/util/attributes/ARIA.json
```diff
@@ -47,7 +47,7 @@
   },
   "ARIA-INVALID": {
     "type": "token",
-    "values": [ "grammar", "false", "spelling", "true" ]
+    "values": [ "grammar", false, "spelling", true ]
```
ljharb: technically doesn't this make it a breaking change?

fracmak: the original functionality never worked, so not really

ljharb: so you're saying previously if i passed `aria-invalid="true"` it didn't work?

fracmak: correct, it said it was an invalid option

ljharb: thanks for confirming!

ljharb: See #127 - looks like "True" and "False" passed previously, which indeed makes this a breaking change.

evcohen: This shouldn't be a breaking change - I think that means there's a bug in the getLiteralPropValue code that doesn't convert `"True"` to `true`

evcohen: @ljharb this is a [downstream bug](https://github.com/evcohen/jsx-ast-utils/blob/master/src/values/Literal.js#L10), this should be normalized, i believe so that the literal `"True"` resolves to `true`

ljharb: The comment by that user implies differently, but maybe they're just testing the function directly. Just being cautious :-)

evcohen: haha appreciate the concern, i can update `jsx-ast-utils` so that this does not constitute a breaking change. in general, how do you deal with breaking changes that are caused by outside dependencies?

ljharb: Any observable change that leaks through a project from its deps is a change, so i either encapsulate it away or I take the major bump.

evcohen: cool, thanks! here's [PR](https://github.com/evcohen/jsx-ast-utils/pull/15) for patching `jsx-ast-utils`

evcohen: merged and published [v1.3.5](https://github.com/evcohen/jsx-ast-utils/releases/tag/1.3.5)

PR #129, opened 2016-12-21
Explain use of role presentation in onclick-has-role.md

jessebeach:
I added a paragraph to onclick-has-role.md to explain the use of role presentation in cases where an interactive role would not be appropriate.


jessebeach:
Yes, I rebased on evcohen/master before pushing the branch.

Issue #133, opened 2016-12-21
[new rule] activedescendant-is-allowed to limit application of aria-activedescendant to composite roles 


jessebeach:
valid:

```
<div role="combobox" aria-activedescendant="js_1">{this.props.children}</div>
<div role="grid" aria-activedescendant="js_1">{this.props.children}</div>
<div role="listbox" aria-activedescendant="js_1">{this.props.children}</div>
<div role="menu" aria-activedescendant="js_1">{this.props.children}</div>
<div role="menubar" aria-activedescendant="js_1">{this.props.children}</div>
<div role="radiogroup" aria-activedescendant="js_1">{this.props.children}</div>
<div role="tablist" aria-activedescendant="js_1">{this.props.children}</div>
<div role="toolbar" aria-activedescendant="js_1">{this.props.children}</div>
<div role="tree" aria-activedescendant="js_1">{this.props.children}</div>
<div role="treegrid" aria-activedescendant="js_1">{this.props.children}</div>
```

invalid:

```
<div role="article" aria-activedescendant="js_1">{this.props.children}</div>
```

evcohen:
seems like this is available for [multiple roles](https://www.w3.org/WAI/PF/aria/states_and_properties#aria-activedescendant): `composite`, `group`, and `textbox` ?

jessebeach:
Composite is abstract. Definitely group and textbox. Nice catch.

mcking65:
textbox is a special case because it allows aria-activedescendant to refer to an element that is a descendant of an element that is controled by the textbox rather than an element that is a direct descendant of the textbox. This is one of the changes we made in aria 1.1 to support a combobox pattern that has a combobox element that contains a textbox. In that case, the textbox must have both activedescendant and DOM focus.

The pattern is:

```
<div role="combobox">
  <input type="text" aria-controls="l" aria-activedescendant="i1">
  <ul id="l" role="listbox">
   <li role="presentation">
      <a id="i1" tabindex="-1" href="#">item 1</a>
    </li>
  </ul>
</div>
```

jessebeach:
@mcking65 I'll make note of this in the rule docs. It doesn't change the implementation of the rule.

JoshuaKGoldberg:
I think this was fixed by #135?

PR #134, opened 2016-12-22
Add aria-activedescendant-has-tabindex rule

jessebeach:
Related to #133 .

This rule enforces that elements with aria-activedescendant are tabbable.


jessebeach:
Based on incorrect history.

PR #135, opened 2016-12-22
Add aria-activedescendant-has-tabindex rule

jessebeach:
Related to #133 .

This rule enforces that elements with aria-activedescendant are tabbable.


src/rules/aria-activedescendant-has-tabindex.js
```diff
@@ -0,0 +1,66 @@
...
+      if (getProp(attributes, 'aria-activedescendant') === undefined) {
+        return;
+      }
+
+      const type = elementType(node);
+      const tabIndex = getTabIndex(getProp(attributes, 'tabIndex'));
+
+      // Do not test higher level JSX components, as we do not know what
+      // low-level DOM element this maps to.
+      if (DOMElementKeys.indexOf(type) === -1) {
```
evcohen: let's bring this early return before getting tabIndex (line 38) to avoid unnecessary processing

src/rules/aria-activedescendant-has-tabindex.js
```diff
@@ -0,0 +1,66 @@
...
+        isInteractiveElement(type, attributes)
+        && (
+          tabIndex === undefined
+          || tabIndex >= 0
+        )
+      ) {
+        return;
+      }
+
+      if (getTabIndex(getProp(attributes, 'tabIndex')) >= 0) {
```
evcohen: can just use the `tabIndex` variable here from line 38 instead of re-calling these functions

src/rules/aria-activedescendant-has-tabindex.js
```diff
@@ -0,0 +1,69 @@
...
+        isInteractiveElement(type, attributes)
+        && (
+          tabIndex === undefined
+          || tabIndex >= 0
+        )
+      ) {
+        return;
+      }
+
+      if (tabIndex >= 0) {
```
evcohen: hm, i'm totally cool with this since it communicates intent, but it seems like anything with `tabIndex >=0` passes regardless if the element is interactive or not. your comment says `(allowing for positive, even though that is not ideal)` - instead, do we want to add configuration for this rule that is like strict mode, where `tabIndex` cannot be greater than 0 for interactive elements?

jessebeach: My thought here is that `tabindex-no-positive` can enforce no tabindices greater than zero. `aria-activedescendant-has-tabindex` is satisfied if `tabIndex` exists with a positive value when `aria-activedescendant` is present.

evcohen: ah this makes sense! i wonder if we can see what configuration a user has for another rule and use it to determine logic in another rule?

ljharb: Eslint doesn't make that possible, no

ljharb:
It'd be great if this could be rebased so as to remove all merge commits.

jessebeach:
Rewritin' history like McFly.

Issue #29, opened 2016-04-16
Find rules from aXe


lencioni:
The aXe browser extensions are used to tell you about accessibility problems on the pages you visit. I bet there are some rules in there that we can use as inspiration for this project.

It looks like they have built an engine that consumes JSON to do the rules. Here are the rules JSON files https://github.com/dequelabs/axe-core/tree/master/lib/rules

Here's their site that lists the rules they have: https://dequeuniversity.com/rules/axe/1.1/

lencioni:
Here are a few that I found that I think might be worth considering:

- [ ] https://dequeuniversity.com/rules/axe/1.1/object-alt object-has-alt
- [ ] https://dequeuniversity.com/rules/axe/1.1/area-alt area-has-alt
- [ ] https://dequeuniversity.com/rules/axe/1.1/frame-title frame-has-title
- [ ] https://dequeuniversity.com/rules/axe/1.1/empty-heading heading-has-content
- [ ] https://dequeuniversity.com/rules/axe/1.1/input-image-alt input-image-has-alt
- [ ] https://dequeuniversity.com/rules/axe/1.1/html-lang html-has-lang
- [ ] https://dequeuniversity.com/rules/axe/1.1/valid-lang lang
- [ ] https://dequeuniversity.com/rules/axe/1.1/link-name link-has-text
- [ ] https://dequeuniversity.com/rules/axe/1.1/scope scope
- [ ] https://dequeuniversity.com/rules/axe/1.1/server-side-image-map image-map-no-server-side

These are probably not super useful, at least in the context of React, but they should be implementable

- [ ] https://dequeuniversity.com/rules/axe/1.1/blink no-blink
- [ ] https://dequeuniversity.com/rules/axe/1.1/marquee no-marquee
- [ ] https://dequeuniversity.com/rules/axe/1.1/meta-refresh meta-no-refresh
- [ ] https://dequeuniversity.com/rules/axe/1.1/meta-viewport meta-viewport-scalable

I think we already have this rule, but there might be some things to look at in here: https://dequeuniversity.com/rules/axe/1.1/aria-required-attr

lencioni:
I think it might make sense to rename img-has-alt to has-alt, alt, alt-text, or something more general like that, and just add `object`, `area`, and `input-image` (need to special case this one in the code I think) to the list of things it checks for.

evcohen:
[...]

PR #137, opened 2016-12-23
[new] - Add iframe-has-title rule.

evcohen:
From #29


__tests__/src/rules/frame-has-title-test.js
```diff
@@ -0,0 +1,48 @@
...
+
+const expectedError = {
+  message: 'Frame elements (frame, iframe) must have a unique title property.',
+  type: 'JSXOpeningElement',
+};
+
+ruleTester.run('html-has-lang', rule, {
+  valid: [
+    { code: '<div />;', parserOptions },
+    { code: '<frame title="Unique title" />', parserOptions },
```
ljharb: do people still use `<frame>` elements? do they even work in browsers anymore? i know framesets don't

evcohen: eh, doesn't seem so: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/frame

evcohen: will change to just iframe

src/rules/iframe-has-title.js
```diff
@@ -0,0 +1,43 @@
...
+    JSXOpeningElement: (node) => {
+      const type = elementType(node);
+
+      if (type && type !== 'iframe') {
+        return;
+      }
+
+      const title = getPropValue(getProp(node.attributes, 'title'));
+
+      if (title) {
```
ljharb: does this mean that the following will fail the rule?
```js
title=""
title={false}
title={''}
title={``}
```
If so, it'd be nice to include this edge case  in the docs and tests.

evcohen: yes these will fail - `title={true}` will pass, which is unfortunate. we can use getLiteralPropValue here and bypass the boolean edge case, but then we'd be failing on `title={foo}` where `foo` is some variable. we can also explicitly check `title && title !== true` if that seems more appropriate?

ljharb: I think that seems reasonable; I mainly want it documented and to have test cases :-)

PR #139, opened 2016-12-24
[new][dev] - Create script to easily scaffold new rules.

evcohen:
Was getting annoying copy and pasting other files to create new rules, so put together a quick generator.

Can run with :
```bash
$ npm run create -- rule-name --author="Name" --description="Basic rule description"
```

also upgraded dependencies


jessebeach:
#142 extends this PR to add some behavior that adds the new rule to index.js. That caught me out the first time I created a new rule.

jessebeach:
I can adjust the output settings so that jscodeshift isn't so verbose when it runs.

PR #140, opened 2016-12-24
[new] - Consolidate no-marquee and no-blink into no-distracting-elements rule.

evcohen:
From #29


evcohen:
@ljharb we can do that since we have `no-marquee` as well. would be breaking change, though

ljharb:
I guess we could add this one now, and then in the next major, remove both in favor of a more generic one

evcohen:
well if #124 is breaking then it makes sense to do right now

ljharb:
Maybe `no-distracting-elements`, make blink and marquee both configurable but included by default? That's a bit more targeted but still extensible

src/rules/no-distracting-elements.js
```diff
@@ -0,0 +1,52 @@
...
+    docs: {},
+    schema: [schema],
+  },
+
+  create: context => ({
+    JSXOpeningElement: (node) => {
+      const options = context.options[0] || {};
+      const elementOptions = options.elements || DEFAULT_ELEMENTS;
+      const type = elementType(node);
+      const distractingElement = elementOptions.find((element) => {
```
jessebeach:
```
const distractingElement = elementOptions.find(el => el === type);
```

Just return a boolean, rather than the element itself, which'll get cast to a truthy value.

evcohen: i did `find` and returned the element to use in the error message so that the message was more specific. is there a better way to do this?

ljharb: `find` takes a predicate - ie, a function that only returns true or false. You're returning a truthy value. @jessebeach's example is identical but with a more efficient and correct callback.

ljharb: Correct, just like filter, some, or every

PR #141, opened 2016-12-25
Get coveralls to 100%

jessebeach:
Trying to get Coveralls as close to 100% as I can.

### DIFF
--- a/__mocks__/IdentifierMock.js
+++ b/__mocks__/IdentifierMock.js
@@ -1,0 +1,6 @@
+export default function IdentifierMock(ident) {
+  return {
+    type: 'Identifier',
+    name: ident,
+  };
+};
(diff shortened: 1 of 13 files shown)


jessebeach:
And THAT'S 100%.

💯

jessebeach:
> Why the mocks? Mocks make tests more brittle - could we get the coverage by providing accurate fixture code to test the rules against?

If we use fixtures to test this util, then we need to run a rule, right? My thought here was we can either have a mock rule (not a bad idea), or a few AST mocks and just call the function directly with some AST mocks and avoid the ESLint test runner. The AST definitions of JS are stable and the mocks are simple. Like a well-dressed bowline, it'll do under tension.

jessebeach:
@evcohen might we get this merged? I promise to fix anything I've broken and I'd love to see the effect on the other PRs in terms of coverage.

evcohen:
@jessebeach haha yes sorry, i've been on vacation with my family 😄 this LGTM, thanks for doing this! one quick thing- can you update the yarn.lock file when adding a dep. i think you can just run `yarn add to-ast`

PR #142, opened 2016-12-27
[new][dev] Add the new rule to the index, extends #139

jessebeach:
Uses JSCodeShift to add the new rule to the index file as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,15 @@ You can also enable all the recommended rules at once. Add `plugin:jsx-a11y/reco
 - [scope](docs/rules/scope.md): Enforce `scope` prop is only used on `<th>` elements.
 - [tabindex-no-positive](docs/rules/tabindex-no-positive.md): Enforce `tabIndex` value is not greater than zero.
 
+## Creating a new rule
+
+If you are developing new rules for this project, you can use the `create-rule`
+script to scaffold the new files.
+
+```
+$ ./scripts/create-rule.js my-new-rule
+```
+
 ## License
 
 eslint-plugin-jsx-a11y is licensed under the [MIT License](LICENSE.md).
--- a/scripts/create-rule.js
+++ b/scripts/create-rule.js
@@ -14,6 +14,14 @@ const rulePath = path.resolve(`src/rules/${ruleName}.js`);
 const testPath = path.resolve(`__tests__/src/rules/${ruleName}-test.js`);
 const docsPath = path.resolve(`docs/rules/${ruleName}.md`);
 
+const jscodeshiftJSON = require('jscodeshift/package.json');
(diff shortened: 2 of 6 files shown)


scripts/create-rule.js
```diff
@@ -29,3 +30,24 @@ const docBoilerplate = docBoilerplateGenerator(ruleName);
 fs.writeFileSync(rulePath, ruleBoilerplate);
 fs.writeFileSync(testPath, testBoilerplate);
 fs.writeFileSync(docsPath, docBoilerplate);
+
+// Add the rule to the index
+exec([
+  './node_modules/jscodeshift/bin/jscodeshift.sh',
```
ljharb: could this use `require.resolve('jscodeshift/bin/jscodeshift.sh')`, or better, `path.join(require.resolve('jscodeshift'), require('jscodeshift/package.json').bin.jscodeshift)`? `node_modules` should ideally never be hardcoded anywhere.

src/index.js
```diff
@@ -15,7 +15,7 @@ module.exports = {
     'img-has-alt': require('./rules/img-has-alt'),
     'img-redundant-alt': require('./rules/img-redundant-alt'),
     'label-has-for': require('./rules/label-has-for'),
-    lang: require('./rules/lang'),
+    'lang': require('./rules/lang'),
```
ljharb: these are valid identifiers, why would they need to be quoted?

the airbnb linter rules require `consistent-as-needed` here, not `consistent`.

jessebeach: They don't need to be, but the string comparisons were being weird and it's late and I was being lazy :)

So, looking into it more, `'scope'` is a Literal and `scope` is an Identifier and of course the value accessor prop is different for the type.

ljharb: Gotcha - it seems prudent for any codemod in the repo to be robust against any potential object key used :-) can we make it support both?

README.md
```diff
@@ -111,6 +111,15 @@ You can also enable all the recommended rules at once. Add `plugin:jsx-a11y/reco
 - [scope](docs/rules/scope.md): Enforce `scope` prop is only used on `<th>` elements.
 - [tabindex-no-positive](docs/rules/tabindex-no-positive.md): Enforce `tabIndex` value is not greater than zero.
 
+## Creating a new rule
```
evcohen: 👍 looks good - just to clarify, i meant automating adding to the list of rules in the README in my previous comment, but should be done in a separate PR.

jessebeach: Derp, ya, I totally misunderstood that. There's a project called remark (http://remark.js.org/) for parsing markdown, but jscodeshift is wedded to recast. It'll take a little time to abstract out the re-casters from jscodeshift, making this kind of codemodding of markdown possible.

PR #144, opened 2016-12-29
Change `npm install` to `npm i` for consistency with previous usage.

ryanvanoss:


ljharb:
My personal preference would be to never use the shortcut in documentation - commands in docs should be googleable, and shortcuts are a personal choice but not what should be popularized.

ryanvanoss:
I don't have much of an opinion either way; I'm more for consistency than anything else. On one hand, shortcuts in docs can teach people things they may have otherwise missed. On the other, they can distract from the intent of the documentation.

@evcohen and fellow collaborators, I'll let you decide the preferred format, and update the PR as necessary.

evcohen:
@ryanvanoss think i would prefer `install` on both. let's leave teaching shortcuts to the npm documentation. also, can you add `yarn add eslint --dev` and `yarn add eslint-plugin-jsx-a11y --dev` to the docs as well? thanks!

PR #147, opened 2017-01-02
[new] - Add no-autofocus rule.

evcohen:
https://twitter.com/stevefaulkner/status/815884132874092544


src/index.js
```diff
@@ -55,6 +56,7 @@ module.exports = {
         'jsx-a11y/label-has-for': 'error',
         'jsx-a11y/mouse-events-have-key-events': 'error',
         'jsx-a11y/no-access-key': 'error',
+        'jsx-a11y/no-autofocus': 'error',
```
lencioni: Modifying recommended rules may break builds for people, so this part of it should probably be reserved for major releases.

evcohen: ah, i've been doing this for every new rule. good to know. thanks. this should be safe, though since next release will be major (due to #140)

__tests__/src/rules/no-autofocus-test.js
```diff
@@ -0,0 +1,47 @@
...
+
+const expectedError = {
+  message: 'The autoFocus prop should not be used, as it can reduce usability and accessibility for users.',
+  type: 'JSXAttribute',
+};
+
+ruleTester.run('no-autofocus', rule, {
+  valid: [
+    { code: '<div />;', parserOptions },
+    { code: '<div autofocus />;', parserOptions },
```
ljharb: Is there another rule that catches improperly React-cased HTML attrs? This seems like *something* should make it invalid :-)

src/index.js
ljharb: Def causing more warnings for people is a semver-major.

__tests__/src/rules/no-autofocus-test.js
evcohen: haha better suited for `eslint-plugin-react`? will send a PR!

lencioni: I think this rule handles that: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-unknown-property.md

The way eslint tests work is you are testing an individual rule, so even if there was one in this project that would flag this, it won't happen in this test.

ljharb: lol yes, thanks, just musing whilst reviewing on mobile. thanks for the link :-)

PR #150, opened 2017-01-04
[new] - Add no-redundant-roles rule

### LINKED ISSUE #149: Rule proposal: no-redunant-roles
https://www.w3.org/TR/html5/dom.html#aria-role-attribute - make sure dom elements don't have default role property set

evcohen:
Fixes #149 

https://www.w3.org/TR/html5/dom.html#aria-role-attribute


__tests__/src/rules/no-redundant-roles-test.js
```diff
@@ -0,0 +1,43 @@
+/* eslint-env jest */
```
lencioni: Why not put this in a `.eslintrc` file in the __tests__ directory?

src/index.js
```diff
@@ -59,6 +60,7 @@ module.exports = {
         'jsx-a11y/no-autofocus': 'error',
         'jsx-a11y/no-distracting-elements': 'error',
         'jsx-a11y/no-onchange': 'error',
+        'jsx-a11y/no-redundant-roles': 'error',
```
lencioni: This line is a semver major change. It might be a good idea to wait on this and bundle up a  few breaking changes at a time.

__tests__/src/rules/no-redundant-roles-test.js
```diff
@@ -0,0 +1,43 @@
...
+
+ruleTester.run('no-redundant-roles', rule, {
+  valid: [
+    { code: '<div />;', parserOptions },
+    { code: '<button role="main" />', parserOptions },
+    { code: '<MyComponent role="button" />', parserOptions },
+  ],
+  invalid: [
+    { code: '<button role="button" />', errors: [expectedError('button', 'button')], parserOptions },
+    { code: '<body role="DOCUMENT" />', errors: [expectedError('body', 'document')], parserOptions },
```
lencioni: Is it worth adding some tests that do string concatenation or template literals? Or maybe a ternary?

src/index.js
evcohen: next release will be major, so this is safe

PR #154, opened 2017-01-11
Flowify isInteractiveRole

jessebeach:
Flowify `isInteractiveRole`.

### DIFF
--- a/src/util/isInteractiveRole.js
+++ b/src/util/isInteractiveRole.js
@@ -43,7 +43,7 @@ const isInteractiveRole = (
 
   const normalizedValues: Array<string> = String(value).toUpperCase().split(' ');
   const isInteractive = normalizedValues.every(
-    (val: string) => VALID_ROLES.indexOf(val) > -1,
+    (val: string): boolean => VALID_ROLES.indexOf(val) > -1,
   );
 
   return isInteractive;
--- a/src/util/isInteractiveRole.js
+++ b/src/util/isInteractiveRole.js
@@ -2,12 +2,9 @@
  * @flow
  */
 import { getProp, getLiteralPropValue } from 'jsx-ast-utils';
+import type { JSXAttribute } from 'ast-types-flow';
 import DOMElements from './attributes/DOM.json';
 import roles from './attributes/role.json';
-
-import type {
-  JSXAttribute,
-} from 'ast-types-flow';
 
 const VALID_ROLES = Object.keys(roles)
   .filter(role => roles[role].interactive === true);
--- a/src/util/isInteractiveRole.js
+++ b/src/util/isInteractiveRole.js
@@ -9,7 +9,7 @@ import type {
   JSXAttribute,
 } from 'ast-types-flow';
 
-const VALID_ROLES: Array<string> = Object.keys(roles)
+const VALID_ROLES = Object.keys(roles)
   .filter(role => roles[role].interactive === true);
 /**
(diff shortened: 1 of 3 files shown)


src/util/isInteractiveRole.js
```diff
@@ -1,9 +1,15 @@
+/**
+ * @flow
```
ljharb: it's really really unfortunate/annoying that this comment is necessary for flow to type a file. Has there been any progress on fixing that?

src/util/isInteractiveRole.js
```diff
@@ -1,9 +1,15 @@
...
 import { getProp, getLiteralPropValue } from 'jsx-ast-utils';
 import DOMElements from './attributes/DOM.json';
 import roles from './attributes/role.json';
 
+import type {
+  JSXAttribute,
+} from 'ast-types-flow';
 
-const VALID_ROLES = Object.keys(roles)
+const VALID_ROLES: Array<string> = Object.keys(roles)
```
ljharb: Why can't this type be inferred, since `Object.keys` is a builtin?

src/util/isInteractiveRole.js
```diff
@@ -32,9 +41,9 @@ const isInteractiveRole = (tagName, attributes) => {
     return false;
   }
 
-  const normalizedValues = String(value).toUpperCase().split(' ');
+  const normalizedValues: Array<string> = String(value).toUpperCase().split(' ');
```
ljharb: similarly here, can't this be inferred to be an array of strings, since it's all builtins?

src/util/isInteractiveRole.js
```diff
@@ -32,9 +41,9 @@ const isInteractiveRole = (tagName, attributes) => {
...
   const isInteractive = normalizedValues.every(
-    val => VALID_ROLES.indexOf(val) > -1,
+    (val: string) => VALID_ROLES.indexOf(val) > -1,
```
ljharb: and since `normalizedValues` is already typed, why is the first argument of `every` not inferrable?

src/util/isInteractiveRole.js
jessebeach: It can be. I was being pedantic. Will remove.

src/util/isInteractiveRole.js
jessebeach: No, that's what we have at the moment.

ljharb: My personal opinion is that this is too high a cost to be worth adding flow.

evcohen: disagree - think adding a comment is trivial compared to the benefit of types and allows for gradual adoption.

ljharb: Config files allow for gradual adoption also; and I think the benefit of types is grossly overstated. They rarely hurt, but honestly, how often are the bugs found in this project related to passing the wrong type of value?

jessebeach:
@evcohen good to merge?

src/util/isInteractiveRole.js
evcohen: @ljharb i don't think types are *that* important for avoiding bugs. tests are better suited for making sure your code is bug-free. I think the real benefit is self-documenting code with type signatures and improved DX with editor plugins. of course this comes at the cost of verbosity, but working with AST is already inherently verbose. i think the pros here outweigh the cons, but i can see why you feel the way you do.

PR #155, opened 2017-01-11
Add the Flow project

jessebeach:
Adds the [Flow](https://flowtype.org/) project that will allow us to strongly type this codebase.

#154 is an example implementation.


ljharb:
What does `npm run flow` output right now?

jessebeach:
From this PR, it outputs:

```
No errors!
```

ljharb:
Is that because it's checking all the files and finding nothing, or is that because it's checking nothing?

jessebeach:
It's checking nothing. This is just to introduce Flow. #154 is an implementation. I'm just trying to get ESLint to parse import destructuring :/

.eslintrc
```diff
@@ -1,4 +1,5 @@
 {
+  parser: "babel-eslint",
```
ljharb: this might need to be defined after "extends", to ensure it takes precedence?

jessebeach: I'll verify.

jessebeach: It doesn't seem to make a difference. It's definitely using the babel-eslint parser in both positions in the object. But I'll move it all the same. It can't hurt.

jessebeach:
@evcohen, what do you think about Flow in the project? The typing should help us catch inconsistencies across the rules and the utils. It's a great layer of consistency as the project grows and attracts more devs.

evcohen:
@jessebeach in favor! never worked with it, but i can totally see the value.

PR #157, opened 2017-01-12
Move parser options into a common module that maps them into test cases

jessebeach:
I moved the `parserOptions` to a common module and map them into test cases so we don't need to repeat them in each case.

The default options can be overridden (shallowly for now) on a case by cases basis by passing in a `parserOptions` property with the case.


jessebeach:
> Hm, wondering if parserOptionsMapper should just be in the tests directory though, since it's really only related to the tests.

I thought so as well and it was there originally, but then jest tried to run it as a test and it filed without any asserts. I could update the test config to only grab files with `-test.js` suffixes if you like and then move this file into `__tests__`.

evcohen:
@jessebeach i like that better! thanks

__tests__/src/util/parserOptionsMapper-test.js
```diff
@@ -0,0 +1,45 @@
+/* eslint-env mocha */
```
evcohen: where are you supposed to write a test for a test util lol 😛 this is fine with me

jessebeach:
Rebased on master.

Issue #156, opened 2017-01-12
Proposal: ignore unneeded files from npm package


zpao:
I filed https://github.com/evcohen/jsx-ast-utils/issues/18 so figured I'd take quick look here too. It looks like that'll shave off ~60kb for a ~44gb saving over the last month.

I know many people like to include other things in their final packages but figured it's worth trying :)

Cheers, and thanks for the great work!

ljharb:
The only truly unneeded files are CI config, git config, and yarn/npm config - I don't think that would be very large. Tests, raw source, and babel config are definitely needed.

evcohen:
for devDependencies only, I presume @ljharb - if the only cost is network bandwidth, then i'm in favor of keeping.

thanks for suggesting @zpao 👍

ljharb:
@evcohen `npm explore foo && npm install && npm test` should always work.

zpao:
@ljharb I strongly disagree but I'm not going to die on that hill (and this isn't the place to have that metadiscussion anyway)

evcohen:
haha @zpao even if it has no bearing on an application/library's bundle size? not sure i agree with 

> should always work

but, for devDependencies it seems reasonable to provide users with as much information as possible with what code they are *actually* working with, not just what is posted on github. there's almost always a disparity between what you've downloaded locally and what's on version control.

jessebeach:
> Tests, raw source, and babel config are definitely needed.

I hope I'm interpreting this comment correctly. Why is source needed for a module package published to NPM? NPM should contain the built module files that have passed lint and tests and that have already passed through transforms. There's no need to burden someone who wants to use the module with the src files which may require who knows what transformation to actually run.

ljharb:
Because I want to be able to work with the raw source for debugging, documentation, and testing even when offline.

jessebeach:
> Because I want to be able to work with the raw source for debugging, documentation, and testing even when offline.

You can just clone the git repo for that, no?

ljharb:
Only if i knew to do that in advance;  and as @evcohen pointed out, the important code is not what is in git, it's what is installed by npm.

jessebeach:
Well then I stand with @zpao in that I disagree, but I'm not going to die on that hill either =D

ljharb:
[...]

PR #158, opened 2017-01-12
Only package lib and docs in the npm published module.

jessebeach:
Responding to #156 

Sets files in the package.json to: lib, docs

I tested the package with `npm pack`. The resulting package contained the following:

```
CHANGELOG.md
LICENSE.md
README.md
docs
lib
package.json
```

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,6 +1,4 @@
 node_modules
 reports
-npm-debug.log
 coverage
 .gitignore
-.DS_Store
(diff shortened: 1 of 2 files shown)


.npmignore
```diff
@@ -1,6 +1,4 @@
 node_modules
 reports
-npm-debug.log
```
jessebeach: This and `.DS_Store` are always ignored: https://docs.npmjs.com/files/package.json#files

PR #159, opened 2017-01-12
Add childrenPresentational property to the role json

jessebeach:
There are 14 roles that require an AT to set the role of all child elements in the AX tree to "presentation", which means they're reduced to text.

http://w3c.github.io/aria-practices/#children_presentational

```
button
checkbox
img
math
menuitemcheckbox
menuitemradio
option
progressbar
radio
scrollbar
separator
slider
switch
tab
```

I've indicated these roles in the roles.json

### DIFF
--- a/src/util/attributes/role.json
+++ b/src/util/attributes/role.json
@@ -21,7 +21,8 @@
       "ARIA-RELEVANT"
     ],
     "abstract": false,
-    "interactive": false
+    "interactive": false,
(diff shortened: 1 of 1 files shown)


evcohen:
i sense a nice rule coming out of this 😃

jessebeach:
You betcha :) This is one of those really nuanced aspects of ARIA.

PR #160, opened 2017-01-12
Add eslint-flowtypes plugin. Add flow tpe definitions for ESLint

jessebeach:
Adds eslinting for Flow syntax.

Further flowification of the rules template.


flow/eslint.js
```diff
@@ -0,0 +1,8 @@
+type ESLintReport = {
+  node: any,
+  message: string,
+};
+
+type ESLintContext = {
```
evcohen: not sure how flow works, but do you need to specify the whole object properties types or just the parts that you care about. the context object has a bunch of other properties that we aren't really making use of (but could in the future): http://eslint.org/docs/developer-guide/working-with-rules#the-context-object

jessebeach: As long as the types represented here cover what we're using, then Flow is happy. I just chatted with @zpao and ideally, we'd have these definitions in a central place for ESLint:

https://github.com/flowtype/flow-typed/tree/master/definitions/npm

@zpao, what do you think? Scaffold what we need in-project for now and port them over later? Or just go with Object in this project now?

zpao: `Object` is never super great (doesn't give you much of anything really) so I'd probably go with getting the minimum you need in your own def like you have here.

jessebeach:
@evcohen good to merge?

PR #161, opened 2017-01-12
Add definitions for React Synthetic Events

jessebeach:
React defines a set of Synthetic Events that proxy DOM events. They are handled through `on*` props on components. The added JSON file lists them.

https://facebook.github.io/react/docs/events.html


evcohen:
is this consistent across all jsx-using libraries? i hope inferno and preact don't diverge here from react :/

jessebeach:
> is this consistent across all jsx-using libraries? i hope inferno and preact don't diverge here from react :/

Oh, good point. I'll check. Sometimes I forget that JSX != React :)

jessebeach:
PReact events: https://github.com/developit/preact/blob/f0625e578c5e9a3f845a8e9461c3425647d3e4fe/src/preact.d.ts#L165

Inferno events:
https://github.com/infernojs/inferno#event-system

Inferno uses `onDblClick`, whereas React and PReact use `onDoubleClick`. I've updated the file to reflect this.

PReact seems to be missing `onLoad` and `onError`. Bizarre.

jessebeach:
hmm, I appear to be out of sync with the master timeline. 

Fixing...

evcohen:
@jessebeach weird... @developit @trueadm any reason why these are different?

jessebeach:
@evcohen @developit @trueadm, `onDoubleClick` is considered easier to read: https://github.com/facebook/react/issues/6809

trueadm:
@jessebeach I get that reasoning, but the aim for Preact and Inferno was to use the native naming schema that already existed. This actually has nothing to do with synthetic events either.

jessebeach:
> This actually has nothing to do with synthetic events either

@trueadm might a better name be `eventHandlers` or `jsxEventHandlers`.

trueadm:
@jessebeach that completely makes sense, or rather `reactEventHandlers` as JSX doesn't really care for naming schemas.

ljharb:
@jessebeach so is there no way that the React team could export their list from a package so we don't have to copy paste it?

developit:
Just seconding what @trueadm noted, by default I don't think either library invent new names for existing browser events. `onLoad` and `onError` are just missing from the TypeScript definition. `onDblClick` should be preferred over `onDoubleClick`, though the latter is supported via compat.

jessebeach:
> so is there no way that the React team could export their list from a package so we don't have to copy paste it?

Looking through the React source code, I can't find a file or set of files where these DOM event handlers are defined. My best guess is they are registered through the event registry system in a way that suits their type. @trueadm is my quick-and-dirty assessment close to the truth?

I totally support you in wanting to pull this info from React rather than hard-coding it in this project. If we can get React to declare that info, I'll swap in the reference and pull out this file. But for now, I'd like to have it to build some rules off of.

ljharb:
Definitely, not blocking whatsoever if this is the best approach for now (we've had to copy-paste a number of things in Airbnb's code because React doesn't yet export them), just poking at it :-)

evcohen:
@ljharb @jessebeach i think that even if react exported this, i'd prefer to maintain our own list for 2 reasons. 1) react is not only lib that uses jsx and 2) it would make react a dependency of this project. seems odd for someone using inferno/preact/etc to have React pulled into their project by a lint plugin devDependency.

ljharb:
My ideal scenario would be that react could extract it to a separate project that both us and react (and enzyme) could import.

lencioni:
This JSON file seems cool, but how is/will it be used in this project? Maybe there is some magic somewhere slurping it up that we don't see in this PR?

Does this type of thing make more sense for [jsx-ast-utils](https://github.com/evcohen/jsx-ast-utils)?

jessebeach:
> Does this type of thing make more sense for jsx-ast-utils?

Yes, in fact it does. I'll propose it there instead.

jessebeach:
@lencioni @evcohen Moved to https://github.com/evcohen/jsx-ast-utils/pull/20

PR #169, opened 2017-01-29
Update img-has-alt docs for latest recommendation

### LINKED ISSUE #162: Update img-has-alt to reflect new suggestion about role='presentation'
Looks like the docs don't reflect the new thinking about role='presentation' vs alt=''.
I'd change them myself but not 100% clear on what the new thinking is yet.

lencioni:
Generally speaking, developers should choose to avoid ARIA attributes if
the same thing can be achieved without them. In the case of images and
alt text, we should prefer alt="" over role="presentation". This was
discussed in #100 and fixed in #102, but the documentation wasn't fully
updated to reflect this behavior.

Fixes #162.


evcohen:
@lencioni thanks for updating!

PR #171, opened 2017-02-01
[new] - Add option to ignore non-DOM elements for rule aria-role.

### LINKED ISSUE #166: aria-role should only be applied to DOM elements
I have some data that is fetched from the server, and it comes as an object with these properties: `{ author, role, photo }`. `role` is the person's role inside the company (like `"Chief Something Officer"`). I can't create a component which takes exactly these properties, because of the `jsx-a11y/aria-role` rule. The rule should be changed to only apply to DOM elements and not React components.

evcohen:
Fixes #166

### DIFF
--- a/src/rules/aria-role.js
+++ b/src/rules/aria-role.js
@@ -7,13 +7,19 @@
 // Rule Definition
 // ----------------------------------------------------------------------------
 
-import { getLiteralPropValue, propName } from 'jsx-ast-utils';
+import { getLiteralPropValue, propName, elementType } from 'jsx-ast-utils';
 import { generateObjSchema } from '../util/schemas';
 import roles from '../util/attributes/role.json';
+import DOMElements from '../util/attributes/DOM.json';
 
 const errorMessage = 'Elements with ARIA roles must use a valid, non-abstract ARIA role.';
 
-const schema = generateObjSchema();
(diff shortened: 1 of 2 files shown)


src/rules/aria-role.js
```diff
@@ -23,29 +29,34 @@ module.exports = {
 
   create: context => ({
     JSXAttribute: (attribute) => {
+      // Determine if ignoreNonDOM is set to true
+      // If true, then do not run rule.
+      const options = context.options[0] || {};
+      const ignoreNonDOM = options.ignoreNonDOM || false;
+
+      if (ignoreNonDOM === true) {
```
ljharb: i think we can just do `if (ignoreNonDOM)` here - the schema ensures it's a boolean.

src/rules/aria-role.js
```diff
@@ -23,29 +29,34 @@ module.exports = {
 
   create: context => ({
     JSXAttribute: (attribute) => {
+      // Determine if ignoreNonDOM is set to true
+      // If true, then do not run rule.
+      const options = context.options[0] || {};
+      const ignoreNonDOM = options.ignoreNonDOM || false;
```
ljharb: `‼options.ignoreNonDOM` should be sufficient here?

PR #176, opened 2017-02-04
Fix syntax error in README

GeorgeTaveras1231:


evcohen:
ah good catch, thanks!

PR #194, opened 2017-03-21
Introduce static/noninteractive/interactive role concepts to the plugin

jessebeach:
# Summary

This PR takes advantage of improvements to the [aria-query](https://www.npmjs.com/package/aria-query) module that expose the inheritance model of ARIA. We can know determine which roles are "interactive" based on their inheritance from the abstract Widget ARIA Role. The set of roles that does not inherit from Widget are considered "noninteractive". aria-query also maps HTML elements to their inherent roles, allowing us to distinguish interactive & noninteractive elements as well as the set of elements that are neither interactive or noninteractive -- the "static" elements that have no semantic mapping.

# Awww crud, how do I review this monster diff?!?

Ya, sorry about the size :/ But I swear, a lot of the size is tests :)

1. Check out the test cases for the rules that pass and fail. See if you agree with the HTML in those buckets. **Call out cases that you disagree with**
1. Read the updated documentation for the rules. **Call out cases and remediations that you disagree with or think could be improved**.
1. Ask me questions in comments like *Will my favorite component pattern XXXX raise a violation under these rules?* and I will answer ;)

Keep in mind that I'm piloting these changes on the Facebook codebase. The goal is to prevent egregious misuse of ARIA while allowing developers to use common patterns that might not hew a strict semantic pedantry. 

# New or altered rules
## no-static-element-interactions

Disallows mouse and keyboard event handlers on static elements.

```
<div onClick={() => {}} />
<footer onClick={() => {}} />
```

Basically, this rule enforces the idea that, if you're going to give an element an interaction handler, it must have an interactive role. 

## no-noninteractive-element-interactions

Similar to the `no-static-element-interactions` rule, this disallows handlers on elements with noninteractive roles or elements that have an inherently noninteractive role.

```
<div onClick={() => {}} role="main" />
<article onClick={() => {}} />
```

This rule enforces the idea that interaction should be location on interactive elements.

## {onclick-has-focus => interactive-supports-focus}

The `onclick-has-focus` rule is renamed to `interactive-supports-focus`. Any element with a handler and an interactive role should also have a tabindex.

```
<div onClick={() => {}} role="button" />
<span onClick={() => {}} role="link" />
```

This rule only throws a violation in the case of interactive elements. Within the FB codebase I was able to reduce the count of about ~650 violations of `onclick-has-focus` to ~50 violations of `interactive-supports-focus` by limiting the criteria to interactive elements. These are 50 clear-cut cases that a developer or a codemod  (or a `--fix` flag?) can fix without ambiguity.

## no-noninteractive-element-to-interactive-role & no-interactive-element-to-noninteractive-role

These two rules work in tandem. They essentially prevent recasting of elements with inherent role activity valences to an opposite valence. For instance:

```
<article role="button" />
<li role="button" />
<button role="section" />
```

When I ran these rules across the FB codebase, I found very few (less than 5) violations of `no-interactive-element-to-noninteractive-role`. Mostly old-old-old positioning tables augmented with roles like "navigation". This rule is one that most folks will not run into; at the same time, it can help prevent really bad use of ARIA.

I found ~50 violations of `no-noninteractive-element-to-interactive-role`. Most were `<li role="button" />`, a pattern that I want to discourage. BUT, I also found patterns like this:

```
<ul role="tablist" />
<ul role="tree" />
<ul role="listbox" />
<li role="row" />
<li role="tab" />
```

The `<ul>` and `<li>` elements are often used to create "semantic" components, even though their base semantics are recast with the application of roles. It's a practice that happens and I don't want to introduce friction into the lives of developers.

So I introduced options for these two rules that allow some roles on the `<ul>` and `<li>` elements. With the options, I reduced violations down to ~25, most of which of `<li role="button" />`. 

## Conclusion

I feel confident that over a codebase of ~35K components, that these 5 rules have surfaced real issues in practice that developers can fix easily, without ambiguity and without risk of introducing bad practice. These rules are mutually exclusive, so one isn't asked to add a role and a tabindex at the same time, risking patterns like `<div role="article" onClick={() => {}} tabIndex="0" />`.

# Documentation

The documentation for these rules has been updated to include a section "How do I resolve this error?" followed by scenarios that the developer might encounter and suggestions to proceed in that case.


src/rules/no-static-element-interactions.js
```diff
@@ -1,46 +1,78 @@
...
-      if (isHiddenFromScreenReader(type, props)) {
+      if (!domElements.includes(type)) {
+        // Do not test higher level JSX components, as we do not know what
+        // low-level DOM element this maps to.
         return;
-      } else if (isInteractiveElement(type, props)) {
+      } else if (
+        !hasInteractiveProps
+        || isHiddenFromScreenReader(type, attributes)
+        || isPresentationRole(type, attributes)
```
jessebeach: @vdh , this addresses your comment: https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/91#issuecomment-252801689

__tests__/src/rules/click-events-have-key-events-test.js
```diff
@@ -26,7 +26,7 @@ const expectedError = {
   type: 'JSXOpeningElement',
 };
 
-ruleTester.run('onclick-has-role', rule, {
+ruleTester.run('click-events-have-key-events', rule, {
```
evcohen: thanks 😄

src/util/isInteractiveRole.js
```diff
@@ -33,10 +37,13 @@ const isInteractiveRole = (tagName, attributes) => {
     return false;
   }
 
+  let isInteractive = false;
   const normalizedValues = String(value).toLowerCase().split(' ');
-  const isInteractive = normalizedValues.every(
-    val => VALID_ROLES.indexOf(val) > -1,
-  );
+  if (normalizedValues.length > 0) {
```
evcohen: can we just reverse then every? i think that makes this clearer:

```js
// The last role value in a series takes precedence
const isInteractive = normalizedValues.reverse().every(
  val => interactiveRoles.indexOf(val) > -1
);
```

src/util/isNonInteractiveElement.js
```diff
@@ -0,0 +1,113 @@
...
+const isNotLink = function isNotLink(attributes) {
+  const href = getPropValue(getProp(attributes, 'href'));
+  const tabIndex = getTabIndex(getProp(attributes, 'tabIndex'));
+  return href === undefined && tabIndex === undefined;
+};
+
+export const nonInteractiveElementsMap = {
+  ...pureNonInteractiveElements,
+  area: isNotLink,
+  input: (attributes) => {
```
evcohen: `a: isNotLink,` as well?

evcohen:
@necolas - ran this against the codebase and there is only one error. great, great work

necolas:
That's good to hear, thanks :)

jessebeach:
> @necolas - ran this against the codebase and there is only one error. great, great work

What was the error? Just curious :)

jessebeach:
> This will be another major bump, so let's try and get all of the outstanding breaking changes in master before releasing (shared settings, for instance - i'll work on that).

Agreed. I've got a few things that I'd like to get in that will be major version. Maybe we should tag the issues in the queue so we know what the burn-down will be.

signalwerk:
wow! thx for your work!

evcohen:
> What was the error? Just curious :)

```js
<div className={className}
    onClick={this.props.onClick}
    role='article'
    tabIndex='0'
 >
```

which is the Tweet container 😛

docs/rules/interactive-supports-focus.md
```diff
@@ -0,0 +1,82 @@
...
+
+Generally buttons, links and form elements should be reachable via tab key presses.
+An element that can be tabbed to is said to be in the _tab ring_.
+
+### Case: This element is part of a group of buttons, links, menu items, etc
+
+One item in a group should have a tabindex of zero, the rest should have a tabindex of -1. A value of zero makes the element _tabbable_. A value of -1 makes the element _focusable_.
+
+```
+<div role="men">
```
jessebeach: lolz

docs/rules/no-noninteractive-element-to-interactive-role.md
```diff
@@ -0,0 +1,48 @@
...
+### References
+
+1. [WAI-ARIA roles](https://www.w3.org/TR/wai-aria-1.1/#usage_intro)
+1. [WAI-ARIA Authoring Practices Guide - Design Patterns and Widgets](https://www.w3.org/TR/wai-aria-practices-1.1/#aria_ex)
+1. [Fundamental Keyboard Navigation Conventions](https://www.w3.org/TR/wai-aria-practices-1.1/#kbd_generalnav)
+1. [Mozilla Developer Network - ARIA Techniques](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_button_role#Keyboard_and_focus)
+
+## Rule details
+
+This rule takes no arguments.
```
jessebeach: yes. Indeed. Updating.

src/util/isInteractiveRole.js
jessebeach: Funny enough, I had a conversation with @mcking65 about this the other day. Turns out it's the **first** valid role in the list that gets used.

jessebeach: I updated and added test coverage for this scenario.

src/util/isNonInteractiveElement.js
jessebeach: I'm treating `<a />` as a static element, because it does not have an inherent role association. 

Currently, the following flow would obtain:

```
<a onClick={function () {}} /> !! No handlers on static elements
<a onClick={function () {}} role="button" /> !! tabIndex is required
```

If we mark it as non-interactive, the following would happen:

```
<a onClick={function () {}} /> !! No handlers on non-interactives
<a onClick={function () {}} role="button" /> !! No non-interactive to interactive casting
<span onClick={function () {}} role="button" /> !! tabIndex is required
```

Essentially, `<a />` becomes unusable and must be replaced by a `<span />` or some such. I think this might cause developer friction since in practice `<a />` and `<span />` aren't different.

.npmignore
```diff
@@ -1,7 +1,15 @@
+.babelrc
+.DS_Store
+.eslintrc
```
ljharb: `npm explore foo && npm install && npm test` should always work - why do we need to blacklist (in this PR, which is unrelated) these negligible and useful files?

jessebeach: Perhaps we should return #156.

> `npm explore foo && npm install && npm test` should always work

This is an opinion, not a law. There are [pros and cons](http://stackoverflow.com/questions/25124844/should-i-npmignore-my-tests/25125863#25125863) to this approach. Including files meant for development in a production environment has knockoff effects. Specifically, the one I ran into is `.flowconfig`. Including this in the release package causes the including project to run flow against this project with an incorrect `<project_root>`. 

Including tests and src in our package makes it larger to download and introduces the possibility for build incompatibilities in third-party inclusions. 

I'll change this back for this PR. I agree, it's not related and it doesn't belong in these changes. For that I apologize. 

But we should get our build files out of the package.

ljharb: semver is an opinion too ;-) Happy to continue this discussion on a separate thread tho :-)

jessebeach: ha! yes. A very confusing opinion in practice =D

jessebeach:
I'm codemodding the FB codebase against these new rules and I'm finding some quirks, like:

`<p onClick={() => {}} />`

This is considered a "static" element because it has no mapping in ARIA, so the rules is asking the user to add a role. Well, the `<p>` tag has a role, it just isn't supplied by ARIA. For that, we need to look to the AXObjects that get created by the browser. For instance, the `<p>` tag has a `ParagraphRole`: https://cs.chromium.org/chromium/src/third_party/WebKit/Source/modules/accessibility/AXObject.h?type=cs&l=137

So, in a follow-up to this diff, I'll be modeling the AXObjects and correlating those objects with concepts of interactivity so that our rules can be more refined. I'm already working on this.

mcking65:
@jessebeach, btw, aria 1.2 scope is to fill such role gaps in aria.

jessebeach:
@mcking65 well, maybe I can give you a list of exactly what those gaps are ;D

mcking65:
Thx @jessebeach , We have the gaps defined in the [issues for aria 1.2](https://github.com/w3c/aria/issues?q=is%3Aopen+is%3Aissue+label%3A%22ARIA+1.2%22)

jessebeach:
@evcohen Shall we get this in? I've already got a followup ready to go:

https://github.com/jessebeach/eslint-plugin-jsx-a11y/compare/master...jessebeach:axmodel-query-integration

I've run the followup against the Facebook codebase and I've eliminated the many violations of no-static-element-interactions. It's running clean and green =D

evcohen:
@jessebeach AH hi sorry, been crazy here everyone else is on PTO right now so I'm flying solo :P let me take a final look at this tonight and get it in ASAP. thanks for the reminder 😄

jessebeach:
Rebased on `evcohen/master`.

PR #205, opened 2017-04-14
Ignore unneeded dot files in npm package

jessebeach:
This one is for @ljharb =D

The .flowconfig file causes havoc when it's included in the built package.

BUT, without it, `npm test` fails.

So I updated the `flow` script to exit gracefully when `.flowconfig` isn't present. 

`.npmignore` is updated to exclude files that aren't necessary in the NPM package.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,6 +1,5 @@
 .DS_Store
 .flowconfig
-.gitignore
 .travis.yml
 coverage
 flow
--- a/.npmignore
+++ b/.npmignore
@@ -1,7 +1,10 @@
-node_modules
-reports
-npm-debug.log
-coverage
-.gitignore
 .DS_Store
+.flowconfig
+.gitignore
+.travis.yml
+coverage
+flow
+node_modules
+npm-debug.log
+reports
 yarn-error.log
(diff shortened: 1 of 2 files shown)


jessebeach:
I alphabetized `.npmignore` as well.

.npmignore
```diff
@@ -1,7 +1,10 @@
+.DS_Store
+.flowconfig
+.gitignore
```
ljharb: i'm pretty sure npm never includes this file, so including it here is unnecessary - unless `npm pack` proves me wrong.

jessebeach:
To test this, I ran `npm pack`, took the package, untarred it and ran `npm i` on it to make sure it builds and passes the test run.

jessebeach:
> Can you elaborate on what kind of havoc it causes to include the file?

When our plugin is included as a dependency in a project that also uses flow, it causes flow to check our module, but the cwd is set to the depender module, so the flow check fails with "can't find such-and-such module" errors.

> Flow should be able to gracefully handle the inclusion or exclusion of any particular kind of file in "not the Flow user's code".

Debatable. We really shouldn't be packaging our build-time resources in the run-time package. But we are and you stated before that you wanted to be able to run `npm test` from the npm package and I want to respect that desire, so I was hoping this PR could be a compromise.

.npmignore
jessebeach: `npm pack` proves you correct! I'll update and remove `.gitignore`.

jessebeach:
@ljharb, found the issue here: https://github.com/facebook/flow/issues/1548

PR #206, opened 2017-04-14
Update Fail example for anchor-has-content rule

liubko:
closing `</a>` tag is missing


liubko:
I'm glad to contribute 😃

PR #207, opened 2017-04-14
Axmodel query integration

jessebeach:
The list of 'static' elements -- elements that are not modeled or referenced in ARIA - is quite long. It includes elements that we would expect to be represented in ARIA like `<label>` and `<p>`.

Browsers *do* model these elements in their AX Tree. So by taking the AXObject models into account, we gain a more complete understanding of how HTML and DOM are represented in the AX Tree.

These changes will disallow constructs like `<label onClick />` and `<label role="button" />`.


.eslintrc
```diff
@@ -8,6 +8,12 @@
     "flowtype"
   ],
   rules: {
-    'no-template-curly-in-string': 'off'
+    'no-template-curly-in-string': 'off',
+    'no-restricted-syntax': [
```
jessebeach: Removed the disallowance of the for..of statement for iterating iterables.

__mocks__/genInteractives.js
```diff
@@ -42,20 +46,29 @@ const interactiveElementsMap = {
...
   select: [],
-  'table[role="grid"]': [{prop: 'role', value: 'grid'}],
-  'td[role="gridcell"]': [{prop: 'role', value: 'gridcell'}],
+  th: [],
   tr: [],
   textarea: [],
+  video: [],
 };
 
 const nonInteractiveElementsMap = {
```
jessebeach: New noninteractives are derived from the AXObject model.

__mocks__/genInteractives.js
```diff
@@ -42,20 +46,29 @@ const interactiveElementsMap = {
   menuitem:[],
   option: [],
   select: [],
-  'table[role="grid"]': [{prop: 'role', value: 'grid'}],
-  'td[role="gridcell"]': [{prop: 'role', value: 'gridcell'}],
+  th: [],
   tr: [],
   textarea: [],
+  video: [],
```
jessebeach: `<video>` is now recognized as an interactive element, which eliminates situations like this, where developers feel they need to add a role:

https://github.com/tootsuite/mastodon/pull/1651#discussion_r111453507

src/util/isInteractiveElement.js
```diff
@@ -31,69 +39,118 @@ const interactiveRoles = new Set(
...
+    ),
+  );
+}
+
+function checkIsInteractiveElement(tagName, attributes): boolean {
+  // Check in configuration for an override.
+
+  // Check in elementRoles for inherent interactive role associations for
+  // this element.
+  for (const [elementSchema] of interactiveElementRoles) {
```
ljharb: Why not use `.some` here and check the Boolean, rather than use an imperative loop?

src/util/isInteractiveElement.js
```diff
@@ -31,69 +39,118 @@ const interactiveRoles = new Set(
...
+      tagName === elementSchema.name
+      && attributesComparator(elementSchema.attributes, attributes)
+    ) {
+      return true;
+    }
+  }
+
+  // Check in elementRoles for inherent interactive role associations for
+  // this element.
+  for (const [elementSchema] of nonInteractiveElementRoles) {
```
ljharb: Same here re some/every

src/util/isInteractiveElement.js
```diff
@@ -105,11 +162,13 @@ const isInteractiveElement = (
   tagName: string,
   attributes: Array<Node>,
 ): boolean => {
-  if ({}.hasOwnProperty.call(interactiveElementsMap, tagName) === false) {
+  // Do not test higher level JSX components, as we do not know what
+  // low-level DOM element this maps to.
+  if ([...dom.keys()].indexOf(tagName) === -1) {
```
ljharb: What construct is `dom` that the most straightforward way of checking if a key is present is reifying all the keys into an array and checking indexOf?

If it's a map or set, we should be able to use `.has`, if an array, `Object.prototype.hasOwnProperty.call`.

src/util/isNonInteractiveElement.js
```diff
@@ -7,81 +7,145 @@ import {
...
-    return nonInteractiveElements;
-  }, {});
-
-const isNotLink = function isNotLink(attributes) {
-  const href = getPropValue(getProp(attributes, 'href'));
-  const tabIndex = getTabIndex(getProp(attributes, 'tabIndex'));
-  return href === undefined && tabIndex === undefined;
-};
+  ) => {
+    if ([...roleSet.keys()].every(
```
ljharb: Sets have the same key and value for each entry, and the default iterator is values - does this need .keys()?

src/util/isInteractiveElement.js
jessebeach: In this case `nonInteractiveElementRoles` is a Map, so I'm using `for..of` to loop through the entries iterator. It doesn't have a `some` method.

src/util/isInteractiveElement.js
jessebeach: Good call.

src/util/isNonInteractiveElement.js
jessebeach: Indeed, good catch, you're right :)

jessebeach:
> General comment about all the explicit keys/entries iterators - most of them seem unnecessary. Can they be omitted?

Which ones do you mean? I'm worried I'm making the wrong assumption about what you mean to indicate.

src/util/isInteractiveElement.js
jessebeach: > Why not use .some here and check the Boolean, rather than use an imperative loop?

Also, the imperative loop allows for a return from the outer function immediately.

ljharb:
@jessebeach np, i'll comment on each explicitly (just didn't want to overwhelm you with non-a11y-related nits)

src/util/isInteractiveElement.js
```diff
@@ -2,25 +2,33 @@
...
+import {
   getLiteralPropValue,
   propName,
 } from 'jsx-ast-utils';
-import getTabIndex from './getTabIndex';
 
-type ElementCallbackMap = {
-  [elementName: string]: (attributes: Array<Node>) => boolean,
-};
+const roleKeys = [...roles.keys()];
```
ljharb: is `roles` a Map, or a Set? If it's a Set, you can just do `[...roles]`

src/util/isInteractiveElement.js
```diff
@@ -2,25 +2,33 @@
...
+
+const nonInteractiveRoles = new Set(
+  roleKeys
+    .filter(name => !roles.get(name).abstract)
+    .filter(name => !roles.get(name).superClass.some(
+      klasses => klasses.includes('widget')),
```
ljharb: `classes` isn't a reserved word :-)

src/util/isInteractiveElement.js
```diff
@@ -31,69 +39,118 @@ const interactiveRoles = new Set(
     ),
 );
 
-// Map of tagNames to functions that return whether that element is interactive or not.
-const pureInteractiveRoleElements = [...elementRoles.entries()]
+
+const nonInteractiveElementRoles = [...elementRoles.entries()]
```
ljharb: is `elementRoles` a Map? if so, `[...elementRoles]` should work.

src/util/isInteractiveElement.js
```diff
@@ -31,69 +39,118 @@ const interactiveRoles = new Set(
...
-    return interactiveElements;
-  }, {});
-
-const isLink = function isLink(attributes) {
-  const href = getPropValue(getProp(attributes, 'href'));
-  const tabIndex = getTabIndex(getProp(attributes, 'tabIndex'));
-  return href !== undefined || tabIndex !== undefined;
-};
+  ) => {
+    if ([...roleSet.keys()].every(
```
ljharb: `[...roleSet]` should be sufficient here.

src/util/isInteractiveElement.js
```diff
@@ -31,69 +39,118 @@ const interactiveRoles = new Set(
...
+      (role): boolean => nonInteractiveRoles.has(role),
+    )) {
+      accumulator.set(elementSchema, roleSet);
+    }
+    return accumulator;
+  }, new Map([]));
```
ljharb: `new Map()` is the same here as `new Map(emptyIterable)` - we can just omit the `[]`

src/util/isInteractiveElement.js
```diff
@@ -31,69 +39,118 @@ const interactiveRoles = new Set(
...
-    getProp(attributes, 'role'),
-  ) === 'gridcell',
-  table: (attributes) => {
-    const role = getLiteralPropValue(
-      getProp(attributes, 'role'),
-    );
-    return (role === 'grid');
-  },
-};
+const interactiveElementRoles = [...elementRoles.entries()]
```
ljharb: if `elementRoles` is a Map, then `[...elementRoles]` should work just fine.

src/util/isInteractiveElement.js
```diff
@@ -31,69 +39,118 @@ const interactiveRoles = new Set(
...
+      roleSet,
+    ],
+  ) => {
+    if ([...roleSet.keys()].some(
+      (role): boolean => interactiveRoles.has(role),
+    )) {
+      accumulator.set(elementSchema, roleSet);
+    }
+    return accumulator;
+  }, new Map([]));
```
ljharb: `new Map()`

src/util/isInteractiveElement.js
```diff
@@ -31,69 +39,118 @@ const interactiveRoles = new Set(
...
+
+const interactiveAXObjects = new Set(
+  [...AXObjects.keys()]
```
ljharb: is `AXObjects` a Map, or a Set? if a Set, `[...AXObjects]` is sufficient.

src/util/isInteractiveElement.js
```diff
@@ -31,69 +39,118 @@ const interactiveRoles = new Set(
...
+    .filter(name => ['widget'].includes(AXObjects.get(name).type)),
```
ljharb: `.filter(name => AXObjects.get(name).type === 'widget')`? unless this array is meant to be expanded later, in which case, perhaps it should be stored in a named variable instead of being created inline?

src/util/isInteractiveElement.js
```diff
@@ -31,69 +39,118 @@ const interactiveRoles = new Set(
...
+);
+
+const interactiveElementAXObjects = [...elementAXObjects.entries()]
```
ljharb: If this is a Map, `[...elementAXObjects]` is sufficient here

src/util/isInteractiveElement.js
```diff
@@ -31,69 +39,118 @@ const interactiveRoles = new Set(
...
+      AXObjectSet,
+    ],
+  ) => {
+    if ([...AXObjectSet.keys()].every(
+      (role): boolean => interactiveAXObjects.has(role),
+    )) {
+      accumulator.set(elementSchema, AXObjectSet);
+    }
+    return accumulator;
+  }, new Map([]));
```
ljharb: `new Map()`

src/util/isInteractiveElement.js
```diff
@@ -31,69 +39,118 @@ const interactiveRoles = new Set(
...
+
+const interactiveElementAXObjects = [...elementAXObjects.entries()]
+  .reduce((
+    accumulator,
+    [
+      elementSchema,
+      AXObjectSet,
+    ],
+  ) => {
+    if ([...AXObjectSet.keys()].every(
```
ljharb: `[...AXObjectSet]` is sufficient here.

src/util/isInteractiveElement.js
```diff
@@ -31,69 +39,118 @@ const interactiveRoles = new Set(
...
-};
+const interactiveElementRoles = [...elementRoles.entries()]
+  .reduce((
+    accumulator,
+    [
+      elementSchema,
+      roleSet,
+    ],
+  ) => {
+    if ([...roleSet.keys()].some(
```
ljharb: `[...roleSet]` is sufficient here

src/util/isInteractiveElement.js
```diff
@@ -31,69 +39,118 @@ const interactiveRoles = new Set(
...
+        let valueMatches = true;
+        // Attribute matches.
+        if (baseAttr.name === propName(attribute).toLowerCase()) {
+          attrMatches = true;
+        }
+        // Value exists and matches.
+        if (baseAttr.value) {
+          valueMatches = baseAttr.value === getLiteralPropValue(attribute);
+        }
+        return attrMatches && valueMatches;
```
ljharb: nbd, but this might be clearer if you do this:
```js
// attribute does not match
if (baseAttr.name !== baseAttr.valuepropName(attribute).toLowerCase()) {
  return false;
}

// value exists and does not match
if (baseAttr.value && baseAttr.value !== getLiteralPropValue(attribute)) {
  return false;
}

return true;
```

if you disagree, not a problem :-)

src/util/isInteractiveElement.js
ljharb: That's why i prefer the iterative approach :-) I find it clearer to do:
```js
const someMeaningfulVarName =  [...interactiveElementRoles].some(([elementSchema]) => (
  tagName === elementSchema.name
  && attributesComparator(elementSchema.attributes, attributes)
));
if (someMeaningfulVarName) {
  return true;
}
```

but this might be personal preference.

If we end up with for loops here, could we use inline override comments instead of disabling the rule for the whole repo?

src/util/isNonInteractiveElement.js
```diff
@@ -7,81 +7,145 @@ import {
...
       klasses => klasses.includes('widget')),
     ),
 );
 
-const pureNonInteractiveElements = [...elementRoles.entries()]
+const interactiveRoles = new Set(
+  roleKeys
+    .filter(name => !roles.get(name).abstract)
+    .filter(name => roles.get(name).superClass.some(
+      klasses => klasses.includes('widget')),
```
ljharb: `classes` is not a reserved word

src/util/isNonInteractiveElement.js
```diff
@@ -7,81 +7,145 @@ import {
...
     .filter(name => !roles.get(name).abstract)
     .filter(name => !roles.get(name).superClass.some(
       klasses => klasses.includes('widget')),
     ),
 );
 
-const pureNonInteractiveElements = [...elementRoles.entries()]
+const interactiveRoles = new Set(
+  roleKeys
+    .filter(name => !roles.get(name).abstract)
```
ljharb: what about:
```js
roleKeys
  .map(name => roles.get(name))
  .filter(role => (
    !role.abstract
    && role.superClass.some(classes => classes.includes('widget'))
  )
```
? nbd, just a thought

src/util/isNonInteractiveElement.js
```diff
@@ -7,81 +7,145 @@ import {
...
-  return href === undefined && tabIndex === undefined;
-};
+  ) => {
+    if ([...roleSet].every(
+      (role): boolean => nonInteractiveRoles.has(role),
+    )) {
+      accumulator.set(elementSchema, roleSet);
+    }
+    return accumulator;
+  }, new Map([]));
```
ljharb: `new Map()`

src/util/isNonInteractiveElement.js
```diff
@@ -7,81 +7,145 @@ import {
...
-};
+const interactiveElementRoles = [...elementRoles.entries()]
+  .reduce((
+    accumulator,
+    [
+      elementSchema,
+      roleSet,
+    ],
+  ) => {
+    if ([...roleSet.keys()].some(
```
ljharb: `[...roleSet]` is sufficient here

src/util/isNonInteractiveElement.js
```diff
@@ -7,81 +7,145 @@ import {
...
+      (role): boolean => interactiveRoles.has(role),
+    )) {
+      accumulator.set(elementSchema, roleSet);
+    }
+    return accumulator;
+  }, new Map([]));
```
ljharb: `new Map()`

src/util/isNonInteractiveElement.js
```diff
@@ -7,81 +7,145 @@ import {
...
+
+const nonInteractiveAXObjects = new Set(
+  [...AXObjects.keys()]
```
ljharb: `[...AXObjects]` is sufficient here if this is a `Set`

src/util/isNonInteractiveElement.js
```diff
@@ -7,81 +7,145 @@ import {
...
+    .filter(name => ['window', 'structure'].includes(AXObjects.get(name).type)),
```
ljharb: where do "window" and "structure" come from? should part of this filterer go in a named helper method?

src/util/isNonInteractiveElement.js
```diff
@@ -7,81 +7,145 @@ import {
...
+
+const nonInteractiveElementAXObjects = [...elementAXObjects.entries()]
+  .reduce((
+    accumulator,
+    [
+      elementSchema,
+      AXObjectSet,
+    ],
+  ) => {
+    if ([...AXObjectSet.keys()].every(
```
ljharb: `[...AXObjectSet]`

src/util/isNonInteractiveElement.js
```diff
@@ -7,81 +7,145 @@ import {
...
+      (role): boolean => nonInteractiveAXObjects.has(role),
+    )) {
+      accumulator.set(elementSchema, AXObjectSet);
+    }
+    return accumulator;
+  }, new Map([]));
```
ljharb: `new Map()`

src/util/isNonInteractiveElement.js
```diff
@@ -7,81 +7,145 @@ import {
...
+  return baseAttributes.every(
+    (baseAttr): boolean => attributes.some(
+      (attribute): boolean => {
+        // Guard against non-JSXAttribute nodes like JSXSpreadAttribute
+        if (attribute.type !== 'JSXAttribute') {
+          return false;
+        }
+        let attrMatches = false;
+        let valueMatches = true;
+        // Attribute matches.
```
ljharb: this logic seems somewhat reused from earlier in the file; could it be shared somehow?

src/util/isNonInteractiveElement.js
```diff
@@ -7,81 +7,145 @@ import {
...
+    ),
+  );
+}
+
+function checkIsNonInteractiveElement(tagName, attributes): boolean {
+  // Check in configuration for an override.
+
+  // Check in elementRoles for inherent non-interactive role associations for
+  // this element.
+  for (const [elementSchema] of nonInteractiveElementRoles) {
```
ljharb: all the same comments about for loops apply here :-)

src/util/isInteractiveElement.js
jessebeach: > True - you could spread to an array and then use some, as you've done elsewhere?

It just gets messy with `.some` in this case. I need to save the bool to a variable, then in a subsequent if check the bool and return if it's true. The `for..of` loop reads much easier (personal opinion, admittedly) in this case. I'm trying to not dig in my heels here :) It just feels more fluent with the loop.

ljharb: No worries then; it's not a hill worth dying on for `for..of` in particular - but it'd still be nice to have inline overrides rather than a repo-wide one :-)

jessebeach: That's what I ended up with as well, your code snippet. I've got a very lightly held opinion on this. Happy to alter and restore the eslint rule for `for..of` loops. Doing that now...

src/util/isInteractiveElement.js
jessebeach: Going with an array literal here for the accumulator. Turns out I'm not using the roleSet later on.

src/util/isInteractiveElement.js
jessebeach: hmm, that's much clearer.

src/util/isNonInteractiveElement.js
jessebeach: Almost. This rill create an array of roles, not role names. But I can tighten this up a bit for sure.

ljharb: Ah, gotcha :-) combining the filters might be nice then, or something.

src/util/isNonInteractiveElement.js
jessebeach: Moved into its own util; tests added as well.

jessebeach:
> Sorry for the deluge; happy to elaborate if any of my comments aren't clear.

No worries. I appreciate the input. I learned a lot about iterators tonight!

src/util/isInteractiveElement.js
```diff
@@ -2,98 +2,135 @@
...
+      roleKeys,
       // 'toolbar' does not descend from widget, but it does support
       // aria-activedescendant, thus in practice we treat it as a widget.
       'toolbar',
     )
-    .filter(name => !roles.get(name).abstract)
-    .filter(name => roles.get(name).superClass.some(
-      klasses => klasses.includes('widget')),
-    ),
+    .filter((name) => {
```
ljharb: this filter function is the same as the one on line 21 - could it be shared?

src/util/isNonInteractiveElement.js
```diff
@@ -7,81 +7,130 @@ import {
...
+          classes => classes.includes('widget'),
+        )
+      );
+    }),
 );
 
-const pureNonInteractiveElements = [...elementRoles.entries()]
+const interactiveRoles = new Set(
+    [].concat(
+      roleKeys,
```
ljharb: `roleKeys.concat('toolbar')` should work here too

src/util/attributesComparator.js
```diff
@@ -0,0 +1,43 @@
...
+  getLiteralPropValue,
+  propName,
+} from 'jsx-ast-utils';
+import type { Node } from 'ast-types-flow';
+
+/**
+ * Returns true if all items in baseAttributes are found in attributes. Always
+ * returns true if baseAttributes is empty.
+ */
+function attributesComparator(
```
evcohen: I think it would be helpful if this was fn was renamed - took me a while to understand what was going on here. It's not really a commutative comparison. We're basically just checking if all of the baseAttributes are in the provided attributes and equivalent. Maybe something like `hasEveryBaseAttribute` because this implies that empty baseAttributes should return true. the tradeoff being that it doesn't communicate that we're checking value equivalence as well to determine their equality. naming is hard 😞 happy to discuss

package.json
```diff
@@ -56,6 +56,7 @@
   "dependencies": {
     "aria-query": "^0.5.0",
     "ast-types-flow": "0.0.7",
+    "axobject-query": "^0.1.0",
```
evcohen: Nice!

src/util/isInteractiveElement.js
evcohen: It could, but I think its fine as is. It makes this easier to follow reading sequentially, as opposed to jumping around to shared functions. sometimes the cost of DRY is higher than duplicating logic

src/util/isInteractiveElement.js
```diff
@@ -2,98 +2,135 @@
...
+  ) => {
+    if ([...AXObjectSet].every(
+      (role): boolean => interactiveAXObjects.has(role),
+    )) {
+      accumulator.push(elementSchema);
+    }
+    return accumulator;
+  }, []);
+
+function checkIsInteractiveElement(tagName, attributes): boolean {
```
evcohen: This is a really great job- easy to follow and understand, and simplifies a ton of the complexities here.

src/util/attributesComparator.js
jessebeach: > `hasEveryBaseAttribute`

Bingo, that's much better. I had a difficult time naming this one ;)

src/util/isInteractiveElement.js
jessebeach: > sometimes the cost of DRY is higher than duplicating logic

So true. A lot of the array manipulations here are just rote transformations and its the transform itself produces the utility of the rule.

jessebeach:
> can you explain the difference between the AXObject and ARIA? Is ARIA the spec and AXObject the implementation?

Sure, I do this in the README

jessebeach:
Will continue to clean in followups that are smaller and targeted. This should be the last of the **big** commits. I've got a couple new rules and rule improvements on the way.

PR #208, opened 2017-04-14
Disregard interactive props with null values

jessebeach:
I found that elements that have nulled out handler values are getting flagged by `no-static-element-interactions` and `no-noninteractive-element-interactions`. 

These rules shouldn't trigger on elements that do not have interaction handling.


jessebeach:
> This seems like it'd be a useful helper in jsx-ast-utils

I had this thought as well. But then I couldn't think of where else it'd be used in this plugin. If I find a third use for the pattern, I'll move it to jsx-ast-utils.

PR #209, opened 2017-04-15
Integrate prettier into toolchain to avoid inconsistent formatting.

evcohen:
👨🏼‍🍳


evcohen:
Woah there- tests are failing!

ljharb:
What does prettier do that `eslint --fix` can't do, except be less configurable?

evcohen:
[This](http://jlongster.com/A-Prettier-Formatter) convinced me of its value. This PR will run prettier over the changed files and then run `eslint --fix` so that we get the benefit of a consistent style format and linting for logic-based rules (like this plugin).

We can update eslint to have fix options for some of these rules but it feels like a good separation of concerns to have a style formatter distinct from a rule-based linter.

ljharb:
1. Since you're running both, I don't think there's a *danger* with using prettier here first.
1. I have a huge problem with using husky; since it installs hooks into .git on *`npm install`*, it has caused unaudited hooks to get installed in the wrong project before (specifically, unaudited hooks in one of my company's private corporate repos, because i ran `npm install $projectContainingHusky` inside `node_modules`. There's an open issue for it). Commit hooks should only *ever* be explicitly, manually installed by a user. Separately, anything not enforced by `npm test` doesn't count, since I can always delete my local git hook after installing. I think this should be opt-in for a precommit hook *only*, and ran as part of `npm test` otherwise.
1. "style" and "linting" aren't separate concerns. Style is just as important as parsing errors and potential bugs, because all of them are related to clarity, robustness, consistency, and avoiding debates on PRs.

The article you linked me to merely convinced me that a developer chose to reinvent a wheel instead of making the defacto standard of code style and linting better for everyone. I would say that anything prettier can do that eslint can't be made to do is a flaw in eslint, and adding build tool complexity isn't a good workaround for an eslint problem.

package.json
```diff
@@ -21,9 +21,9 @@
     "prepublish": "npm run lint && npm run flow && npm run test && npm run build",
     "coveralls": "cat ./reports/lcov.info | coveralls",
     "flow": "if [ ! -e ./.flowconfig ]; then echo \"Could not find .flowconfig\"; else flow; test $? -eq 0 -o $? -eq 2; fi",
-    "lint": "eslint  --config .eslintrc src __tests__",
+    "lint": "eslint  --config .eslintrc src __tests__ __mocks__",
     "lint:fix": "npm run lint -- --fix",
-    "pretest": "npm run lint:fix && npm run flow",
```
jessebeach: Does the `flow` check still get run?

jessebeach:
It solves a consistency problem today. I approve.

jessebeach:
Does this change the build or release process? Just curious.

jessebeach:
Let's get this in!

src/util/isPresentationRole.js
```diff
@@ -1,12 +1,8 @@
...
-const presentationRoles = new Set([
-  'presentation',
-  'none',
-]);
+const presentationRoles = new Set(['presentation', 'none']);
 
-const isPresentationRole = (tagName, attributes) => presentationRoles.has(
-    getLiteralPropValue(getProp(attributes, 'role')),
-  );
+const isPresentationRole = (tagName, attributes) =>
```
ljharb: i'm not a fan of this indentation.

This should be:
```js
const isPresentationRole = (tagName, attributes) => (
  presentationRoles.has(getLiteralPropValue(getProp(attributes, 'role'))
);
```

src/util/schemas.js
```diff
@@ -14,12 +14,13 @@ export const arraySchema = {
 /**
  * JSON schema to accept an array of unique strings from an enumerated list.
  */
-export const enumArraySchema = (enumeratedList = []) => Object.assign({}, arraySchema, {
-  items: {
-    type: 'string',
-    enum: enumeratedList,
-  },
-});
+export const enumArraySchema = (enumeratedList = []) =>
```
ljharb: Similarly, the hanging nothing after the `=>` is weird, without wrapping parens.

src/util/isNonInteractiveElement.js
```diff
@@ -17,48 +13,52 @@ import {
...
-              }
-              return controlAttr.name === propName(attr).toLowerCase()
-                && controlAttr.value === getLiteralPropValue(attr);
-            },
-          ),
-        );
-      return passedAttrCheck && [...roleSet.keys()].every(
-        (roleName): boolean => nonInteractiveRoles.has(roleName),
+const pureNonInteractiveElements = [
+  ...elementRoles.entries(),
```
ljharb: I'd prefer `[...elementRoles.entries()].reduce((accumulator: ElementCallbackMap, [` here

PR #210, opened 2017-04-17
No noninteractive tabindex

jessebeach:
This rule is intended to disallow constructs like:

```jsx
<li tabIndex="0" />
<div role="presentation" tabIndex="0" />
```

I ran this over ~33K components and found ~170 violations.

### DIFF
--- a/docs/rules/no-noninteractive-tabindex.md
+++ b/docs/rules/no-noninteractive-tabindex.md
@@ -1,10 +1,54 @@
 # no-noninteractive-tabindex
 
-Write a useful explanation here!
+Tab key navigation should be limited to elements on the page that can be interacted with. Thus it is not necessary to add a tabindex to items in an unordered list, for example, to make them navigable through assistive technology. These applications already afford page traversal mechanisms based on the HTML of the page. Generally, we should try to reduce the size of the page's tab ring rather than increasing it. 
+
+## How do I resolve this error?
+
+### Case: I am using an `<a>` tag. Isn't that interactive?
+
+The `<a>` tag is tricky. Consider the following:
+
+```
+<a>Edit</a>
+<a href="#">Edit</a>
+<a role="button">Edit</a>
+```
+
+The bare `<a>` tag is an _anchor_. It has no semantic AX API mapping in either ARIA or the AXObject model. It's as meaningful as `<div>`, which is to say it has no meaning. An `<a>` tag with an `href` attribute has an inherent role of `link`. An `<a>` tag with an explicit role obtains the designated role. In the example above, this role is `button`.
+
+### Case: I am using "semantic" HTML. Isn't that interactive?
+
+If we take a step back into the field of linguistics for a moment, let's consider what it means for something to be "semantic". Nothing, in and of itself, has meaning. Meaning is constructed through dialogue. A speaker intends a meaning and a listener/observer interprets a meaning. Each participant constructs their own meaning through dialogue. There is no intrinsic or isolated meaning outside of interaction. Thus, we must ask, given that have a "speaker" who communicates via "semantic" HTML, who is listening/observing?
+
+In our case, the observer is the Accessibility (AX) API. Browsers interpret HTML (inflected at times by ARIA) to construct a meaning (AX Tree) of the page. Whatever the semantic HTML intends has only the force of suggestion to the AX API. Therefore, we have inconsistencies. For example, there is not yet an ARIA role for `text` or `label` and thus no way to change a `<label>` into plain text or a `<span>` into a label via ARIA. '<div>' has an AXObject correpondant `DivRole`, but no such object maps to `<span>`.
+
+What this lint rule endeavors to do is apply the AX API understanding of the semantics of an HTML document back onto your code. The concept of interactivity boils down to whether a user can do something with the indicated or focused component.
+
+Common interactive roles include:
+
+  1. `button`
+  1. `link`
+  1. `checkbox`
+  1. `menuitem`
+  1. `menuitemcheckbox`
+  1. `menuitemradio`
+  1. `option`
+  1. `radio`
+  1. `searchbox`
+  1. `switch`
+  1. `textbox`
+
+Endeavor to limit tabbable elements to those that a user can act upon.
+
+### Case: Shouldn't I add a tabindex so that users can navigate to this item?
+
+It is not necessary to put a tabindex or an `<article>`, for instance or on `<li>` items; assistive technologies provide affordances to users to find and traverse these containers. Most elements that require a tabindex -- `<a href>`, `<button>`, `<input>`, `<textarea>` -- have it already.
+
+Your application might require an exception to this rule in the case of an element that captures incoming tab traversal for a composite widget. In that case, turn off this rule on a per instance basis. This is an uncommon case.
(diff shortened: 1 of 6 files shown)


src/index.js
```diff
@@ -118,6 +130,7 @@ module.exports = {
         'jsx-a11y/no-interactive-element-to-noninteractive-role': 'error',
         'jsx-a11y/no-noninteractive-element-interactions': 'error',
         'jsx-a11y/no-noninteractive-element-to-interactive-role': 'error',
+        'jsx-a11y/no-noninteractive-tabindex': 'error',
```
ljharb: adding the rule to an exported config makes this semver-major - could we add the rule without enabling it in a config as semver-minor, and add it to the config later?

docs/rules/no-noninteractive-tabindex.md
```diff
@@ -0,0 +1,77 @@
...
+  1. `switch`
+  1. `textbox`
+
+Endeavor to limit tabbable elements to those that a user can act upon.
+
+### Case: Shouldn't I add a tabindex so that users can navigate to this item?
+
+It is not necessary to put a tabindex or an `<article>`, for instance or on `<li>` items; assistive technologies provide affordances to users to find and traverse these containers. Most elements that require a tabindex -- `<a href>`, `<button>`, `<input>`, `<textarea>` -- have it already.
+
+Your application might require an exception to this rule in the case of an element that captures incoming tab traversal for a composite widget. In that case, turn off this rule on a per instance basis. This is an uncommon case.
```
ljharb: Would it be hard to add an exceptions list to the config, so this use case could be covered?

src/index.js
evcohen: The next release will be a major bump, so this is safe

docs/rules/no-noninteractive-tabindex.md
```diff
@@ -0,0 +1,77 @@
...
+<a>Edit</a>
+<a href="#">Edit</a>
+<a role="button">Edit</a>
+```
+
+The bare `<a>` tag is an _anchor_. It has no semantic AX API mapping in either ARIA or the AXObject model. It's as meaningful as `<div>`, which is to say it has no meaning. An `<a>` tag with an `href` attribute has an inherent role of `link`. An `<a>` tag with an explicit role obtains the designated role. In the example above, this role is `button`.
+
+### Case: I am using "semantic" HTML. Isn't that interactive?
+
+If we take a step back into the field of linguistics for a moment, let's consider what it means for something to be "semantic". Nothing, in and of itself, has meaning. Meaning is constructed through dialogue. A speaker intends a meaning and a listener/observer interprets a meaning. Each participant constructs their own meaning through dialogue. There is no intrinsic or isolated meaning outside of interaction. Thus, we must ask, given that have a "speaker" who communicates via "semantic" HTML, who is listening/observing?
```
evcohen: nit - Thus, given a "speaker" who communicates... or Thus, given that we have a "speaker"...

docs/rules/no-noninteractive-tabindex.md
jessebeach: > Would it be hard to add an exceptions list to the config, so this use case could be covered?

I couldn't come up with a role or tag that this would map to. Maybe an exception by className?

docs/rules/no-noninteractive-tabindex.md
evcohen: nit - Thus, given a "speaker" who communicates... or Thus, given that we have a "speaker"...

docs/rules/no-noninteractive-tabindex.md
```diff
@@ -0,0 +1,77 @@
...
+  1. `radio`
+  1. `searchbox`
+  1. `switch`
+  1. `textbox`
+
+Endeavor to limit tabbable elements to those that a user can act upon.
+
+### Case: Shouldn't I add a tabindex so that users can navigate to this item?
+
+It is not necessary to put a tabindex or an `<article>`, for instance or on `<li>` items; assistive technologies provide affordances to users to find and traverse these containers. Most elements that require a tabindex -- `<a href>`, `<button>`, `<input>`, `<textarea>` -- have it already.
```
evcohen: "It is not necessary to put a tabindex *on an `<article>`, for instance, or on `<li>` items...

docs/rules/no-noninteractive-tabindex.md
```diff
@@ -0,0 +1,77 @@
...
+
+Your application might require an exception to this rule in the case of an element that captures incoming tab traversal for a composite widget. In that case, turn off this rule on a per instance basis. This is an uncommon case.
+
+### References
+
+  1. [Fundamental Keyboard Navigation Conventions](https://www.w3.org/TR/wai-aria-practices-1.1/#kbd_generalnav)
+
+## Rule details
+
+This rule takes no arguments.
```
evcohen: Looks like this does take arguments in the recommended config, but the code only reflects an empty object schema

src/rules/no-noninteractive-tabindex.js
```diff
@@ -0,0 +1,88 @@
...
+  propName,
+} from 'jsx-ast-utils';
+import isInteractiveElement from '../util/isInteractiveElement';
+import isInteractiveRole from '../util/isInteractiveRole';
+import { generateObjSchema } from '../util/schemas';
+
+const errorMessage =
+  '`tabIndex` should only be declared on interactive elements.';
+
+const schema = generateObjSchema();
```
evcohen: If we want this to reflect what the recommended config has:

```
import { generateObjSchema, arraySchema } from '../util/schemas';

const schema = generateObjSchema({
  roles: arraySchema,
  tags: arraySchema
});

File: src/rules/no-noninteractive-tabindex.js [hunk 7]
```diff
@@ -0,0 +1,88 @@
...
+  create: (context: ESLintContext) => {
+    const options = context.options;
+    return {
+      JSXOpeningElement: (
+          node: JSXOpeningElement,
+        ) => {
+        const type = elementType(node);
+        const attributes = node.attributes;
+        const tabIndexProp = getProp(attributes, 'tabIndex');
+        const tabIndex = getLiteralPropValue(tabIndexProp);
```
evcohen: We have a util to getTabIndex because its a little more complicated than it seems: `src/util/getTabIndex`

docs/rules/no-noninteractive-tabindex.md
ljharb: I guess I would expect tag names, like `div`, `MyComponent`, etc - both for avoiding warnings when they'd appear by default, or enabling them when they wouldn't normally appear.

jessebeach: > I guess I would expect tag names, like div, MyComponent, etc - both for avoiding warnings when they'd appear by default, or enabling them when they wouldn't normally appear.

This kind of exception really is a one-off special case. I don't think we can generalize to a tag or component (which just renders down to a tag eventually).

docs/rules/no-noninteractive-tabindex.md
jessebeach: Good catch!

jessebeach:
@evcohen I've made the requested changes.

PR #212, opened 2017-04-18
Add media-has-caption rule

evcohen:
Satisfies audio-has-caption in #29 but also works for `<video>` as well.

### DIFF
--- a/__tests__/src/rules/media-has-caption-test.js
+++ b/__tests__/src/rules/media-has-caption-test.js
@@ -23,6 +23,14 @@ const expectedError = {
   type: 'JSXOpeningElement',
 };
 
+const customSchema = [
+  {
+    audio: ['Audio'],
+    video: ['Video'],
+    track: ['Track'],
+  },
+];
+
 ruleTester.run('media-has-caption', rule, {
   valid: [
     { code: '<div />;' },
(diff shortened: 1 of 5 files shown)


evcohen:
@jessebeach I'm having trouble with the flow annotations for this one 😞 may need some help understanding how to resolve

evcohen:
@ljharb I was hoping to implement global settings in this major version so that we could map custom components to DOM elements across all rules, so I held off on implementing in this pass.

But you're right - I think its still valuable to provide per-rule configuration options in case devs want to override the global settings (which I can't think of a case where they would for DOM elements). I'll add the schema in this PR because it definitely makes sense to make use of both global and local config APIs

ljharb:
I don't have a use case for a custom captions prop; ours would be the same for consistency with the DOM. I wouldn't object to that being added, but it might be worth waiting til the use case comes up.

It's worth noting that if all of the tag names and prop names are configurable, then this rule should instead be a custom configuration item in a generic rule about ensuring components have certain props - so I think we want to be careful about allowing 100% configurability :-)

evcohen:
> It's worth noting that if all of the tag names and prop names are configurable, then this rule should instead be a custom configuration item in a generic rule about ensuring components have certain props - so I think we want to be careful about allowing 100% configurability :-)

Haha too true 😂

__tests__/src/rules/media-has-caption-test.js
```diff
@@ -0,0 +1,103 @@
+/* eslint-env jest */
```
ljharb: this should probably go in `__tests__/.eslintrc` so it doesn't have to be repeated in every file?

evcohen: We can do that! It's just a part of the create-rule script so it gets added to every file. Can send a separate PR to eliminate

evcohen:
> (it's your repo so your call, but i'd suggest a rebase to excise the merge commit ;-) )

Haha, I usually squash since I'm not very good at keeping a detailed history.

PR #213, opened 2017-04-18
Add ignoreNonDOM option to no-autofocus.

### LINKED ISSUE #179: Ignore composite components for no-autofocus rule
`<Foo autoFocus />` should be ignored.

evcohen:
Fixes #179

### DIFF
--- a/__tests__/src/rules/no-autofocus-test.js
+++ b/__tests__/src/rules/no-autofocus-test.js
@@ -23,11 +23,19 @@ const expectedError = {
   type: 'JSXAttribute',
 };
 
+const ignoreNonDOMSchema = [
+  {
+    ignoreNonDOM: true,
+  },
+];
+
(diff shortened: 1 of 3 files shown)


fjaguero:
Hello! I have a quick question regarding this. Does this makes sense even in a login page with two fields? The user will be moved to the "email" input, but there is nothing on top.

ljharb:
Does what make sense - using the autoFocus attribute?

fjaguero:
Yes. I mean to do it in a case where you only have a small login form.

ljharb:
Yes, even on a page that had nothing on it but a single input, autofocus wouldn't be a great idea. The [rule docs] state "Autofocusing elements can cause usability issues for sighted and non-sighted users, alike.", and the number of elements on the page has nothing to do with this.

PR #215, opened 2017-04-18
[WIP] - Implement global settings to map DOM elements to custom components

evcohen:
This is a work in progress, just wanted to post to see if everyone thought the direction for this looked OK. 

The goal here is to provide an API for devs to map DOM elements to their custom components for these rules to be checked on. These custom components should be the first wrapper around the DOM element and have a consistent API with the DOM element (i.e. if `a` maps to `Link`, `Link` needs to have `href` as a prop for `href-no-hash` to work correctly.) 

Rule configurations will override global settings (however, I can't think of a use case to override custom component naming on a per-rule level, maybe we should deprecate that in favor of bringing these to the global level).

### DIFF
--- a/docs/rules/anchor-has-content.md
+++ b/docs/rules/anchor-has-content.md
@@ -13,13 +13,13 @@ This rule takes one optional object argument of type object:
 {
     "rules": {
         "jsx-a11y/anchor-has-content": [ 2, {
-            "components": [ "Anchor" ],
+            "a": [ "Anchor" ],
           }],
     }
 }
 ```
 
-For the `components` option, these strings determine which JSX elements (**always including** `<a>`) should be checked for having content. This is a good use case when you have a wrapper component that simply renders an `a` element (like in React):
+For the `a` option, these strings determine which JSX elements (**always including** `<a>`) should be checked for having content. This is a good use case when you have a wrapper component that simply renders an `a` element (like in React):
 
 ```js
 // Anchor.js
--- a/__tests__/src/rules/anchor-has-content-test.js
+++ b/__tests__/src/rules/anchor-has-content-test.js
@@ -10,8 +10,7 @@
 
 import { RuleTester } from 'eslint';
 import parserOptionsMapper from '../../__util__/parserOptionsMapper';
-import rule,
-  { determineChildType } from '../../../src/rules/anchor-has-content';
+import rule, { determineChildType } from '../../../src/rules/anchor-has-content';
 
 // -----------------------------------------------------------------------------
 // Tests
@@ -27,9 +26,7 @@ const expectedError = {
 describe('determineChildType', () => {
   describe('default case', () => {
     it('should return false', () => {
-      expect(
-        determineChildType({}),
-      ).toBe(false);
+      expect(determineChildType({}, {})).toBe(false);
     });
   });
 });
--- a/__tests__/src/rules/heading-has-content-test.js
+++ b/__tests__/src/rules/heading-has-content-test.js
@@ -10,8 +10,7 @@
 
 import { RuleTester } from 'eslint';
 import parserOptionsMapper from '../../__util__/parserOptionsMapper';
-import rule,
-  { determineChildType } from '../../../src/rules/heading-has-content';
+import rule, { determineChildType } from '../../../src/rules/heading-has-content';
 
 // -----------------------------------------------------------------------------
 // Tests
@@ -27,12 +26,16 @@ const expectedError = {
 describe('determineChildType', () => {
   describe('default case', () => {
     it('should return false', () => {
-      expect(
-        determineChildType({}),
-      ).toBe(false);
+      expect(determineChildType({}, {})).toBe(false);
     });
   });
 });
+
+const customTypeSchema = [
+  {
+    components: ['Heading'],
+  },
+];
 
 ruleTester.run('heading-has-content', rule, {
   valid: [
@@ -49,10 +52,12 @@ ruleTester.run('heading-has-content', rule, {
     { code: '<h1>{foo.bar}</h1>' },
     { code: '<h1 dangerouslySetInnerHTML={{ __html: "foo" }} />' },
     { code: '<h1 children={children} />' },
+    { code: '<Heading>{foo}</Heading>', options: customTypeSchema },
   ].map(parserOptionsMapper),
   invalid: [
     { code: '<h1 />', errors: [expectedError] },
     { code: '<h1><Bar aria-hidden /></h1>', errors: [expectedError] },
     { code: '<h1>{undefined}</h1>', errors: [expectedError] },
+    { code: '<Heading>{undefined}</Heading>', options: customTypeSchema, errors: [expectedError] },
   ].map(parserOptionsMapper),
 });
--- a/__tests__/src/rules/href-no-hash-test.js
+++ b/__tests__/src/rules/href-no-hash-test.js
@@ -23,16 +23,22 @@ const expectedError = {
   type: 'JSXOpeningElement',
 };
 
-const components = [{
-  components: ['Anchor', 'Link'],
-}];
-const specialLink = [{
-  specialLink: ['hrefLeft', 'hrefRight'],
-}];
-const componentsAndSpecialLink = [{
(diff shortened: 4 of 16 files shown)


jessebeach:
A few high-level questions:

1. Would this type of composition checking not be better located in a system like `PropTypes` in React or Flow? 
1. Is it necessary to limit this to components that directly return HTML? Essentially, what the mapping says is "Treat XComponent like `<a href />`. 

What we might do instead is treat the map key more like a strong association:

```
`div[role="button"]`: ['FancyButton']
```

Which would throw on something like: `<FancyButton role="heading" />`

I'm having trouble coming up with use cases, though. Could you list a few you have in mind?

ljharb:
Airbnb's extensive use cases include, we have custom `Video` and `Track` and `Source` elements, we have `Link` and `Button`, and a number of others - all of whose purpose is to prevent developers from using HTML directly and having to know details about a11y, HTML, CSS, etc. We have lint/review rules that prevent direct usage of HTML tags. By telling this plugin to, for example, treat `Button` as a `button`, we can keep prop names consistent, and still get the benefit of linting errors in product developer code, instead of in infra developer code.

docs/rules/anchor-has-content.md
```diff
@@ -13,13 +13,13 @@ This rule takes one optional object argument of type object:
 {
     "rules": {
         "jsx-a11y/anchor-has-content": [ 2, {
-            "components": [ "Anchor" ],
+            "a": [ "Anchor" ],
```
ljharb: i don't think we should be changing the schema for individual rules, especially since tag names here might go right alongside *not* tag names, so even if we have strict schema validation (which we should), it's more confusing and less intuitive.

I think that individual rule component overrides should remain, but should be ignored when a global config exists for a given tag name. If we later want to make per-rule overrides, then we should do that via a *new* per-rule schema mechanism, to avoid conflating things with the existing non-cohesive ones.

evcohen:
> Would this type of composition checking not be better located in a system like PropTypes in React or Flow?

Not 100% sure what you mean here, sorry :/. We can potentially enforce that a wrapper component exposes necessary a11y-related props in its API via `PropTypes` or Flow, but I'm unsure how it is related to ensuring that an instance of that component in the wild actually passes down those props. My goal with this was to free ourselves from just applying these rules to DOM elements and start running them on custom components, as well. We just need to know what the custom components map to in any given application. We are probably missing a lot of rules related to `<a>` because many people are using `Link` from react-router and its just being passed over.

> Is it necessary to limit this to components that directly return HTML? Essentially, what the mapping says is "Treat XComponent like <a href />.

No, it isn't necessary, but I think it's a good first pass for simplicity's sake. My concern is that we could have lots of false errors because we don't actually know if the underlying component is actually using the props it's passing down. `<FancyButton role="heading" />` could just be dropping that role prop and we would have no idea. This could cause developer friction and ultimately result in people turning off rules.

evcohen:
I may be crazy but I've been thinking about this a little more and I think we can do better. I'm not 100% sure how we can do this yet, but if there is a way to build this map internally I think it could drastically improve the DX for this tool. Developers having to maintain a global map of their components and what they render could get quite cumbersome, especially if things are changing often. I can this working fine for stable apps like Airbnb, but more trouble than its worth for apps that are in their early stages. 

So, the idea here would be to statically analyze and detect the root element of a component by looking into the render method and seeing the first type of what it returns. If the root is a valid DOM element (not a custom component) then we add it to an in-memory map.

Outstanding questions:

1. Can we preprocess a project to build the map?
2. Can we keep state across all rules?

If the answer to the first 2 of these is yes, I think that this could be achieved simply and well. Given a .js* file, first detect that we're working with a (p)react component. Next, find its render method. Finally, find the first JSX element that it returns. If it is a DOM element, then add to list of custom components that maps to that DOM element; if not, drop it.


Thoughts?

jessebeach:
> Airbnb's extensive use cases include, we have custom Video and Track and Source elements, we have Link and Button, and a number of others - all of whose purpose is to prevent developers from using HTML directly and having to know details about a11y, HTML, CSS, etc. We have lint/review rules that prevent direct usage of HTML tags. By telling this plugin to, for example, treat Button as a button, we can keep prop names consistent, and still get the benefit of linting errors in product developer code, instead of in infra developer code.

What props would be set on `<Link>` or `<Button>` that aren't already on the HTML that element that these components return? I'd think that `<Button onClick>` => `<button onClick>`. 

What does one need to specify on `<Link>` to `<Button>` to ensure that it ends up being accessible in your specific case?

I'm not opposed to this capability. I just don't understand it yet.

ljharb:
> statically analyze and detect the root element of a component by looking into the render method and seeing the first type of what it returns

This absolutely won't work when the inner component is dynamic (`const tag = isLink ? 'a' : 'div'; … return <tag />`), or when the outer component is the one I want linted and the HTML element is multiple layers deep.

I think that, in fact we can't do better, and to try to will make things worse.

I would also posit that we should start with an explicit map of components, and then in parallel investigate how to begin inferring that information without the explicit map - rather than preventing valid use cases while we chase solutions that are impossible to fully achieve statically in a dynamic language.

ljharb:
@jessebeach you're exactly right; for consistency we'd keep the props identical. However, the actual `Button` component wouldn't live in the main website repo - it's in a separate package for our core design components. Running eslint on product-devs' usage of it won't ever reach the HTML tags, and it's those developers that most need the linter (as opposed to the infra devs who are building `Button` and know about a11y concerns, and also have eslint to check them in that repo).

Certainly I could build lots of custom propTypes on my custom components to duplicate all of the checks in this plugin - and that would work without requiring this plugin to know or care about my custom components - but the likely scenario is that if it's difficult and requires porting the logic, we simply won't allot the resources to make the checks, and a11y will suffer as a result.

jessebeach:
> the actual Button component wouldn't live in the main website repo

Aha, I understand now. We've been addressing this issue via polymorphism and Flow types. 

```
type Props = {
  label: string,
  pressed: boolean,
};

type DefaultProps = {
  pressed: boolean,
};

class ToggleButton extends React.Component<DefaultProps, Props, void> {
  props: Props;
  static defaultProps: DefaultProps = {
    pressed: false,
  };
  render() {
    const {
      label,
      pressed,
      ...rest,
    } = this.props;
    return (
      <div
        role="button"
        aria-label={label}
        aria-pressed={pressed}
        {...rest}>
        {this.props.children}
      </div>
    );
  }
}
```

Declaring the Prop types shape via Flow in this toy example makes `label` required, but `pressed` is not (because it's given a default value). Flow is doing static analysis across a dependency graph, so the types flow (!) into other files. It's how we're slowly making the `label` prop required across different button flavors in our internal components. 

I get why you'd want to do this via linting instead. I don't think we'd used the feature, but I'm happy to review and help get it integrating in.

jessebeach:
> I may be crazy but I've been thinking about this a little more and I think we can do better. I'm not 100% sure how we can do this yet, but if there is a way to build this map internally I think it could drastically improve the DX for this tool.

Love the idea, but I think this is a different tool. Let's get this one squared away and settled, then we can dig into it :)

jessebeach:
@evcohen are you still working on this?

jessebeach:
@evcohen still working on this? I'd love to get our PR queue down to in-progress items to start out 2018.

evcohen:
@jessebeach happy to finish this off if we still think this is the right direction to go in. I am unsure how often this will be used since it's some work for the end-user to maintain this map correctly. Ideally, I would want to come up with a solution where we can figure out a component -> DOM element map ourselves.

jessebeach:
> @jessebeach happy to finish this off if we still think this is the right direction to go in.

Some thoughts:

  1. If a custom component renders down to HTML, then we'll probably already flag issues at the point where the HTML is defined.
  2. We don't want folks needlessly adding ARIA roles and props to custom components if those requirements can be easily set via Flow or React types
  3. It strikes me that this mapping could be used incorrectly and lead to incorrect application of props or overwriting correct props in composed components.

I haven't had a strong use case for this feature in the few months since it was proposed. That indicates to me at least that you can safely abandon it.

evcohen:
I agree, let's hold off on this until we find a major need.

PR #218, opened 2017-04-19
Add options to element interaction rules

jessebeach:
Add options to `no-static-element-interactions` and `no-noninteractive-element-interactions` rules that allow an implementor to specify the handlers that should be considered when triggering the rule. 

I plan to start with just `onClick` for FB initially and expand to more handlers as I fix violations and understand patterns of use better. The options allow for a slower roll-out and less developer friction.


src/index.js
```diff
@@ -151,7 +169,7 @@ module.exports = {
         'jsx-a11y/no-noninteractive-tabindex': 'error',
         'jsx-a11y/no-onchange': 'error',
         'jsx-a11y/no-redundant-roles': 'error',
-        'jsx-a11y/no-static-element-interactions': 'warn',
+        'jsx-a11y/no-static-element-interactions': 'error',
```
ljharb: just highlighting that these config changes make the PR semver-major

src/rules/no-noninteractive-element-interactions.js
```diff
@@ -30,64 +30,64 @@ import isPresentationRole from '../util/isPresentationRole';
...
 
-const schema = generateObjSchema();
-
 const domElements = [...dom.keys()];
-const interactiveProps = [
-  ...eventHandlersByType.mouse,
-  ...eventHandlersByType.keyboard,
-];
+const defaultInteractiveProps = eventHandlers;
+const schema = arraySchema;
```
ljharb: this is kind of a useless rename - could we either import `arraySchema as schema`, or rename uses of `schema` here to `arraySchema`?

src/rules/no-static-element-interactions.js
```diff
@@ -31,62 +31,62 @@ import isPresentationRole from '../util/isPresentationRole';
...
 
-const schema = generateObjSchema();
-
 const domElements = [...dom.keys()];
-const interactiveProps = [
-  ...eventHandlersByType.mouse,
-  ...eventHandlersByType.keyboard,
-];
+const defaultInteractiveProps = eventHandlers;
+const schema = arraySchema;
```
ljharb: same here

src/rules/no-noninteractive-element-interactions.js
evcohen: also, we should keep all rule configuration schemas as objects. objects can be easily extended without breaking changes, whereas arrays are rigid and less configurable. if we wanted to add an option for something else down the line, a simple change would cause a semver major which is bleh 😄 

I think it should be:

```
generateObjSchema({
  handlers: arraySchema
});
```

src/rules/no-static-element-interactions.js
evcohen: same re: object

src/index.js
jessebeach: Definitely. Master since 21435147ee29bf9efa0aeda7ff328bea25b247ae has been semver-major upgrade-necessary.

src/rules/no-noninteractive-element-interactions.js
evcohen: on another note, should this be a blacklist or a whitelist?

jessebeach:
> on another note, should this be a blacklist or a whitelist?

There are 68 possible handler props. The whitelist approach is less prone to accidental omitting and is shorter to specify. The recommended configuration provides a good baseline. It's where I want to get to in the FB codebase. I've found that elements with `onMouseOver` and `onMouseOut` aren't always the best site to also include focus/blur handling. Guidance here isn't always straightforward.

jessebeach:
@evcohen requested changes have been made. The argument to keep the options as objects is a good one. I'll keep it in mind for future changes.

jessebeach:
> LGTM! ah one more thing- lets add this to the docs

I am literally the worst at remembering to do this!

jessebeach:
@evcohen docs updated.

PR #219, opened 2017-04-23
Add prefer-button rule

AlmeroSteyn:
This is a proposal to add a rule to check for `onClick` handlers on `anchors` and hint to rather use a `button` element.

Still outstanding: 
1. Documentation: Will document this fully if this proposal gets accepted in principle.
2. Complete href checks: Companion to this rule there should be a check that extends the current check for "#' in the `href` to also warn when no `href` is present or when something like `javascript:void(0)` is used. Will also make these changes if this proposal gets accepted in principle.

I have also added the check to the strict rules section only, pending discussion.

Looking forward to hearing your ideas and opinions around this proposal.

P.S. Following issue has been solved below: 

On a clean clone the following rule in the `eslint-config-airbnb-base` file causes an error on just about every single line of the repo on Windows:

```
    // disallow mixed 'LF' and 'CRLF' as linebreaks
    // http://eslint.org/docs/rules/linebreak-style
    'linebreak-style': ['error', 'unix'],
```
Also the `flow` script in `package.json` fails with an `! was unexpected at this time.` error.

I could work around it but should I raise an issue? Or do would you rather stay with the current config and propose a workaround for Windows devs in the docs?

### DIFF
--- a/docs/rules/no-noninteractive-element-interactions.md
+++ b/docs/rules/no-noninteractive-element-interactions.md
@@ -55,6 +55,40 @@ Headers often double as expand/collapse controls for the content they headline. 
 </ul>
 ```
 
+### Case: This element is a table cell
+
+Table cells (and tables in general) are meant to contain data. ARIA provides us with a construct called a [Grid](http://w3c.github.io/aria-practices/#grid) that is essentially a 2 dimensional logical container for content and interactive elements.
+
+You have two options in this case.
+
+#### Option 1, move the interactive content inside the table cells
+
+For instance, move the button inside the cell:
+
+```
+<table>
+  <tr>
+    <td><button>Sort</button></td>
+  </tr>
+</table>
+```
+
+This preserves the table cell semantics and the button semantics; the two are not conflated on the cell.
+
+#### Option 2, convert the table into an ARIA grid
+
+If you're user interface has a table-like layout, but is filled with interactive components in the cells, consider converting the table into a grid.
+
+```
+<table role="grid">
+  <tr>
+    <td role="gridcell" onClick={this.sort}>Sort</td>
+  </tr>
+</table>
+```
+
+You can also put the interactive content inside the grid cell. This maintains the semantic distinction between the cell and the interaction content, although a grid cell can be interactive.
+
 ### References
 
   1. [WAI-ARIA roles](https://www.w3.org/TR/wai-aria-1.1/#usage_intro)


ljharb:
@AlmeroSteyn you can fix your windows issues by setting up git to automatically convert between LF and CRLFs, and then delete and re-clone all your repos. I'm not aware of a way around doing that for Windows users for general dev across platforms.

AlmeroSteyn:
@ljharb Aaaah! That'll sort it, thank you.

src/rules/prefer-button.js
```diff
@@ -0,0 +1,45 @@
...
+    JSXOpeningElement: (node: JSXOpeningElement) => {
+      const type = elementType(node);
+
+      if (type && type !== 'a') {
+        return;
+      }
+
+      const anchorClick = getProp(node.attributes, 'onClick');
+
+      if (!anchorClick) {
```
jessebeach: The "anchorness" rests on the presence of an `href` attribute, right?

jessebeach: Do you believe that it is never proper to have event handlers on a `<a href>` element? Are there any handlers that would be acceptable?

I have lots of comments about the nuts and bolts of the code in terms of modules and patterns for this project. Once we get past the conceptual issues, we'll focus on the code; but don't worry about that just yet.

AlmeroSteyn: Never thought of it that way but in a sense that could be true. However, I think the meaning is primarily portrayed by the element itself, just like all semantic elements portray the element's meaning to many. But if you remove the `href`, or provide an invalid value, the browser does indeed strip its "anchorness" in essence. Even adding a `tabindex` to it again just makes a screen reader read 'clickable`.

Dunno if I am answering the question here :-)

AlmeroSteyn: Can only speak from personal experience but I can't think of, with the availability of `button`, that I have needed to add event handlers to an `anchor`. That, of course, does not say that I am saying there is no use case on the web ever where some event handler could be useful, just can't think of a use case I can't solve differently with standard HTML or React with less code and ARIA.

But the `onClick` event seems to be a surefire indication that the anchor is running script as well. Even with a proper `href` present I find it to be bothersome as it navigates away while running some script. With the exception of hash links (whereI can't think of a reason to run code when navigating in-page) the React lifecycle hooks always served to trigger any code that may be needed to cleanup the expiring page's components or setup of the new page.

This is, of course, coming from the idea that anchors navigate and therefore causes a complete context change instantly.

These days I often find myself going to this article, that explains a lot better than I can here how I view links and buttons: [Buttons vs Links](https://marcysutton.com/links-vs-buttons-in-modern-web-applications/)

But that said I do concede that it is not yet a totally cut a dried discussion in even the accessibility community as a whole.

jessebeach: > However, I think the meaning is primarily portrayed by the element itself, just like all semantic elements portray the element's meaning to many.

Meaning is both conveyed and interpreted :) In this case, the thing doing the interpreting, that AX layer, treats `<a>` as text. The `<a href="#">` element is upcast to a link:

a_pen_by_j__renee_beach_and_messenger

`<a role="button">` and `<a href="#" role="button">` are semantically indistinguishable:

a_pen_by_j__renee_beach

accessibility_internals

A browser does strip the 'anchorness' without an href :)

jessebeach: Gotcha. So this rule is enforcing that "links navigate; buttons do". How will it treat `<a href="#" role="button">`, which is semantically a button?

ljharb: Ideally it would forbid that. If it's semantically a button, it should be a button.

jessebeach: That's the hook then. If you're putting `role="button"` on elements, well, it should just be a `<button>` then. The documentation can add some information about how to knock out styling and deal with the inline-block quirkiness.

evcohen: > The documentation can add some information about how to knock out styling and deal with the inline-block quirkiness.

like such, @jessebeach? is the following better than `<a /* style */ href="#foo" role="button" onClick={handler}>Link</a>`?  ->

```jsx
<div
  // Add styling to this?
  role="button"
  onClick={() => {}}
  onKeyPress={() => {}}
  tabIndex="0">
```

`react-router` uses an anchor with a click handler fwiw: https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/modules/Link.js#L65

AlmeroSteyn: Thanks for the explanation! I am aware of how the roles of elements translate to the accessibility tree and am in no way disputing that ARIA works and that you can create accessible solutions using ARIA. 

However, I think that the elements also convey meaning to humans or we could build almost the entire web out of `div` with added handlers and roles. Many developers, reviewers, and testers also attach meaning to the semantics of HTML even though the element may be totally redone with ARIA it now takes more time to read and understand and implement.

In fact, as you know, even the ARIA guidelines recommend not doing this and using ARIA sparingly (rule 1) and to avoid changing the role of a semantic element (rule 2).

The spirit of me submitting this proposal is not to say ARIA should be avoided but to support the ARIA principles, hint towards less code and leaning more heavily on browser standard implementation and to try and help dispell the view by many product owners and managers (also devs and designers, mind you) that accessible pages costs a lot more work to make.

The example presented above is only a semantic button, but not a functional button. For that I have to do:
```
const DemoButtonLink = ({ onClickHandler }) => (
  <a role="button"
     tabIndex="0"
     className="style-me-as-button-and-set-mouse-pointer"
     onClick={onClickHandler}
     onKeyDown={(e) => {
       if (e.keyCode === 32 || e.keyCode === 13) {
         onClickHandler();
       }
     }}>Button test</a>
);
```
That is indeed a totally functional button element. But if hinted in this direction and the developer forgets to add the space key check, the button is not fully accessible anymore.  That's a lot to remember and possible points of failure on just one button.

Compare this to:
```
const DemoButton = ({ onClickHandler }) => (
  <button onClick={onClickHandler}>Button test</button>
);
```

Agreed, both achieve the same result but with the second I think I can convince even the most difficult accessibility critic (if they care about this level of detail) that accessibility does not cost more while getting a fully functional accessible button with all bells and whistles attached out of the box every time.

Therefore I would think that there could be space for a rule like this in a linter specifically dealing with accessibility. For JavaScript code our linters are pretty strict to ensure good, solid code. They can even fail entire build if spaces or line feeds are wrong. I would personally not find it too weird that an accessibility linter would recommend using a button as a button, especially if it also cuts down the lines of code I have to write and remember in the process.

But, as said before, these are my opinions only and this is only a proposal :-)

AlmeroSteyn: > That's the hook then. If you're putting role="button" on elements, well, it should just be a button then.

> react-router uses an anchor with a click handler fwiw: https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/modules/Link.js#L65

@evcohen Ah yeah there's a use case. Dealing with context things on navigation. 

The check for just `onClick` is therefore not always valid. 

The `role="button"` trigger could be more sane but thats kinda like Tortuga in the Pirates movies, can only be found by those who already know where it is. A dev will generally need to know about ARIA already to have added this to be warned about ARIA.

Would a combination of `role="button"` and / or a missing or invalid `href` value perhaps do the job? Or sticking to `onClick` in conjunction with a missing or invalid `href` while allowing `onClick` if the href is properly filled?

jessebeach:
`<a href="#" onClick />` is a pretty solid indication that this element is
being used a button. That's with the href set to '#'. The href content
could be made configurable.

ljharb:
An href of "#" or an href starting with "javascript:" should both be flagged as "should be a button", imo. Anything else could be a link with progressively-enhanded onClick behavior, which could be fine.

evcohen:
> An href of "#" or an href starting with "javascript:" should both be flagged as "should be a button", imo. Anything else could be a link with progressively-enhanded onClick behavior, which could be fine.

If we are going in the direction with a more semantic rule like this, should we deprecate `href-no-hash` and move that logic into here? This can report on the following:

`<a href="#" />` -> use an href that is a valid reference to something a user can navigate to

`<a href="javascript:void(0)" />` -> use an href that is a valid reference to something a user can navigate to

`<a href="#" onClick={foo} />` -> use a button

`<a href={somethingValid} role="button" onClick={foo} `> -> use a button

`<a href={somethingValid} onClick={foo} />` -> PASSES

is this correct?

ljharb:
I'd say `<a href="javascript:void(0)" />` is "use a button", not "use a better link" - otherwise that sounds good.

jessebeach:
Likewise, sounds good.

AlmeroSteyn:
If we are combining the two validations, maybe the name should change as it then covers more than hinting at buttons.

How about `anchor-is-valid` with the rules mentioned above?

And what about `anchor-has-content`? Should that be pulled in as well? 

Will make these changes in the coming day or three in the PR.

So as I see it (for now without `anchor-has-content`):

**Should be button**
```
<a href="javascript:void(0)" />
<a href="javascript:void(0)" onClick={foo} />
<a href="#" onClick={foo}/>
<a onClick={foo}/>
```

Add a valid href
```
<a href="#" />
<a>
```

PASSES
```
<a href={somethingValid} />
<a href={somethingValid} onClick={foo} />
```

Thank you all for considering this, by the way!

ljharb:
`button-is-button`?

AlmeroSteyn:
I have added the logic discussed.

Unfortunately, I seemed to have accidentally pulled some old commits into the history here. My latest commit shows only the files I have changed.

If required I can open up a new pull request with a clean history and use the new rule name.

Does this cover everything? And do you think we should add the empty anchor check here?

evcohen:
@AlmeroSteyn can you rebase? I think a bunch of @jessebeach commits are in here not sure what happened

ljharb:
It looks like she's just the pusher; i.e., she pushed their commits (two avatars show up)

AlmeroSteyn:
@evcohen It happened when i squashed my commits. Something went wrong so it pulled these old commits in. Will see what I can do in the morrow. My very last commit contains all the changes related to this PR.

AlmeroSteyn:
Hmm sorry everyone, in trying to fix it I properly broke it. Reset back to upstream as the commits were already registered in the rebase and rebasing did not seem to work and that autoclosed this PR. So I had to create a new one in #224. Sorry for the inconvenience.

¯\(°_°)/¯

jessebeach:
I find `git rebase -i master` to be very helpful when history gets wrinkled. It let's you choose the commits to include in a rebase, so you can ignore the extra commits that are already in the base.

ljharb:
I use `git rebase -i origin/master` (and usually just always delete my local master) to ensure that i'm always using the latest commits.

AlmeroSteyn:
@jessebeach @ljharb Thanks! Oh the dark art of Git history (which loosely translates to I flipped every switch Google could find till I broke it and it did not include this switch so I have made an immediate note of it for future).

PR #220, opened 2017-04-25
Refactor img-has-alt to alt-text

evcohen:
Suggested in #29, we can consolidate `img-has-alt` to `alt-text` to support all elements that need alternative text.


src/rules/alt-text.js
```diff
@@ -0,0 +1,205 @@
...
+  area: arraySchema,
+  'input[type="image"]': arraySchema,
+});
+
+const ruleByElement = {
+  img(context, node) {
+    const nodeType = elementType(node);
+    const roleProp = getProp(node.attributes, 'role');
+    const roleValue = getPropValue(roleProp);
+    const isPresentation = roleProp && typeof roleValue === 'string'
```
jessebeach: There's an `isPresentationRole` module in utils that checks `presentation` and `none`.

src/rules/alt-text.js
```diff
@@ -0,0 +1,205 @@
...
+      && roleValue.toLowerCase() === 'none';
+
+    const altProp = getProp(node.attributes, 'alt');
+
+    // Missing alt prop error.
+    if (altProp === undefined) {
+      if (isPresentation || isRoleNone) {
+        context.report({
+          node,
+          message: `Prefer alt="" over role="${roleValue}". First rule of aria is to not use aria if it can be achieved via native HTML.`,
```
jessebeach: You make ARIA sound like a fight club =D

src/rules/alt-text.js
```diff
@@ -0,0 +1,205 @@
...
+  meta: {
+    docs: {},
+    schema: [schema],
+  },
+
+  create: context => ({
+    JSXOpeningElement: (node) => {
+      const options = context.options[0] || {};
+      // Elements to validate for alt text.
+      const elementOptions = options.elements || DEFAULT_ELEMENTS;
```
jessebeach: You can process the options once in the body of the create function rather than in the AST Node visitor function... I *think*. `context` shouldn't be changing between visitor calls, unless it's a reference to a mutable object somewhere else? @ljharb thoughts?

src/rules/alt-text.js
```diff
@@ -0,0 +1,205 @@
...
+      // Get custom components for just the elements that will be tested.
+      const customComponents = elementOptions
+        .map(element => options[element])
```
jessebeach: Ditto to above.

src/rules/alt-text.js
```diff
@@ -0,0 +1,205 @@
...
+
+      // Map nodeType to the DOM element if we are running this on a custom component.
+      if (elementOptions.indexOf(DOMElement) === -1) {
+        DOMElement = elementOptions.find((element) => {
+          const customComponentsForElement = options[element] || [];
+          return customComponentsForElement.indexOf(nodeType) > -1;
+        });
+      }
+
+      ruleByElement[DOMElement](context, node);
```
jessebeach: I *think* everything above here can be processed in the `create` function body.

evcohen: Ah, i got most of it out of the visitor function- need reference to the element type for some of it. but this is good advice overall, I dont think most of the rules do this so we can optimize! learned something new today! thanks 😄

__mocks__/genInteractives.js
```diff
@@ -13,37 +13,37 @@ const domElements = [...dom.keys()];
...
-  'input[type=\"reset\"]': [{prop: 'type', value: 'reset'}],
-  'input[type=\"search\"]': [{prop: 'type', value: 'search'}],
-  'input[type=\"submit\"]': [{prop: 'type', value: 'submit'}],
-  'input[type=\"tel\"]': [{prop: 'type', value: 'tel'}],
-  'input[type=\"text\"]': [{prop: 'type', value: 'text'}],
-  'input[type=\"time\"]': [{prop: 'type', value: 'time'}],
-  'input[type=\"url\"]': [{prop: 'type', value: 'url'}],
-  'input[type=\"week\"]': [{prop: 'type', value: 'week'}],
-  menuitem:[],
+  'input[type=\"button\"]': [{ prop: 'type', value: 'button' }],
```
evcohen: A lot of these changes are just my editor fixing lint errors on save. I don't think the lint script checks `__mocks__` but can fast-follow to resolve this. just didn't want to add more cruft to this PR

README.md
```diff
@@ -104,7 +105,6 @@ You can also enable all the recommended rules at once. Add `plugin:jsx-a11y/reco
 - [href-no-hash](docs/rules/href-no-hash.md): Enforce an anchor element's `href` prop value is not just `#`.
 - [html-has-lang](docs/rules/html-has-lang.md): Enforce `<html>` element has `lang` prop.
 - [iframe-has-title](docs/rules/iframe-has-title.md): Enforce iframe elements have a title attribute.
-- [img-has-alt](docs/rules/img-has-alt.md): Enforce that `<img>` JSX elements use the `alt` prop.
```
ljharb: is there a semver-major coming up? if not, can we just deprecate this rule instead of removing it, to make this PR semver-minor?

__mocks__/genInteractives.js
```diff
@@ -13,37 +13,37 @@ const domElements = [...dom.keys()];
 const roleNames = [...roles.keys()];
 
 const interactiveElementsMap = {
-  a: [{prop: 'href', value: '#'}],
-  area: [{prop: 'href', value: '#'}],
+  a: [{ prop: 'href', value: '#' }],
```
ljharb: it's unfortunate there's unrelated whitespace changes in this PR :-/

README.md
evcohen: ya was going to publish major after this, master already riddled with breaking changes

__mocks__/genInteractives.js
evcohen: yes sorry! `__mocks__` directory isn't linted, so when I had to get in here and saved the file, `eslint --fix` ran in my text editor 😞

evcohen:
ping @jessebeach @ljharb - hoping to get this in so we can publish v5.0.0 soon!

src/rules/alt-text.js
```diff
@@ -0,0 +1,202 @@
...
+    // Get custom components for just the elements that will be tested.
+    const customComponents = elementOptions
+      .map(element => options[element])
+      .reduce(
+        (components, customComponentsForElement) => components.concat(
+          customComponentsForElement || [],
+        ),
+        [],
+      );
+    const typesToValidate = []
```
ljharb: should this be passed into `new Set` instead, to make lookups O(1)?

src/rules/alt-text.js
```diff
@@ -0,0 +1,202 @@
...
+        return type;
+      });
+
+
+    return {
+      JSXOpeningElement: (node) => {
+        const nodeType = elementType(node);
+        if (typesToValidate.indexOf(nodeType) === -1) { return; }
+
+        let DOMElement = nodeType;
```
ljharb: this could be collapsed into a `const` and a ternary.

evcohen: would violate `no-nested-ternary` lint rule, but can do. thought this was more readable than

```
        const DOMElement = nodeType === 'input'
          ? 'input[type="image"]'
          : elementOptions.has(nodeType) === false
              ? elementOptions.find((element) => {
                const customComponentsForElement = options[element] || [];
                return customComponentsForElement.indexOf(nodeType) > -1;
              })
              : nodeType;
```

ljharb: ah, for the const that's true. ¯\\\_(ツ)_/¯

docs/rules/alt-text.md
```diff
@@ -0,0 +1,133 @@
...
+// Header.js (for example)
+...
+return (
+  <header>
+    <Image alt="Logo" src="logo.jpg" />
+  </header>
+);
+```
+
+Note that passing props as spread attribute without explicitly the necessary accessibility props defined will cause this rule to fail. Explicitly pass down the set of props needed for rule to pass. Use `Image` component above as a reference for destructuring and applying the prop. **It is a good thing to explicitly pass props that you expect to be passed for self-documentation.** For example:
```
jessebeach: `without explicitly the necessary`

Seems to be missing a word.

docs/rules/alt-text.md
```diff
@@ -0,0 +1,133 @@
...
+
+Enforce that all elements that require alternative text have meaningful information to relay back to the end user. This is a critical component of accessibility for screenreader users in order for them to understand the content's purpose on the page. By default, this rule checks for alternative text on the following elements: `<img>`, `<area>`, `<input type="image">`, and `<object>`.
+
+#### Resources
+  1. [aXe object-alt](https://dequeuniversity.com/rules/axe/2.1/object-alt)
+  2. [aXe image-alt](https://dequeuniversity.com/rules/axe/2.1/image-alt)
+  3. [aXe input-image-alt](https://dequeuniversity.com/rules/axe/2.1/input-image-alt)
+  4. [aXe area-alt](https://dequeuniversity.com/rules/axe/2.1/area-alt)
+
+## How to resolve
```
jessebeach: Can you add a small section about using an image as an icon for a button or control? We should recommend using alt="" in this case. Icons are not meant to have labels; we don't want to encourage developers to add labels *all* images, thereby creating duplication in verbosity in control component labels.

jessebeach:
145 problems in ~33K components for an error rate of 0.4%.

Looking at some of the errors, I realized that we need a section in the documentation for this rule that suggests using `alt=""` on icons so that we're not encouraging duplicate or verbose labels on control elements.

With that addition to the documentation, it's definitely a rule I'll probably turn on in FB codebase.

evcohen:
@jessebeach let me know if the documentation update works, don't want to get your words mixed up 😄 will merge and publish if its ok! (sidenote: this changelog is gonna suck)

docs/rules/alt-text.md
```diff
@@ -10,7 +10,7 @@ Enforce that all elements that require alternative text have meaningful informat
 
 ## How to resolve
 ### `<img>`
-An `<img>` must have the `alt` prop set with meaningful text or as an empty string to indicate that it is an image for decoration.
+An `<img>` must have the `alt` prop set with meaningful text or as an empty string to indicate that it is an image for decoration. For images that are being used as icons for a button or control, the `alt` prop should be set to an empty string (`""`), as well, since icons are not meant to have duplicative labels.
```
jessebeach:
```
For images that are being used as icons for a button or control, the `alt` prop should be set to an empty string.

```
<button>
  
  Save
</button>
```

The content of an alt attribute is used to calculate the accessible label of a element when the text content is used to produce a label. For this reason, adding a label to an icon can produce a confusing or duplicated label on a control that already has appropriate text content.
```

jessebeach: Maybe something like that?

evcohen: you're very good at writing docs 👍 will update

felquis:
Not sure if it broke my project, I had to put `"jsx-a11y/img-has-alt": 0` in .eslint for the moment.

It was breaking in any file like this one...
```js
import React from 'react'
import { Link } from 'react-router'
import classNames from 'classnames'
import styles from './styles.css'

const MenuSubItem = (props) => {
  const subItemClasses = classNames(styles.subItem, {
    [styles.active]: props.active,
  })

  return (
    <Link
      className={subItemClasses}
      to={props.to}
    >
      {props.label}
    </Link>
  )
}

MenuSubItem.propTypes = {
  to: React.PropTypes.string.isRequired,
  label: React.PropTypes.string.isRequired,
  active: React.PropTypes.bool.isRequired,
}

export default MenuSubItem
```

@evcohen any suggest?

evcohen:
@felquis either downgrade to v4 or change your eslint config to have `jsx-a11y/alt-text: 2`

PR #224, opened 2017-05-01
Add anchor-is-valid rule

AlmeroSteyn:
Continuing where #219 left off.

Summary:

Adding rule to check anchor validity:

**Should be button**
```
<a href="javascript:void(0)" />
<a href="javascript:void(0)" onClick={foo} />
<a href="#" onClick={foo}/>
<a onClick={foo}/>
```

Add a valid href
```
<a href="#" />
<a>
```

PASSES
```
<a href={somethingValid} />
<a href={somethingValid} onClick={foo} />
```


src/index.js
```diff
@@ -148,6 +149,7 @@ module.exports = {
       rules: {
         'jsx-a11y/accessible-emoji': 'error',
         'jsx-a11y/anchor-has-content': 'error',
+        'jsx-a11y/anchor-is-valid': 'error',
```
jessebeach: Make sure the rule is added to both the recommended and strict sections.

ljharb: Just a heads up that adding enabled rules to exported configs is a semver-major change.

AlmeroSteyn: Added in latest commit

AlmeroSteyn:
@jessebeach Thanks!

I have added documentation. Please let me know what you think.

I also added some more tests as the valid `onClick` scenarios were not being explicitly tested.

While I was doing this and digging for the docs two questions came up:
1.  Currently, the rule will block `href` if it contains the word `javascript`. However I was thinking this will also block `href="/docs/javascript/something`. I think this should be allowed of course, but how can this best be checked? Another rule option?
2. The HTML spec for `<a>` calls an anchor without an `href` a placeholder. I can see that some user somewhere might raise this issue one day. But from an accessibility point of view, although it is technically valid, it should be blocked as invalid as most cases for this are from wrong use and I don't see that we can check this. https://www.w3.org/TR/2016/REC-html51-20161101/textlevel-semantics.html#the-a-element

Can I go ahead and remove the `href-no-hash` rule as it has now been fully consumed into this one? Or is there a deprecation path here?

AlmeroSteyn:
Was going through the docs for some other rules and saw that [interactive-supports-focus](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/interactive-supports-focus.md) shows an `<a>` tag with `href="javascript:void(0)"` under the *Succeed* examples. 

Should I change that bit as part of this PR? It technically succeeds under that rule but will just fail under this one.

ljharb:
That's fine; not all rules need be enabled at once, and that this rule prevents something that another rule allows is perfectly fine.

AlmeroSteyn:
Thanks for the response. Clear :-)

On the subject of other rules, I would still like to know what your views are on the `href-no-hash` rule. Can I remove it?

ljharb:
If this new rule covers 100% of the use cases of that one, then sure, you can mark it deprecated (it can't be removed til we're ready for the next semver-major). However, if that's not the case, then both should exist.

AlmeroSteyn:
@ljharb It does cover all the cases, with the exception of cases like `href="javascript:void(0)"` which passed under `href-no-hash` but fails under this rule.

May be a stupid question but how do I deprecate the rule? Set at deprecated in the meta? Or docs alone?

ljharb:
Both the metadata and the docs, I think. I think it's ok to do that in a separate PR tho.

AlmeroSteyn:
@ljharb Cool. Then this PR is ready barring any review change requests.

AlmeroSteyn:
@ljharb I cleverly created this PR from my master branch. Is it ok if I make the deprecation PR once this is merged? Can't be merged before this one anyway.

Unless there is a clever way to do this without messing up this PR by accident haha. Still learning the finer points of GitHub...

ljharb:
Indeed, when a pr is made from master you kind of have to leave master untouched til the pr is merged :-/ hopefully soon tho.

src/index.js
ljharb: To be clear, I'd prefer *not* to add the rule to the recommended configs in this pr, so it can be semver-minor.

That will also give people the chance to test out the rule intentionally before it gets enabled on a wider scale.

AlmeroSteyn: @ljharb I, for once, would be super excited if this gets merged before the next major version release. 

I have removed the rule from the rule from the recommended section. It is now only in strict.

ljharb:
(and pending another rebase)

__tests__/src/rules/anchor-is-valid-test.js
```diff
@@ -0,0 +1,348 @@
...
+const invalidHrefexpectedError = {
+  message: invalidHrefErrorMessage,
+  type: 'JSXOpeningElement',
+};
+
+const components = [{
+  components: ['Anchor', 'Link'],
+}];
+const specialLink = [{
+  specialLink: ['hrefLeft', 'hrefRight'],
```
jessebeach: What are the `hrefLeft` and `hrefRight` options?

AlmeroSteyn:
@ljharb I rebased. 

Try as I may I seem to always pollute the history at this stage. Perhaps you could help me?

We use TFS at work so have to hone my Git skills here ¯\\_(ツ)_/¯

I did the following:
```
git fetch upstream
git rebase upstream/master
git push
```
Would really appreciate some expert advice here.

__tests__/src/rules/anchor-is-valid-test.js
AlmeroSteyn: These options were taken over from `href-no-hash`. These allow for checking custom `href` props that get passed to a wrapper custom component containing `<a>` tags which then consume these props.

In retrospect, the docs should perhaps include an example as `href-no-hash` does.

ljharb:
The first two commands are good, but that last command should fail - you'd want `git push origin master -f` to make it work.

AlmeroSteyn:
Ah yes, of course!!! **smacks forehead** 

Any way to get back to that situation from where I am now?

ljharb:
Yes, just repeat the commands.

AlmeroSteyn:
Thank you! Fixed.

__tests__/src/rules/anchor-is-valid-test.js
AlmeroSteyn: I added the missing example from `href-no-hash` as an explanation for the rule config.

jessebeach:
Let me run this rule against my codebase to evaluate the behavior. Nothing like a set of 35K+ components to test a rule on ;D

AlmeroSteyn:
Whoohoo!! **runs away and watches from a safe distance**

jessebeach:
Ran into an error:

```
value.toLowerCase is not a function
TypeError: value.toLowerCase is not a function
    at ./node_modules/eslint-plugin-jsx-a11y/lib/rules/anchor-is-valid.js:87:57
```

Fixed with:

```
// Hrefs have been found, now check for validity.
      const invalidHrefValues = values
        .filter(value => value !== undefined && value !== null)
        .filter(value =>
          typeof value === 'string' &&
          (
            !value.length
            || value === '#'
            || /^\W*?javascript/.test(value)
          )
        );
```

docs/rules/anchor-is-valid.md
```diff
@@ -0,0 +1,128 @@
...
+## How do I resolve this error?
+
+### Case: The anchor should be a button
+
+The native user agent implementations of the `<a>` and `<button>` elements not only differ in how they look and how they act when activated, but also in how the user is expected to interact with them. Both are perfectly clickable when using a mouse, but keyboard users expect `<a>` to activate on `enter` only and `<button>` to activate on _both_ `enter` and `space`.
+
+This is exacerbated by the expectation sighted users have of how _buttons_ and _anchors_ work based on their appearance. Therefore we find that using _anchors_ as _buttons_ can easily create confusion without a relatively complicated ARIA and CSS implementation that only serves to create an element HTML already offers and browsers already implement fully accessibly.
+
+We are aware that sometimes _anchors_ are used instead of _buttons_ to achieve a specific visual design. When using the `<button>` element this can still be achieved with styling but, due to the meaning many people attach to the standard underlined `<a>` due its appearance, please reconsider this in the design.
+
```
jessebeach: I think we need one more case. I found something like this in my evaluation of my codebase:

```
<a
  className={'thing'}
  onMouseEnter={() => this.setState({showSomething: true})}>
  {label}
</a>
```

What I don't want is for the developer to do something like this:

```
<a
  href="#sdfsdfsdfsdfsdfsdfsdfsdf"
  className={'thing'}
  onMouseEnter={() => this.setState({showSomething: true})}>
  {label}
</a>
```

Which satisfies this rule and the no '#' rule. So we need a case like:

> Case: I need the HTML to be interactive, don't I need to use an `a` tag for that?

With some explanation like:

> An `a` tag is not inherently interactive. Without an `href` attribute, it is really is no different to a `div`. If you need produce a link, make sure this element has an `href` attribute with a value URL or path as the value. If you need to create an interface element that the user can click on, consider using a `button`. If you need to create an interface element that the user can mouse over or mouse out of, consider using a `div` element. In this case, you may need to apply a role of `presentation` or an interactive role. Interactive ARIA roles include `button`, `link`, `checkbox`, `menuitem`, `menuitemcheckbox`, `menuitemradio`, `option`, `radio`, `searchbox`, `switch` and `textbox`.

docs/rules/anchor-is-valid.md
```diff
@@ -0,0 +1,128 @@
...
+In fact, the interactive, underlined `<a>` element has become so synonymous with web navigation that this expectation has become entrenched inside browsers, assistive technologies such as screen readers and in how people generally expect the internet to behave. In short, anchors should navigate.
+
+The use of JavaScript frameworks and libraries, like _React_, has made it very easy to add or subtract functionality from the standard HTML elements. This has led to _anchors_ often being used in applications based on how they look and function instead of what they represent.
+
+Whilst it is possible, for example, to turn the `<a>` element into a fully functional `<button>` element with ARIA, the native user agent implementations of HTML elements are to be preferred over custom ARIA solutions.
+
+## How do I resolve this error?
+
+### Case: The anchor should be a button
```
jessebeach: I like to phrase the Cases as something that a developer might be thinking in their head.

> I need a clickable UI element.

jessebeach: Great documentation below by the way. A+

jessebeach:
I had a 3% error rate for this rule across ~35K files. It picks up a lot of subtle misuses of anchor tags that might not necessary create a bad user experience; but they are certainly bad code smells and indicative of a deeper misunderstanding of HTML. I think this rule can help people understand how to use HTML correctly.

src/rules/anchor-is-valid.js
```diff
@@ -0,0 +1,101 @@
...
+// Rule Definition
+// ----------------------------------------------------------------------------
+
+import { elementType, getProp, getPropValue } from 'jsx-ast-utils';
+import type { JSXOpeningElement } from 'ast-types-flow';
+import { generateObjSchema } from '../util/schemas';
+
+const preferButtonErrorMessage = 'Anchor used as a button. ' +
+  'Anchors are primarily expected to navigate. ' +
+  'Rather use the button element.';
```
AlmeroSteyn: Changed.

src/rules/anchor-is-valid.js
```diff
@@ -0,0 +1,101 @@
...
+
+const noHrefErrorMessage = 'The href attribute is required on an Anchor. ' +
```
AlmeroSteyn: Changed.

src/rules/anchor-is-valid.js
```diff
@@ -0,0 +1,101 @@
...
+  'Provide a valid navigable address as href.';
```
AlmeroSteyn: Changed.

src/rules/anchor-is-valid.js
```diff
@@ -0,0 +1,101 @@
...
+
+const invalidHrefErrorMessage = 'Anchor has an invalid href attribute. ' +
+  'The href should not be empty, only contain "#" or contain the word "javascript". ' +
```
AlmeroSteyn: Removed.

src/rules/anchor-is-valid.js
```diff
@@ -0,0 +1,101 @@
...
+  'Provide a valid navigable address as an href.';
```
AlmeroSteyn: Changed.

docs/rules/anchor-is-valid.md
AlmeroSteyn: Updated. And thank you!!

docs/rules/anchor-is-valid.md
AlmeroSteyn: I have added the case in the docs. Let me know what you think. I used this example and also used it to point out very briefly that full reliance on mouse events should be avoided.

AlmeroSteyn:
@jessebeach I have applied all the changes, including the fix for the `undefined` code issue.

Really great to hear the feedback from the test you did.

src/rules/anchor-is-valid.js
```diff
@@ -0,0 +1,106 @@
...
+import { generateObjSchema } from '../util/schemas';
+
+const preferButtonErrorMessage = 'Anchor used as a button. ' +
+  'Anchors are primarily expected to navigate. ' +
+  'Use the button element instead.';
+
+const noHrefErrorMessage = 'The href attribute is required on an anchor. ' +
+  'Provide a valid, navigable address as the href value.';
+
+const invalidHrefErrorMessage = 'The href attribute is required on an anchor. ' +
```
jessebeach:
```
'The href attribute requires a valid address'
```

Otherwise the error message is the same as `noHrefErrorMessage`.

@AlmeroSteyn What do you think about adding configuration to allow a developer to toggle the 3 errors flavors on and off?  Something like:

```
'jsx-a11y/no-static-element-interactions': [
'error',
  {
    aspects: ['preferButton', 'noHref', 'invalidHref'],
  },
],
```

I'm going to mark this PR as accepted because I'd rather have a conversation about options in a smaller PR and we can make the recommended configuration the same as the behavior of this PR.

AlmeroSteyn: Wow , did not see that happening! Thanks for spotting it. I changed the error message :-)

I think that is a great idea. What I also like about it is that it will allow dev slowly update the code if the rule hits them too hard. 

As I created the PR from my master branch (duhr) I will submit the PR as soon as this one gets merged.

AlmeroSteyn:
@jessebeach Thank you so much for the kind words. It has been nothing but a pleasure to work on this. Would love to keep helping where possible on this amazing rule engine.

src/rules/anchor-is-valid.js
AlmeroSteyn: Just for the record, this was changed in `anchor-is-valid.js` as well.

AlmeroSteyn:
Just did a rebase to latest upstream.

jessebeach:
Merging...

jessebeach:
@AlmeroSteyn What do you think about making the different checks in this rule on/off toggleable through configuation?

AlmeroSteyn:
@jessebeach Definitely. Will get a PR together as soon as possible. Putting together a talk for role=drinks in Amsterdam where I am going to mention this great module. Once that is done, the PR :-)

jessebeach:
@AlmeroSteyn awesome, thank you!!!

PR #231, opened 2017-05-07
Fix: backward compatibility with node.js v4 and v5.

evilebottnawi:
Ref: https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/230


evcohen:
@evilebottnawi [v5.0.1](https://github.com/evcohen/eslint-plugin-jsx-a11y/releases/tag/5.0.1) released! thanks for the fix, really appreciate it

PR #236, opened 2017-05-09
Split interactive supports focus into tabbable and focusable cases

jessebeach:
Updating the `interactive-supports-focus` rule to distinguish between elements that should have a tabindex of 0 (and thus be tabbable) and elements that should have a tabindex of -1 (and thus be focusable).

We don't want to encourage developers to add `tabIndex="0"` willy-nilly to their components. Too many tabindices can create a degraded experience.

The recommended set of roles that should be assigned a tabindex of 0 is given in the recommended settings:

```
'jsx-a11y/interactive-supports-focus': [
  'error',
  {
    tabbable: [
      'button',
      'checkbox',
      'link',
      'searchbox',
      'spinbutton',
      'switch',
      'textbox',
    ],
  },
]
```


jessebeach:
Tests should be passing now. I'd love a review!

evcohen:
@jessebeach got you on a review tmrw!

docs/rules/interactive-supports-focus.md
```diff
@@ -4,7 +4,9 @@ Elements with an interactive role and interaction handlers (mouse or key press)
 
 ## How do I resolve this error?
 
-### Case: This element is a stand-alone control like a button, a link or a form element
+### Case: Elements with the '${role}' interactive role must be tabbable
```
evcohen: `'${role}'` ?

src/rules/interactive-supports-focus.js
```diff
@@ -26,10 +30,45 @@ import getTabIndex from '../util/getTabIndex';
 // Rule Definition
 // ----------------------------------------------------------------------------
 
-const errorMessage =
-  'Elements with interactive roles must be focusable.';
 
-const schema = generateObjSchema();
+const schema = generateObjSchema({
+  tabbable: enumArraySchema([
```
evcohen: can we use `aria-query` for this? i.e. like `...roles.tabbable()`?

src/rules/interactive-supports-focus.js
```diff
@@ -43,10 +82,18 @@ module.exports = {
...
-  create: (context: ESLintContext) => ({
+  create: (context: ESLintContext & {
+    options: {
+      tabbable: Array<string>
+    }
+  }) => ({
     JSXOpeningElement: (
       node: JSXOpeningElement,
     ) => {
+      let tabbable = context.options
```
evcohen: you should be able to make this a const with one line:

```
const tabbable = (context.options && context.options[0] && context.options[0].tabbable) || [];
```

docs/rules/interactive-supports-focus.md
evcohen: oh you're referencing the error message. i think this can be clearer:

```md
#### Case: I got the error "Elements with the '${role}' interactive role must be tabbable." How can I fix this?
```

or something like that

jessebeach: Great suggestion!

src/rules/interactive-supports-focus.js
jessebeach: It's just the interactives. I'll get those from aria-query.

src/rules/interactive-supports-focus.js
jessebeach: Lint complains here when mixing multiple `&&`s and then finishing with `||`. I originally had it like you suggested.

ljharb: It complains even with the wrapping parens?

jessebeach:
@evcohen I've made the requested changes.

src/rules/interactive-supports-focus.js
jessebeach: > It complains even with the wrapping parens?

Indeed, it does not!! Updated.

jessebeach:
Shall we get this in?

jessebeach:
Rebased on current master branch.

jessebeach:
Any objections to merging this?

PR #238, opened 2017-05-10
Update README with explanation recommended/strict

Donaldini:
Added section to readme that explains the difference between recommended and strict rule configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,8 @@ Then configure the rules you want to use under the rules section.
 ```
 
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -89,7 +90,130 @@ You can also enable all the recommended rules at once. Add `plugin:jsx-a11y/reco
 }
 ```
 
+### Difference between 'recommended' and 'strict' mode
```
evcohen: This is great! thank you!! can we move this below the list of supported rules, though?

Donaldini: @evcohen Sure, I was doubting whether to put it under or above myself :)
Fixed it.

PR #240, opened 2017-05-11
Allow label-has-for rule to pass if containing nesting content.

### LINKED ISSUE #51: label-has-for rule should pass if containing nesting content
I am getting hit by the "label-has-for" rule in my code. It appears that every label is required to have an "htmlFor" that matches an ID for an element on the page, but I believe that this should be valid:

```
<label>
  <input />
</label>
```

According to this: https://www.w3.org/TR/html401/interact/forms.html#edef-LABEL 

Referring to the **for** attribute: "When present, the value of this attribute must be the same as the value of the id attribute of some other control in the same document. **When absent, the label being defined is associated with the element's contents.**"

The main hiccup I can think of is if you have multiple elements inside a label.

ignatiusreza:
close #51

### DIFF
--- a/__tests__/src/rules/label-has-for-test.js
+++ b/__tests__/src/rules/label-has-for-test.js
@@ -19,13 +19,21 @@ import rule from '../../../src/rules/label-has-for';
 const ruleTester = new RuleTester();
 
 const expectedError = {
-  message: 'Form controls using a label to identify them must be ' +
-  'programmatically associated with the control using htmlFor',
+  message: 'Form label must have associated control',
   type: 'JSXOpeningElement',
 };
 
 const array = [{
   components: ['Label', 'Descriptor'],
+}];
+const optionsRequiredNesting = [{
+  required: 'nesting',
+}];
+const optionsRequiredSome = [{
+  required: { some: ['nesting', 'id'] },
+}];
+const optionsRequiredEvery = [{
+  required: { every: ['nesting', 'id'] },
 }];
 
 ruleTester.run('label-has-for', rule, {
@@ -54,6 +62,10 @@ ruleTester.run('label-has-for', rule, {
     { code: '<Descriptor htmlFor={`${id}`} />', options: array },
     { code: '<div />', options: array },
     { code: '<Descriptor htmlFor="foo">Test!</Descriptor>', options: array },
+    { code: '<label><input /></label>', options: optionsRequiredNesting },
+    { code: '<label><input /></label>', options: optionsRequiredSome },
+    { code: '<label htmlFor="foo" />', options: optionsRequiredSome },
+    { code: '<label htmlFor="foo"><input /></label>', options: optionsRequiredEvery },
   ].map(parserOptionsMapper),
   invalid: [
     // DEFAULT ELEMENT 'label' TESTS
@@ -62,6 +74,7 @@ ruleTester.run('label-has-for', rule, {
     { code: '<label htmlFor={`${undefined}`} />', errors: [expectedError] },
     { code: '<label>First Name</label>', errors: [expectedError] },
     { code: '<label {...props}>Foo</label>', errors: [expectedError] },
+    { code: '<label><input /></label>', errors: [expectedError] },
 
     // CUSTOM ELEMENT ARRAY OPTION TESTS
     { code: '<Label id="foo" />', errors: [expectedError], options: array },
@@ -102,5 +115,11 @@ ruleTester.run('label-has-for', rule, {
       errors: [expectedError],
       options: array,
     },
+    { code: '<label htmlFor="foo" />', errors: [expectedError], options: optionsRequiredNesting },
+    { code: '<label>First Name</label>', errors: [expectedError], options: optionsRequiredNesting },
+    { code: '<label>First Name</label>', errors: [expectedError], options: optionsRequiredSome },
+    { code: '<label htmlFor="foo" />', errors: [expectedError], options: optionsRequiredEvery },
+    { code: '<label><input /></label>', errors: [expectedError], options: optionsRequiredEvery },
+    { code: '<label>First Name</label>', errors: [expectedError], options: optionsRequiredEvery },
   ].map(parserOptionsMapper),
 });
--- a/docs/rules/label-has-for.md
+++ b/docs/rules/label-has-for.md
@@ -1,6 +1,13 @@
 # label-has-for
 
-Enforce label tags have htmlFor attribute. Form controls using a label to identify them must have only one label that is programmatically associated with the control using: label htmlFor=[ID of control].
+Enforce label tags have associated control.
+
+There are two supported ways to associate a label with a control:
+
+- nesting: by wrapping a control in a label tag
+- id: by using the prop `htmlFor` as in `htmlFor=[ID of control]`
+
+To fully cover 100% of assistive devices, you're encouraged to validate for both nesting and id.
 
 ## Rule details
 
@@ -11,7 +18,10 @@ This rule takes one optional object argument of type object:
     "rules": {
         "jsx-a11y/label-has-for": [ 2, {
             "components": [ "Label" ],
-          }],
+            "required": {
+                "every": [ "nesting", "id" ]
+            }
+        }],
     }
 }
 ```
@@ -42,6 +52,14 @@ return (
   </form>
 );
 ```
+
+The `required` option (defaults to `"required": "id"`) determines which checks are activated. You're allowed to pass in one of the following types:
+
+- string: must be one of the acceptable strings (`"nesting"` or `"id"`)
+- object, must have one of the following properties:
+
+  - some: an array of acceptable strings, will pass if ANY of the requested checks passed
+  - every: an array of acceptable strings, will pass if ALL of the requested checks passed
 
 Note that passing props as spread attribute without `htmlFor` explicitly defined will cause this rule to fail. Explicitly pass down `htmlFor` prop for rule to pass. The prop must have an actual value to pass. Use `Label` component above as a reference. **It is a good thing to explicitly pass props that you expect to be passed for self-documentation.** For example:
 
--- a/src/rules/label-has-for.js
+++ b/src/rules/label-has-for.js
@@ -8,12 +8,45 @@
 // ----------------------------------------------------------------------------
 
 import { getProp, getPropValue, elementType } from 'jsx-ast-utils';
-import { generateObjSchema, arraySchema } from '../util/schemas';
+import { generateObjSchema, arraySchema, enumArraySchema } from '../util/schemas';
 
-const errorMessage = 'Form controls using a label to identify them must be ' +
-  'programmatically associated with the control using htmlFor';
+const errorMessage = 'Form label must have associated control';
 
-const schema = generateObjSchema({ components: arraySchema });
+const enumValues = ['nesting', 'id'];
+const schema = {
+  type: 'object',
+  properties: {
+    components: arraySchema,
+    required: {
+      oneOf: [
+        { type: 'string', enum: enumValues },
+        generateObjSchema({ some: enumArraySchema(enumValues) }, ['some']),
+        generateObjSchema({ every: enumArraySchema(enumValues) }, ['every']),
+      ],
+    },
+  },
+};
+
+const validateNesting = node => !!node.parent.children.find(child => child.type === 'JSXElement');
+const validateId = (node) => {
+  const htmlForAttr = getProp(node.attributes, 'htmlFor');
+  const htmlForValue = getPropValue(htmlForAttr);
+
+  return htmlForAttr !== false && !!htmlForValue;
+};
+const validate = (node, required) => (
+  required === 'nesting' ? validateNesting(node) : validateId(node)
+);
+
+const isValid = (node, required) => {
+  if (Array.isArray(required.some)) {
+    return required.some.some(rule => validate(node, rule));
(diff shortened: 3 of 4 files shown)


__tests__/src/rules/label-has-for-test.js
```diff
@@ -23,10 +23,26 @@ const expectedError = {
...
   type: 'JSXOpeningElement',
 };
+const expectedErrorNesting = {
+  message: 'Form controls using a label to identify them must have the control as one of its children',
+  type: 'JSXOpeningElement',
+};
+const expectedErrorEither = {
+  message: 'Form controls using a label to identify them must be ' +
+  'programmatically associated with the control using htmlFor ' +
+  'or for the label have the control as one of its children',
```
ljharb: this sentence reads awkwardly to me - perhaps `Form controls using a label to identify must be either programmatically associated with the control using htmlFor, or must have the control as one of its children"?

docs/rules/label-has-for.md
```diff
@@ -43,6 +44,8 @@ return (
 );
 ```
 
+While the `required` option, determine which checks are activated. Use this if the default rules `[ "id" ]`, which would check for pair of `id` and `htmlFor` doesn't fit your needs. The alternative check, `nesting`, would pass if the label has nested JSX elements.
```
ljharb: the default should be both - `nesting`, and `id`.

src/rules/label-has-for.js
```diff
@@ -12,9 +12,37 @@ import { generateObjSchema, arraySchema } from '../util/schemas';
 
 const errorMessage = 'Form controls using a label to identify them must be ' +
   'programmatically associated with the control using htmlFor';
+const errorMessageNesting =
+  'Form controls using a label to identify them must have the control as one of its children';
+const errorMessageEither = `${errorMessage} or for the label have the control as one of its children`;
 
 const schema = generateObjSchema({ components: arraySchema });
```
ljharb: I don't see the schema here being modified?

__tests__/src/rules/label-has-for-test.js
```diff
@@ -23,10 +23,26 @@ const expectedError = {
...
+  type: 'JSXOpeningElement',
+};
 
 const array = [{
   components: ['Label', 'Descriptor'],
 }];
+const optionsRequiredNesting = [{
+  required: ['nesting'],
+}];
+const optionsRequiredEither = [{
```
ljharb: This should be "both", not "either".

ignatiusreza: using `both` makes it sounds like we need to provide both `htmlFor` AND `nesting` for the rule to pass.. is that okay?

docs/rules/label-has-for.md
ignatiusreza: that would change existing behavior though.. is that okay? I'm not so sure on how backward compatibility is considered in this project..

__tests__/src/rules/label-has-for-test.js
ljharb: Yes, that's *exactly* what it should require, and what the rule already requires by default.

ljharb: If we want to allow nesting, id, both, and either, then we need to come up with a different schema - but the original plan on #51 was not to have "either".

ignatiusreza: oh, sorry.. I misunderstood the discussion.. will update

__tests__/src/rules/label-has-for-test.js
ignatiusreza: in respect of the rule should requiring both, should the message then become something like `Form controls using a label to identify must be both programmatically associated with the control using htmlFor, and have the control as one of its children`?

src/rules/label-has-for.js
ignatiusreza: sorry, missed this one, will update

docs/rules/label-has-for.md
ljharb: hmm - certainly the default should be whatever the existing behavior is.

ignatiusreza:
updated! thanks for the feedback

ignatiusreza:
btw, wdyt if we use `required: string` instead where valid value for string is `id`, `nesting`, `either`, or `both`? that way, we can cover all cases..

ljharb:
Probably better than that would be `{ required: { id: Boolean, nesting: Boolean } }` (as long as both keys are explicitly required (meaning that providing only "id" or only "nesting" would be an error).

ljharb:
hmm, actually that wouldn't cut it.

I don't like the idea of a string "both" or "either", and i think combining "both" and "either" semantics in the same schema is really problematic.

I agree it makes sense to support those 4 possibilities, but I don't want to box us into a corner where we couldn't add a third mechanism ("aria", say) later.

docs/rules/label-has-for.md
evcohen: we can file a ticket to update the default to both after next breaking change gets merged & before it gets published 😄

ljharb: Sounds good :-)

ignatiusreza:
You're right.. in that case how about having another option, e.g. `minMatch: 1`, which would define how many rule must be satisfied

ljharb:
I'm not sure that's sufficient either, because it'll probably always be important *which* rules are required.

Perhaps a better option would be:
1. just nested: `[['nested']]`
1. just ID: `[['id']]`
1. both: `[['id', 'nested']]` (or `[['nested','id']]`)
1. either: `[['id'], ['nested']]` (or `[['nested'], ['id']]`)

(note: I explicitly required the subarrays to always be arrays, and not to allow `['id']`, to avoid having multiple ways to specify the same thing)

This way, the outer array is like a collection of requirement groups, and a requirement can currently only be "id" or "nested". Thoughts?

ignatiusreza:
It would work, but it feels rather cryptic.. I would rather be more explcit e.g.

```json
required: {
  some: ['id', 'nested']
}
```

for either, or

```json
{
  required: {
    every: ['id', 'nesting']
  }
}
```

for both...

ljharb:
That works for two items - but what about in the future, if we add something like "aria"? How would you support all 3 combinations of "some", of "every", and also ["id", "nesting", "aria and nesting"], let's say?

ignatiusreza:
Something like:

```json
"required": {
  "some": [
    "id",
    "nesting",
    { "every": [ "aria", "nesting" ] }
  ]
}
```

basically, `required: rule`, where `rule` can either be `string`, or object of shape `some: arrayOf(rule)` or `every: arrayOf(rule)`..

not difficult to implement, but I'm not so sure on how to express that in the schema, and what kind of error message should be displayed..

ljharb:
I'm not sure that's any less opaque than what I suggested, but it does seem like a much more complex (and recursive) schema to implement.

ignatiusreza:
`some` and `every` are standard javascript array function..so I thought that it would help in giving clue in what it's doing..

ljharb:
I agree that it does :-) however, the nested case seems problematic, because it'd be complicated to allow 2 levels deep but disallow 3.

ignatiusreza:
Hmmm.. why do we need to disallow? What's the benefit?

ljharb:
If people are allowed to nest them infinitely, someone will make a config with 100 levels deep that crashes eslint :-)

It's always best to restrict things as much as possible.

ignatiusreza:
updated to use `some` & `every`.. related to deep nesting.. I've put in explicit check to disallow nesting deeper than 2 levels.. but, I was not able to write test for this.. because the schema definition actually prohibit deep nesting.. I wonder if relying on the schema is enough, if yes, then we don't need to have explicit check.. wdyt?

ljharb:
Yes, if the schema disallows it, then that should be fine.

docs/rules/label-has-for.md
```diff
@@ -43,6 +53,14 @@ return (
 );
 ```
 
+While the `required` option, determine which checks are activated. You're allowed to pass in one of the following type to `required`:
```
ignatiusreza: `With the required option, determine which checks are activated.` sounds weird... how about `The required option determine which checks are activated.`?

ljharb: `The "required" option determines which checks are activated.` seems good.

ignatiusreza:
documentation updated

docs/rules/label-has-for.md
```diff
@@ -1,6 +1,13 @@
 # label-has-for
 
-Enforce label tags have htmlFor attribute. Form controls using a label to identify them must have only one label that is programmatically associated with the control using: label htmlFor=[ID of control].
+Enforce label tags have associated control.
+
+There're two supported ways to associate label with control:
```
jessebeach: "There are" please 😄 

Exotic contractions can be flummoxing for non-native English speakers. And if you are a non-native English speaker, kudos to your exotic contracting! 

"to associate a label with a control"

docs/rules/label-has-for.md
```diff
@@ -1,6 +1,13 @@
...
+
+- nesting: by wrapping the control as child of the label
```
jessebeach: "by wrapping a control in a label tag"

docs/rules/label-has-for.md
```diff
@@ -1,6 +1,13 @@
...
+- id: by using htmlFor=[ID of control]
```
jessebeach: "by using the prop `htmlFor` as in `htmlFor=[ID of control]`"

__tests__/src/rules/label-has-for-test.js
```diff
@@ -54,6 +70,11 @@ ruleTester.run('label-has-for', rule, {
     { code: '<Descriptor htmlFor={`${id}`} />', options: array },
     { code: '<div />', options: array },
     { code: '<Descriptor htmlFor="foo">Test!</Descriptor>', options: array },
+    { code: '<label><input /></label>', options: optionsRequiredNesting },
+    { code: '<label><input /></label>', options: optionsRequiredSome },
+    { code: '<label htmlFor="foo" />', options: optionsRequiredSome },
+    { code: '<label htmlFor="foo"><input /></label>', options: optionsRequiredEvery },
+    { code: '<label htmlFor="foo"><input /></label>', options: optionsRequiredSomeEvery },
```
jessebeach: What is an example where `optionsRequiredEvery` would provide a different result than `optionsRequiredSomeEvery`? I'm having trouble understanding why the recursively nested options are necessary and having an example that fails because it falls outside the recursion stack size would be illustrative.

ignatiusreza: currently `optionsRequiredSomeEvery` doesn't really do anything interesting.. it's there only to make sure that 2 level nesting works, in preparation of if we want to add a 3rd option for association label to control, e.g. `aria`.. on which, we could do something like:

```json
"required": {
  "some": [
    "id",
    { "every": [ "aria", "nesting" ] }
  ]
}
```

ljharb: I suppose for not, `some: ['id', { 'every': ['id', 'nesting'] }]` would cover that right now, even though that's not particularly useful?

ignatiusreza: yeah, given that now, we only support 2 different checks, it's practically impossible to come up with a real life example..

jessebeach: > it's there only to make sure that 2 level nesting works

Please add a test case that includes two levels of nesting and a test case (fail) that includes 3 levels of testing.

ignatiusreza:
updated.. though I have to say that it doesn't feels good to add a failing test.. :cold_sweat: .. is there a way to test schema using RuleTester? I can't find any.. :crying_cat_face:

ljharb:
I don't think it's possible to write a failing test for the schema. (@jessebeach) I think we just have to forgo that.

ignatiusreza:
hmmm.. I took a bit of look around `aria`, and IIUC, it is meant to be set on the control, instead of the label... which makes me second guess if we actually need to support 2 level nesting on label, since we only have to support `id` and `nesting`..

jessebeach:
> hmmm.. I took a bit of look around aria, and IIUC, it is meant to be set on the control, instead of the label... which makes me second guess if we actually need to support 2 level nesting on label, since we only have to support id and nesting..

Indeed, you wouldn't set `aria-label` et. al. on a `label` tag. Great insight. Hopefully that can allow you to simplify the logic a bit.

ignatiusreza:
cool, updated! :+1:

ignatiusreza:
something like this?

"The `required` option (defaults to `required: 'id'`) determines which checks are activated.`

ljharb:
Sure, that sounds great.

ljharb:
Theoretically this could also be `required: { "some": ["id"] }` or `required: { "every": ["id"] }`, right?

ignatiusreza:
Correct, though `required: "id"` looks nicer.. 😐

ljharb:
Just confirming; let's use the shorter form when describing the default.

jessebeach:
Neat, I'll run this and see what kind of ruckus it raises :) Back in a bit with some results.

ljharb:
I think we should file a separate issue to add this to both configs in the next semver-major.

jessebeach:
@ljharb Shall we get this in?

ljharb:
@jessebeach I think we're still missing the default options being described in docs; but that could go in in a followup

ignatiusreza:
Give me 5 minutes, I'll add in the default options in doc

PR #243, opened 2017-05-15
Fixing #241

### LINKED ISSUE #241: Flow: identifier `ESLintContext`. Could not resolve name
Using Flow 0.42.0 and eslint-plugin-jsx-a11y 5.0.1.

Got the following error when running Flow check on my project:

```
$ flow check 

node_modules/eslint-plugin-jsx-a11y/src/rules/interactive-supports-focus.js:46
 46:   create: (context: ESLintContext) => ({
                         ^^^^^^^^^^^^^ identifier `ESLintContext`. Could not resolve name

node_modules/eslint-plugin-jsx-a11y/src/rules/media-has-caption.js:43
 43:   create: (context: ESLintContext) => ({
                         ^^^^^^^^^^^^^ identifier `ESLintContext`. Could not resolve name

node_modules/eslint-plugin-jsx-a11y/src/rules/no-interactive-element-to-noninteractive-role.js:47
 47:   create: (context: ESLintContext) => {
                         ^^^^^^^^^^^^^ identifier `ESLintContext`. Could not resolve name

node_modules/eslint-plugin-jsx-a11y/src/rules/no-interactive-element-to-noninteractive-role.js:51
 51:         attribute: ESLintJSXAttribute,
                        ^^^^^^^^^^^^^^^^^^ identifier `ESLintJSXAttribute`. Could not resolve name

node_modules/eslint-plugin-jsx-a11y/src/rules/no-noninteractive-element-interactions.js:46
 46:   create: (context: ESLintContext) => {
                         ^^^^^^^^^^^^^ identifier `ESLintContext`. Could not resolve name

node_modules/eslint-plugin-jsx-a11y/src/rules/no-noninteractive-element-to-interactive-role.js:47
 47:   create: (context: ESLintContext) => {
                         ^^^^^^^^^^^^^ identifier `ESLintContext`. Could not resolve name

node_modules/eslint-plugin-jsx-a11y/src/rules/no-noninteractive-element-to-interactive-role.js:51
 51:         attribute: ESLintJSXAttribute,
                        ^^^^^^^^^^^^^^^^^^ identifier `ESLintJSXAttribute`. Could not resolve name

node_modules/eslint-plugin-jsx-a11y/src/rules/no-noninteractive-tabindex.js:45
 45:   create: (context: ESLintContext) => {
                         ^^^^^^^^^^^^^ identifier `ESLintContext`. Could not resolve name

node_modules/eslint-plugin-jsx-a11y/src/rules/no-noninteractive-tabindex.js:49
 49:           node: JSXOpeningElement,
                     ^^^^^^^^^^^^^^^^^ identifier `JSXOpeningElement`. Could not resolve name

node_modules/eslint-plugin-jsx-a11y/src/rules/no-static-element-interactions.js:47
 47:   create: (context: ESLintContext) => {
                         ^^^^^^^^^^^^^ identifier `ESLintContext`. Could not resolve name


Found 10 errors
error Command failed with exit code 2.
```

Looks like `src/` folder should be excluded from the module.

leos:
Making the flow type imports explicit. This also revealed a potential issue in no-noninteractive-element-to-interactive-role.js.

Fixes #241.

### DIFF
--- a/flow/eslint.js
+++ b/flow/eslint.js
@@ -1,9 +1,12 @@
-type ESLintReport = {
+/*
+ * @flow
+ */
+export type ESLintReport = {
   node: any,
   message: string,
 };
 
-type ESLintContext = {
+export type ESLintContext = {
   options: Array<Object>,
   report: (ESLintReport) => void,
 };
--- a/src/rules/no-noninteractive-element-to-interactive-role.js
+++ b/src/rules/no-noninteractive-element-to-interactive-role.js
@@ -16,6 +16,7 @@ import {
   elementType,
   getProp,
   getLiteralPropValue,
+  propName,
 } from 'jsx-ast-utils';
 import type {
   JSXIdentifier,
@@ -52,7 +53,7 @@ module.exports = {
       JSXAttribute: (
         attribute: ESLintJSXAttribute,
       ) => {
-        const attributeName: JSXIdentifier = attribute.name.name;
+        const attributeName: JSXIdentifier = propName(attribute);
         if (attributeName !== 'role') {
           return;
         }
--- a/src/rules/no-noninteractive-element-to-interactive-role.js
+++ b/src/rules/no-noninteractive-element-to-interactive-role.js
@@ -21,6 +21,8 @@ import type {
   JSXIdentifier,
 } from 'ast-types-flow';
 import includes from 'array-includes';
+import type { ESLintContext } from '../../flow/eslint';
+import type { ESLintJSXAttribute } from '../../flow/eslint-jsx';
 import isNonInteractiveElement from '../util/isNonInteractiveElement';
 import isInteractiveRole from '../util/isInteractiveRole';
 
--- a/.flowconfig
+++ b/.flowconfig
@@ -2,6 +2,3 @@
 <PROJECT_ROOT>/lib/.*
 <PROJECT_ROOT>/docs/.*
 <PROJECT_ROOT>/reports/.*
-
-[libs]
-flow
--- a/__tests__/src/rules/no-interactive-element-to-noninteractive-role-test.js
+++ b/__tests__/src/rules/no-interactive-element-to-noninteractive-role-test.js
@@ -288,6 +288,9 @@ const alwaysValid = [
   { code: '<div role="term" />;' },
   { code: '<div role="timer" />;' },
   { code: '<div role="tooltip" />;' },
+  /* Namespaced roles are not checked */
+  { code: '<div mynamespace:role="term" />' },
+  { code: '<input mynamespace:role="img" />' },
 ];
 
 const neverValid = [
--- a/flow/eslint-jsx.js
+++ b/flow/eslint-jsx.js
@@ -1,8 +1,11 @@
+/*
+ * @flow
+ */
 import type {
   JSXAttribute,
   JSXOpeningElement,
  } from 'ast-types-flow';
 
-type ESLintJSXAttribute = {
+export type ESLintJSXAttribute = {
   parent: JSXOpeningElement
 } & JSXAttribute;
--- a/src/rules/interactive-supports-focus.js
+++ b/src/rules/interactive-supports-focus.js
@@ -13,6 +13,7 @@ import {
 } from 'jsx-ast-utils';
 import type { JSXOpeningElement } from 'ast-types-flow';
 import includes from 'array-includes';
+import type { ESLintContext } from '../../flow/eslint';
 import { generateObjSchema } from '../util/schemas';
 import isHiddenFromScreenReader from '../util/isHiddenFromScreenReader';
 import isInteractiveElement from '../util/isInteractiveElement';
--- a/src/rules/media-has-caption.js
+++ b/src/rules/media-has-caption.js
@@ -10,6 +10,7 @@
 
 import type { JSXElement, JSXOpeningElement } from 'ast-types-flow';
 import { elementType, getProp, getLiteralPropValue } from 'jsx-ast-utils';
+import type { ESLintContext } from '../../flow/eslint';
 import { generateObjSchema, arraySchema } from '../util/schemas';
 
 const errorMessage = 'Media elements such as <audio> and <video> must have a <track> for captions.';
--- a/src/rules/no-interactive-element-to-noninteractive-role.js
+++ b/src/rules/no-interactive-element-to-noninteractive-role.js
@@ -20,6 +20,8 @@ import {
 } from 'jsx-ast-utils';
 import type { JSXIdentifier } from 'ast-types-flow';
 import includes from 'array-includes';
+import type { ESLintContext } from '../../flow/eslint';
+import type { ESLintJSXAttribute } from '../../flow/eslint-jsx';
 import isInteractiveElement from '../util/isInteractiveElement';
 import isNonInteractiveRole from '../util/isNonInteractiveRole';
 import isPresentationRole from '../util/isPresentationRole';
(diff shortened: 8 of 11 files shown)


ljharb:
Seems legit to me, but I'll defer to reviewers who are more familiar with Flow.

src/rules/no-noninteractive-element-to-interactive-role.js
```diff
@@ -50,7 +53,7 @@ module.exports = {
       JSXAttribute: (
         attribute: ESLintJSXAttribute,
       ) => {
-        const attributeName: JSXIdentifier = attribute.name.name;
+        const attributeName: JSXIdentifier = propName(attribute);
```
ljharb: it would be good to add a regression test on this tho, if this makes a difference.

flow/eslint.js
```diff
@@ -1,9 +1,12 @@
...
+/*
+ * @flow
+ */
+export type ESLintReport = {
   node: any,
   message: string,
 };
 
-type ESLintContext = {
+export type ESLintContext = {
```
evcohen: isn't there a way to make types global for your project?

leos: I'm not an expert on Flow, but it can be done is a [libdef](https://flow.org/en/docs/libdefs/creation/#toc-declaring-a-global-type). The way it is currently being done is problematic for two reasons:

1) It requires an entry in .flowconfig which doesn't work correctly when the project is sitting in  node_modules (see #241)
2) Again, not sure what the underlying reason was but it looks like doing it that way masked a bug - `attribute.name.name`.

There's a bigger decision which you guys as project owners should make is whether you want to make these types of imports explicit. Coming from Python, IMO explicit is better than implicit.

ljharb: fwiw, I'm a huge fan of explicit over implicit, and I've never written Python. We should be trying to make fewer globals, not more.

src/rules/no-noninteractive-element-to-interactive-role.js
jessebeach: Doesn't make a difference and in fact I should have used the `propName` function originally to deal with namespaces.

flow/eslint.js
jessebeach: Well that's just really weird. I reproduced the typing issue raised in `no-noninteractive-element-to-interactive-role` when the eslint types are explicitly exported. Nice catch @leos .

I agree with @ljharb here on explicit vs. implicit.

src/rules/no-noninteractive-element-to-interactive-role.js
ljharb: If it doesn't make a difference tho, why does the `propName` function need to exist?

If `propName` handles some edge case, then it'd be good to have a regression test that fails without usage of `propName`.

flow/eslint.js
evcohen: cool! agreed, just wondering 😄

src/rules/no-noninteractive-element-to-interactive-role.js
evcohen: it does handle multiple cases- an attribute can have a namespace:

https://github.com/evcohen/jsx-ast-utils/blob/master/src/propName.js#L9

ljharb: Great - let's add a regression test with a namespaced attribute then :-)

leos:
Sorry, just to clarify, do you want me to make the test update as part of the PR?

evcohen:
> Sorry, just to clarify, do you want me to make the test update as part of the PR?

@leos yes please, since we are updating to `propNames` in this PR, we should be adding the test in same PR. thanks for all of your work with this!

leos:
Ok, can one of you either point me to a similar test that I can replicate or lay out what the logic I'm testing for is? I copied the `propName()` usage from `no-interactive-element-to-noninteractive-role.js` just to clear the Flow error. Briefly glancing though the tests I could find on that tests namespaces (and there are several other uses of `propName()`).

I don't grok what this module does or what interactive/noninteractive roles are - should `<input type="img" mycustomnamespace:role="img" />` be valid? If so, the code in the PR won't act on it since `propName` will return `mycustomnamespace:role`. I assume the original code would have had the same behavior.

Might be my unfamiliarity with the test framework but trying to add something like this to `alwaysValid` in the tests :
```
  /* Namespaced roles are not checked */
  { code: '<div mynamespace:role="term">' },
  { code: '<input mynamespace:role="img">' },
```
results in a `Parsing error: Unexpected token`.

ljharb:
You seem to be missing a closing slash; not sure if that's the problem.

leos:
Thanks @ljharb ! That fixed it, is the intent/logic correct?

evcohen:
@leos this LGTM - @jessebeach @ljharb does this require a patch release?

evcohen:
[v5.0.2](https://github.com/evcohen/eslint-plugin-jsx-a11y/releases/tag/5.0.2) published with fix. Thanks for all of your help @leos & everybody!!! let me know if its resolved in `v5.0.2`

leos:
Ugh, I'm sorry, when it's packaged up apparently the 'flow' directory doesn't come along for the ride so the imports are broken.
```
node_modules/eslint-plugin-jsx-a11y/src/rules/interactive-supports-focus.js:16
 16: import type { ESLintContext } from '../../flow/eslint';
                                        ^^^^^^^^^^^^^^^^^^^ ../../flow/eslint. Required module not found

node_modules/eslint-plugin-jsx-a11y/src/rules/media-has-caption.js:13
 13: import type { ESLintContext } from '../../flow/eslint';
                                        ^^^^^^^^^^^^^^^^^^^ ../../flow/eslint. Required module not found

node_modules/eslint-plugin-jsx-a11y/src/rules/no-interactive-element-to-noninteractive-role.js:23
 23: import type { ESLintContext } from '../../flow/eslint';
                                        ^^^^^^^^^^^^^^^^^^^ ../../flow/eslint. Required module not found

node_modules/eslint-plugin-jsx-a11y/src/rules/no-interactive-element-to-noninteractive-role.js:24
 24: import type { ESLintJSXAttribute } from '../../flow/eslint-jsx';
                                             ^^^^^^^^^^^^^^^^^^^^^^^ ../../flow/eslint-jsx. Required module not found

node_modules/eslint-plugin-jsx-a11y/src/rules/no-noninteractive-element-interactions.js:22
 22: import type { ESLintContext } from '../../flow/eslint';
                                        ^^^^^^^^^^^^^^^^^^^ ../../flow/eslint. Required module not found

node_modules/eslint-plugin-jsx-a11y/src/rules/no-noninteractive-element-to-interactive-role.js:25
 25: import type { ESLintContext } from '../../flow/eslint';
                                        ^^^^^^^^^^^^^^^^^^^ ../../flow/eslint. Required module not found

node_modules/eslint-plugin-jsx-a11y/src/rules/no-noninteractive-element-to-interactive-role.js:26
 26: import type { ESLintJSXAttribute } from '../../flow/eslint-jsx';
                                             ^^^^^^^^^^^^^^^^^^^^^^^ ../../flow/eslint-jsx. Required module not found

node_modules/eslint-plugin-jsx-a11y/src/rules/no-noninteractive-tabindex.js:23
 23: import type { ESLintContext } from '../../flow/eslint';
                                        ^^^^^^^^^^^^^^^^^^^ ../../flow/eslint. Required module not found

node_modules/eslint-plugin-jsx-a11y/src/rules/no-static-element-interactions.js:23
 23: import type { ESLintContext } from '../../flow/eslint';
                                        ^^^^^^^^^^^^^^^^^^^ ../../flow/eslint. Required module not found
```

I think the best approach is to move flow into `src` with everything else.

evcohen:
I think we can just remove `flow` from `.npmignore`

leos:
That would work too. Do you want me to do the PR?

PR #251, opened 2017-05-31
Add sub rule configuration to anchor-is-valid.

AlmeroSteyn:
This adds configuration to the `anchor-is-valid` rule.

It will allow the user to toggle which of the three sub rules are run with the following configuration:

```
'jsx-a11y/no-static-element-interactions': [
'error',
  {
    aspects: ['preferButton', 'noHref', 'invalidHref'],
  },
],
```

`preferButton`: This will report on when the anchor is used as a button.
`noHref`: This will report when no href is present on an anchor.
`invalidHref`: This will report when the href is filled incorrectly.

This way the user can more fully control the rule in restrictive environments.

The PR does not contain tests and documentation yet to first discuss the actual rule change (and to get the ball rolling :-) ).

@jessebeach What do you think? Do you think this fully covers what you had in mind?

### DIFF
--- a/__tests__/src/rules/anchor-is-valid-test.js
+++ b/__tests__/src/rules/anchor-is-valid-test.js
@@ -47,9 +47,40 @@ const components = [{
 const specialLink = [{
   specialLink: ['hrefLeft', 'hrefRight'],
 }];
+const noHrefAspect = [{
+  aspects: ['noHref'],
+}];
+const invalidHrefAspect = [{
+  aspects: ['invalidHref'],
+}];
+const preferButtonAspect = [{
+  aspects: ['preferButton'],
+}];
+const noHrefInvalidHrefAspect = [{
+  aspects: ['noHref', 'invalidHref'],
+}];
+const noHrefPreferButtonAspect = [{
+  aspects: ['noHref', 'preferButton'],
+}];
+const preferButtonInvalidHrefAspect = [{
+  aspects: ['preferButton', 'invalidHref'],
+}];
+
 const componentsAndSpecialLink = [{
   components: ['Anchor'],
   specialLink: ['hrefLeft'],
+}];
+
+const componentsAndSpecialLinkAndInvalidHrefAspect = [{
+  components: ['Anchor'],
+  specialLink: ['hrefLeft'],
+  aspects: ['invalidHref'],
+}];
+
+const componentsAndSpecialLinkAndNoHrefAspect = [{
+  components: ['Anchor'],
+  specialLink: ['hrefLeft'],
+  aspects: ['noHref'],
 }];
 
 ruleTester.run('anchor-is-valid', rule, {
@@ -59,9 +90,8 @@ ruleTester.run('anchor-is-valid', rule, {
     { code: '<a href="foo" />' },
     { code: '<a href={foo} />' },
     { code: '<a href="/foo" />' },
+    { code: '<a href="https://foo.bar.com" />' },
     { code: '<div href="foo" />' },
-    { code: '<a href={`${undefined}foo`}/>' },
-    { code: '<a href={`#${undefined}foo`}/>' },
     { code: '<a href={`#foo`}/>' },
     { code: '<a href={"foo"}/>' },
     { code: '<a href="#foo" />' },
@@ -73,9 +103,8 @@ ruleTester.run('anchor-is-valid', rule, {
     { code: '<Anchor href="foo" />', options: components },
     { code: '<Anchor href={foo} />', options: components },
     { code: '<Anchor href="/foo" />', options: components },
+    { code: '<Anchor href="https://foo.bar.com" />', options: components },
     { code: '<div href="foo" />', options: components },
-    { code: '<Anchor href={`${undefined}foo`}/>', options: components },
-    { code: '<Anchor href={`#${undefined}foo`}/>', options: components },
     { code: '<Anchor href={`#foo`}/>', options: components },
     { code: '<Anchor href={"foo"}/>', options: components },
     { code: '<Anchor href="#foo" />', options: components },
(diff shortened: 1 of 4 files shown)


src/rules/anchor-is-valid.js
```diff
@@ -23,7 +23,17 @@ const noHrefErrorMessage = 'The href attribute is required on an anchor. ' +
 const invalidHrefErrorMessage = 'The href attribute requires a valid address. ' +
   'Provide a valid, navigable address as the href value.';
 
-const schema = generateObjSchema();
+const noAspectsErrorMessage = 'The aspects array in the rule configuration should not be empty. ' +
```
ljharb: this message shouldn't be needed; eslint's own json schema enforcement should be able to enforce that there's at least 1 item in the array.

src/rules/anchor-is-valid.js
```diff
@@ -23,7 +23,17 @@ const noHrefErrorMessage = 'The href attribute is required on an anchor. ' +
...
+  'Remove it or choose specific aspects to run. By default all aspects are included.';
+
+const invalidAspectErrorMessage = aspect => `Aspect "${aspect}" in the rule configuration is invalid. ` +
```
ljharb: Similarly, the eslint schema alone should be validating that only the provided enum types are valid.

src/rules/anchor-is-valid.js
```diff
@@ -44,6 +54,34 @@ module.exports = {
...
+        context.report({
+          node,
+          message: noAspectsErrorMessage,
+        });
+        return;
+      }
+
+      // Create active aspect flag object. Failing checks will only report
+      // if the related flag is set to true.
+      const activeAspects = Object.assign({},
```
ljharb: can we use object spread syntax here instead of Object.assign?

AlmeroSteyn:
@ljharb I have made the requested changes:

Enforce no empty aspects array:
For the empty enum array check, I had to change `enumArraySchema` to accept the minimum item amount. However, the error message is a bit cryptic when you have an empty array: `Value "" has less items than allowed`.  What is your feeling on that? And are you OK with the change to the schema file?

Dunno if there is another way to do that with the schema. Another option would be to run all sub-rules if the array is empty. Just feels a bit weird to override the config like that.

Enforce enum values:
Now being done with the schema.

Object.assign to Object spread:
I did that, but in the end I went to a `forEach` which ended up being a lot more simple and readable. 

I will add the docs and tests ASAP.

AlmeroSteyn:
@ljharb Sorry I see that array.includes also is not supported here yet. Will fix ASAP to indexOf.

AlmeroSteyn:
The `array.include` has been replaced with `indexOf` :-)

ljharb:
* docs, and tests :-)

AlmeroSteyn:
I have added the tests and doc changes.

The tests do not cover every possible permutation of configuration and scenario but I believe the test pack is solid enough as a whole.

During testing I found an edge case that was not working correctly. An anchor with no `href` and an `onClick` did not report when the `noHref` option was on but the `preferButton` option was off.  The following code change fixed it:

From:
```
 if (!hasSpreadOperator && !onClick && activeAspects.noHref) {
          context.report({
            node,
            message: noHrefErrorMessage,
          });
        }
```
To:
```
if (!hasSpreadOperator && activeAspects.noHref &&
          (!onClick || (onClick && !activeAspects.preferButton))) {
          context.report({
            node,
            message: noHrefErrorMessage,
          });
        
```

The PR should be complete now, barring any review comments :-)

__tests__/src/rules/anchor-is-valid-test.js
```diff
@@ -198,6 +229,60 @@ ruleTester.run('anchor-is-valid', rule, {
...
+    { code: '<a />', options: preferButtonInvalidHrefAspect },
+    { code: '<a href={undefined} />', options: preferButtonInvalidHrefAspect },
+    { code: '<a href={null} />', options: preferButtonInvalidHrefAspect },
+
+    // INVALID HREF
+    { code: '<a href="" />;', options: preferButtonAspect },
+    { code: '<a href="#" />', options: preferButtonAspect },
+    { code: '<a href={"#"} />', options: preferButtonAspect },
+    { code: '<a href={`#${undefined}`} />', options: preferButtonAspect },
+    { code: '<a href={`${undefined}`} />', options: preferButtonAspect },
```
ljharb: this will toString to `"undefined"` - why is `href="undefined"` invalid? That could certainly be a valid relative URL.

AlmeroSteyn: OK that's weird. When in test mode the value is read off the `href` as empty strring in this case. But in an actual component, it renders to 'undefined'. Will investigate further in the morning and try and see why.

jessebeach: It's a pretty minor case. Just removing these instances is fine as well. I don't think they're really testing anything.

ljharb: It's the same thing as `href="undefined"` - i agree that that isn't special, but i think it's worth testing an anchor href with normal content behind it - ie, not just "#" - as well as a normal href (with a URL)

AlmeroSteyn:
Had a look at the `href="${undefined}"` thing. As mentioned it render to `href="undefined"` in an actual React component. However `getPropValue` from `jsx-ast-utils` returns an empty string for the `${undefined}`. So `${undfined}foo` comes back as `foo`.

This pretty much renders all those tests unpredictable and useless. So I have removed all tests with that in the `href`. 

@ljharb  I have also added some tests with a full URL.

Now I think the PR is finished, hehe.

ljharb:
@AlmeroSteyn can you file a separate issue then about `getPropValue` returning an empty string for `undefined`? that doesn't seem correct.

AlmeroSteyn:
@ljharb I found the following test in `jsx-ast-utils` lib:

```
    it('should drop variables in template literals that are literally undefined', () => {
      const prop = extractProp('<div foo={`bar ${undefined}`} />');

      const expected = 'bar ';
      const actual = getPropValue(prop);

      assert.equal(expected, actual);
    });
```

The results experienced here is specifically being tested for there. Do you think it is still an issue as such? This is for actual undefined variables. I cannot come up with a use case where someone would want to navigate to `undefined` if his code yields an undefined variable that cannot be solved more elegantly?

Although having this function differ from the actual behaviour in JSX could be a point for discussion.

Does this impact this PR further? Still happy to open the issue on that side so we can discuss it there further?

ljharb:
I'm pretty confident that test is in error.

AlmeroSteyn:
Issue logged here: https://github.com/evcohen/jsx-ast-utils/issues/40

jessebeach:
Cool, let's merge this. I'll be going over what's in master tomorrow to determine what kind of release we need and I'll compile the release notes.

PR #255, opened 2017-06-04
Adding NPM5 package-lock.json file

jessebeach:
Adding the NPM@5 package-lock.json file.

I ran `rm -rf node_modules && npm i`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,4565 @@
+{
+  "name": "eslint-plugin-jsx-a11y",
+  "version": "5.0.3",
+  "lockfileVersion": 1,
+  "dependencies": {
+    "abab": {
+      "version": "1.0.3",
+      "resolved": "https://registry.npmjs.org/abab/-/abab-1.0.3.tgz",
+      "integrity": "sha1-uB3l9ydOxOdW15fNg08wNkJyTl0=",
+      "dev": true
+    },
+    "acorn": {
+      "version": "5.0.3",
+      "resolved": "https://registry.npmjs.org/acorn/-/acorn-5.0.3.tgz",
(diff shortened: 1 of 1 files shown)


jessebeach:
> If we do want this file, we should ditch yarn, because it makes no sense to have two competing lockfiles in the same repo (and I do think that npm5's lockfile is a strict improvement over yarn, ftr)

Good points. I have zero opinion here. I'd rather stick with NPM if it's all the same. @evcohen ?

evcohen:
does older versions of npm respect the lock file? that would be my only reason not to

jessebeach:
My understanding is that older versions of NPM don't need the lock file.

ljharb:
Older versions of npm don't respect it; they need it just the same (where "need" means, "need it in order to create a deterministic node_modules).

If you want a lockfile despite being a non-app, you want developers but not consumers to be impacted by this lockfile, and you want to support using the lockfile on npm versions older than 5, yarn is the only choice.

evcohen:
i'm missing the point of it then. this doesn't solve the "works on my machine" problem for all users using npm < 5. this requires all of us to upgrade our npm version or else its rendered useless. 

i think i am in favor of removing all lockfiles. I've only seen them get in the way of development and rarely have I seen a downstream dependency break this project (especially given its lack of dependencies, three of which one of us owns)

ljharb:
I think "devs must use npm 5+" is the same caliber of requirement as "devs must use yarn". I would definitely prefer to see us [remove all lockfiles](https://github.com/airbnb/javascript/commit/5ba31498feaec1c977008b62c2768f4ea5d498df).

jessebeach:
Cool, I'll update the diff and remove the lock files.

evcohen:
Fixed in #261 per discussion above!

PR #256, opened 2017-06-04
Update aria-query to 0.7.0

jessebeach:


jessebeach:
Depends on #255

evcohen:
Finished in #260 - going to send separate PR to remove all lock files.

Issue #250, opened 2017-05-31
Make default for anchor-is-valid both id and nesting


evcohen:
To avoid breaking changes, we left it to default to just check `htmlFor`, but ideally it would default to both `htmlFor` and nesting. When the first necessary breaking change comes in, we should go back and update this.

AlmeroSteyn:
@evcohen Should this be for `label-has-for` instead?

evcohen:
yep- my bad was getting caught up on all of the changes and misread haha. thanks @AlmeroSteyn

jessebeach:
That time is now! Let's get this in for a v6 release.

jessebeach:
@AlmeroSteyn whomever of us gets to it first :)

AlmeroSteyn:
@jessebeach Going to try and be the first, but my agenda for the coming two weeks may have other opinions about that hehe.

But, and this is probably a discussion that has already happened elsewhere, are we happy with the `label-has-for` rule? 

Currently, it only checks if a label has a filled `htmlFor` prop or if a label contains any JSX element as content.

So the following will pass (I just added `form` to illustrate that the dangling labels pass by themselves):
```
<form>
   <label htmlFor='id-does-nog-exist' />
</form>
```
and
```
<form>
   <label>
       <div />
   </label>
</form>
```

It certainly increases the chance that labels are associated with form controls, however, it does not check whether form controls have an accessible name. Also, it does not check if the label contains any text at all.

What I do find worrying is that the following fails:
```
<label>
   {children>
</label>
```

That would certainly be a way I would want to use the nesting functionality combined with React components.

Now, with props spreading, componentization and such I can imagine that this could be quite a feat to check properly without creating blocking scenarios. But I was wondering if this is something you would like to invest time in?

To switch on the `nesting` by default is no huge issue but, with it not catering for `children`, I don't think we can simply do that.

Opinions?

jessebeach:
> Now, with props spreading, componentization and such I can imagine that all these checks could be quite difficult to check properly without creating blocking scenarios.

I'm pretty convinced that we have a hard-ceiling with linting and I'm OK with at. It's not meant to be wholistic solution; it's a chapter in a quality story. There's only so much we can provide a go/no-go evaluation for at this level. My team's next project is to provide quality signal at the composition level. 

```
<label>
   {children>
</label>
```

You're right that this shouldn't be failing. Logging a bug.

AlmeroSteyn:
[...]

PR #257, opened 2017-06-04
Bump, major version

jessebeach:
Bump to v6.0.0

Depends on #255, #256


evcohen:
what is breaking in this release/can we mark it in the changelog as breaking? if it's breaking, then let's include #250

jessebeach:
It seemed break-ish, given the update to the configs. But let's put in #250 and make it break-ing.

evcohen:
haha sounds like a plan 👍🏼

AlmeroSteyn:
Currently, the `anchor-is-valid` rule is only included in the strict config. Should it, as it is breaking, be included in the recommended config, albeit with not all aspects activated? 

It is also overlapping the `href-no-hash` rule. The `invalidHref` aspect is incorporating the same check.

jessebeach:
> It is also overlapping the href-no-hash rule. The invalidHref aspect is incorporating the same check.

Anyone opposed to deleting `href-no-hash` in favor of `anchor-is-valid`.

ljharb:
As long as the test cases for href-no-hash are all indeed covered by anchor-is-valid, sgtm

jessebeach:
> As long as the test cases for href-no-hash are all indeed covered by anchor-is-valid, sgtm

I'll verify.

CHANGELOG.md
```diff
@@ -1,3 +1,14 @@
+6.0.0 / 2017-06-05
+=================
+- [new] Add rule `anchor-is-valid`. See documentation for configuration options. Thanks @AlmeroSteyn.
+- [breaking] `href-no-hash` replaced with `anchor-is-valid` in the recommended and strict configs. `href-no-hash` will be deprecated in a future release.
```
ljharb: This text may need updating

jessebeach: Indeed. I had unstaged changes for this text. Changes are forthcoming. I've got some test fails from running the `href-no-hash` tests in `anchor-is-valid`. Just trying to work through them first.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "eslint-plugin-jsx-a11y",
-  "version": "5.0.3",
+  "version": "6.0.0",
```
evcohen: Can we also remove support for `eslint` v2 per #268

AlmeroSteyn:
Are there any specific issues left you would like to get into this release? Would be happy to help with something.

jessebeach:
@evcohen ready to merge!

fearphage:
This caused facebookincubator/create-react-app#2631. There's nothing to be done here. The other issue is just locked so I can't comment there, but they need to be associated with each other.

ljharb:
It didn’t “cause” it; its just that peer deps were incompatible.

PR #259, opened 2017-06-07
update label-has-for children content handling

### LINKED ISSUE #258: label-has-for should not fail on children content
As mentioned in #250, the following pattern fails in `label-has-for`.

```
<label>
   {children}
</label>
```

This pattern should be allowed. We cannot know what `{children}` will render, so we can't raise a failure with certainty.

mjaltamirano:
This should fix #258.

It updates the label-has-for nesting validation to check for a `JSXExpressionContainer` in addition to a `JSXElement` (which is the default behavior).

In addition it adds tests cases testing `<label>{children}</label>`. Currently `<label>{children}<label>` passes for the options `required some` and `required nesting`, but fails w/ no options and the options  `array` and `required every`

Please let me know if there is any follow-up work required for this PR! I wasn't sure if the fix was fine as-is or require extra configuration/documentation based on the comments in #258.


__tests__/src/rules/label-has-for-test.js
```diff
@@ -66,6 +66,8 @@ ruleTester.run('label-has-for', rule, {
     { code: '<label><input /></label>', options: optionsRequiredSome },
     { code: '<label htmlFor="foo" />', options: optionsRequiredSome },
     { code: '<label htmlFor="foo"><input /></label>', options: optionsRequiredEvery },
+    { code: '<label>{children}</label>', options: optionsRequiredSome },
```
ljharb: these should be invalid by default - we should be adding a new option to make these valid, for those that wish to enable it (see https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/258#issuecomment-306637159)

mjaltamirano: Noted! Thanks for the quick review. Will get on that.

src/rules/label-has-for.js
```diff
@@ -19,24 +19,37 @@ const schema = {
     components: arraySchema,
     required: {
       oneOf: [
-        { type: 'string', enum: enumValues },
+        { type: 'string', enum: enumValues.concat('children') },
```
ljharb: hmm, i'm not sure this is the proper way to add this option.

If it's in the same place "nesting" and "id" goes, then it should be an alternative method to ensure a label has "for". In this case, this is a special separate option that bypasses the other options, no?

mjaltamirano: I wrote up a response but I'm going to approach this a different way first and see if I can address your concerns. (I had concerns about the hackiness of the approach, too.)

Thank you for the prompt review!

mjaltamirano:
Let’s see if third time is a charm!

I’ve added a new object schema for validation, `children`, that requires opting-into an explicit `children` required option in order to be valid. I’ve also incorporated the `hasAccessibleChild` util method in order to do this. (Thanks again for pointing it out @evcohen !)

If all looks well I am happy to work on updating the documentation next. Let me know!

__tests__/src/rules/label-has-for-test.js
```diff
@@ -35,6 +35,9 @@ const optionsRequiredSome = [{
 const optionsRequiredEvery = [{
   required: { every: ['nesting', 'id'] },
 }];
+const optionsRequiredChildren = [{
+  required: { children: ['children'] },
```
ljharb: hm - i wonder if "required" is the right word to use here. Maybe just a boolean, `allowChildren`?

src/rules/label-has-for.js
```diff
@@ -22,27 +23,39 @@ const schema = {
...
         generateObjSchema({ some: enumArraySchema(enumValues) }, ['some']),
         generateObjSchema({ every: enumArraySchema(enumValues) }, ['every']),
+        generateObjSchema({ children: enumArraySchema(['children']) }, ['children']),
       ],
     },
   },
 };
 
-const validateNesting = node => !!node.parent.children.find(child => child.type === 'JSXElement');
+const validateNesting = node => (!!node.parent.children.find(child => child.type === 'JSXElement'));
```
ljharb: instead of `!!arr.find(…)`, do `arr.some(…)` (i realize this was already in master)

mjaltamirano: Updated to arr.some in my most recent commit (also did not mean to change this the last time around by adding parentheses--whoops!)

__tests__/src/rules/label-has-for-test.js
mjaltamirano: Noted! I was working with the existing infrastructure, but in my most recent commit added a third property, allowChildren, which accepts a boolean type, so in the test file this just reads allowChildren: true. 

In the source file the most significant changes were to the `const required` assignment in create at the bottom of the file and in the validate case statement. I added a comment to try and clarify the nature of the allowChildren property, but I'm happy to work it in in a more congruous manner if there are better options.

__tests__/src/rules/label-has-for-test.js
```diff
@@ -35,6 +35,9 @@ const optionsRequiredSome = [{
 const optionsRequiredEvery = [{
   required: { every: ['nesting', 'id'] },
 }];
+const optionsRequiredChildren = [{
```
ljharb: not a huge deal, but let's rename this var to `optionsChildrenAllowed` or something :-)

src/rules/label-has-for.js
```diff
@@ -24,19 +25,30 @@ const schema = {
...
-const validate = (node, required) => (
-  required === 'nesting' ? validateNesting(node) : validateId(node)
-);
+
+const validate = (node, required) => {
+  switch (required) {
+    case 'nesting':
+      return validateNesting(node);
+    // Case covers optional "allowChildren" property if set true
+    case true:
```
ljharb: this in particular makes me think a switch is really not appropriate for this function. What about:
```js
if (required === 'nesting') {
  return validateNesting(node);
}
if (required === true) {
  return hasAccessibleChild(node.parent);
}
return validateId(node);
```

Also, it seems pretty weird that "required" can be either a string, or a boolean - can we maybe make it always be a string, or maybe make `allowChildren` a separate boolean from `required`, which only deals with "nesting" and/or "id"?

__tests__/src/rules/label-has-for-test.js
mjaltamirano: Definitely. The naming inconsistency was an oversight.

src/rules/label-has-for.js
mjaltamirano: How about passing in a third argument to isValid, `allowChildren`, with a default set to false? So the `isvalid` method would read as follows:

```
const validate = (node, required, allowChildren = false) => {
  if (required === 'nesting') {
    return validateNesting(node);
  }
  if (allowChildren === true) {
    return hasAccessibleChild(node.parent);
  }
  return validateId(node);
};
```

I'll update the create method to pass in the allowChildren boolean separately so that we're not mixing types with required.

ljharb: yes, that's precisely what i'm suggesting :-)

docs/rules/label-has-for.md
```diff
@@ -16,11 +16,12 @@ This rule takes one optional object argument of type object:
 ```json
 {
     "rules": {
-        "jsx-a11y/label-has-for": [ 2, {
+        "jsx-a11y/label-has-for": [ 3, {
```
ljharb: um, 3 isn't a thing? it's 0/1/2 or off/warn/error.

ljharb:
(this needs a rebase on the command line tho)

jessebeach:
@mjaltamirano, Let's get this in today and we can roll it into the v6 release.

mjaltamirano:
Please ignore the above merge--working on rebasing from the command line locally.

mjaltamirano:
I added `object-assign` as a devDependency to `package.json` because, when I initially pushed my changes without it, CI failed, so I thought it may have been due to its absence.

package.json
```diff
@@ -48,6 +48,7 @@
     "jest": "^20.0.0",
     "jscodeshift": "^0.3.30",
     "minimist": "^1.2.0",
+    "object-assign": "^4.1.1",
```
ljharb: Please use `object.assign` instead of `object-assign` - it's more spec-compliant.

mjaltamirano: Google user error. 😢  Apologies, since you had specified `object.assign`

jessebeach:
I just fixed the merge conflict in the Github GUI. I'm running a large-scale test of the rule right now. I'll merge once it returns without errors. 

Thank you @mjaltamirano !!!

PR #261, opened 2017-06-07
Remove all lockfiles.

### LINKED ISSUE #248: Consider removing yarn or greenkeeper
These two tools don't play nice together. When greenkeeper tries to upgrade a dependency, it doesn't update the yarn lock file, so there are constantly failing builds on those pull requests. yarn essentially renders greenkeeper useless other than reminders to upgrade the lock file manually.

evcohen:
Discussed in #255 - fixes #248


jessebeach:
Nice, thanks for doing this.

PR #274, opened 2017-06-28
Add npm package.json whitelist

gaearon:
There seems to be no need to ship `src`, `flow`, or `scripts` to the user.

Fixes https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/249 (which currently breaks zero config CRA Flow integration).

Verified this works by removing all but these folders.

A whitelist is better than a blacklist in my experience because as project grows, people add infra-related folders to it.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,8 +1,0 @@
-.DS_Store
-.flowconfig
-.travis.yml
-coverage
-node_modules
-npm-debug.log
-reports
-yarn-error.log
--- a/package.json
+++ b/package.json
@@ -16,6 +16,13 @@
     "url": "https://github.com/evcohen/eslint-plugin-jsx-a11y"
   },
   "main": "lib/index.js",
+  "files": [
+    "CHANGELOG.md",
+    "LICENSE.md",
+    "README.md",
+    "docs",
+    "lib"
+  ],
   "scripts": {
     "build": "rimraf lib && babel src --out-dir lib --copy-files",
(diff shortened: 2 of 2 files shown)


evcohen:
thanks @gaearon - @zpao proposed this a while ago in #156, and I believe we decided to ship `src` down for offline use. Since this is a `devDependency`, it's only cost wrt tarball size should be during installation time and should not affect any bundle size. any reason you feel we shouldn't ship `src`?

gaearon:
I'm okay with shipping `src` but shipping `flow` really bit us. I also don't see the point in shipping `.babelrc` (which sometimes confuses users trying to compile `node_modules` with incompatible Babel version—don't ask me why people do that), `__tests__` and `__mocks__` (I can imagine somebody running Jest over `node_modules` for some obscure reason), etc.  In my experience issues like this keep coming up.

But I'm also cool with just adding `flow` to the ignore file if you think that's better. It is currently breaking Flow integration with CRA.

gaearon:
Seems like this is more controversial than I thought 😞 
What about https://github.com/evcohen/eslint-plugin-jsx-a11y/pull/275?

evcohen:
Haha yeah strong opinions both ways 🤷‍♂️ ...100% cool with not shipping `flow`. Will keep this open for a little, but will move convo over to #275

gaearon:
I guess let's keep it broken. I can't think of how to fix this issue while satisfying @ljharb's requirement of keeping the `src`.

PR #275, opened 2017-06-28
Exclude flow from compiled package

gaearon:
Alternative to https://github.com/evcohen/eslint-plugin-jsx-a11y/pull/274.
Should fix https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/249.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -2,6 +2,7 @@
 .flowconfig
 .travis.yml
 coverage
+flow
 node_modules
 npm-debug.log
 reports


evcohen:
So, by not including `flow` directory in this package, flow will ignore this dependency for top-level projects using flow?

gaearon:
Oh, actually maybe I'm wrong 😞 
It's probably `src` causing the issue since it contains `@flow`.

gaearon:
Closing since excluding `src` was already ruled out in past discussion.

Issue #253, opened 2017-06-01
 no-noninteractive-element-interactions rule is triggered with onSubmit event


fernandopasik:
With this example the rule triggers an error:

```
<form onSubmit={ this.handleSubmit }>
</form>
```

axelboc:
I'm experiencing this issue with iframe and onLoad.

axelboc:
The rule description suggests that the rule only applies to mouse and keyboard events, but this doesn't seem to be the case in practice:

> Non-interactive elements should not be assigned mouse or keyboard event listeners.

ljharb:
onSubmit is both a mouse and keyboard event listener.

onLoad, I agree should be allowed on `img`, `iframe`, and `body` tags.

axelboc:
Throwing my two cents about the form: if `onSubmit` event is considered a mouse and keyboard event listener, then shouldn't the `form` be considered as an interactive element?

axelboc:
Throwing my two cents about the form: I don't see any accessibility bad practice in @fernandopasik's example. What about the `reset` event, or events that bubble up like `change` (I think)? Predicting whether a developer uses these events in a non-accessible manner seems way out of scope. I feel like the term "interactions" should only cover what browsers call "user gestures" (the gestures a browser requires to allow the use of some APIs, like the Full Screen API). If I'm not mistaken, these are simply mouse and keyboard events.

fernandopasik:
I thought **onSubmit** and **onReset** are form events, not keyboard nor mouse events. @ljharb please could you share documentation regarding this?

ljharb:
@fernandopasik you're right, they're form events, but form events result from user interaction (keyboard and/or mouse), so conceptually they apply here.

That said, I totally agree that `onSubmit` makes sense on `<form>` (and only on `<form>`), which is why I added the "bug" label immediately :-)

fernandopasik:
@ljharb oh I'm sorry, I didn't notice the bug label! Thanks!

jessebeach:
Agree. Good points. Will update the rule.

jessebeach:
Fixed as of #281

PR #281, opened 2017-07-02
Limit failures in non-interactive interacts to keyboard, mouse and fo…

jessebeach:
Limit failures in non-interactive interactives to keyboard, mouse and focus event handler types.

Addresses #253


ljharb:
@jessebeach There's a lot of handler tests removed and not re-added on "valid" - is there a reason not to preserve them?

jessebeach:
> There's a lot of handler tests removed and not re-added on "valid" - is there a reason not to preserve them?

None whatsoever. I'll put them back. Good catch.

jessebeach:
@ljharb Added back in in #284

PR #286, opened 2017-07-04
Improve travis-ci tests

ljharb:
Runs tests up to node 8; allows failures on EOL'd nodes; speeds up the tests; avoids redundant script runs; splits up flow/lint/tests; only run flow/lint once instead of "every time".


ljharb:
(still at 99.37% for branches, but pretty close to full 100)

src/rules/aria-props.js
```diff
@@ -36,7 +36,7 @@ module.exports = {
   create: context => ({
     JSXAttribute: (attribute) => {
       const name = propName(attribute);
-      const normalizedName = name ? name.toLowerCase() : '';
+      const normalizedName = name.toLowerCase();
```
jessebeach: `propName` can return `undefined` if the type of the AST node is neither `JSXIdentifier` or `JSXNamespacedName`.

src/rules/aria-proptypes.js
```diff
@@ -63,7 +63,7 @@ module.exports = {
   create: context => ({
     JSXAttribute: (attribute) => {
       const name = propName(attribute);
-      const normalizedName = name ? name.toLowerCase() : '';
```
jessebeach: propName can returned undefined if the type of the AST node is neither JSXIdentifier or JSXNamespacedName.

src/rules/aria-role.js
```diff
@@ -41,10 +41,9 @@ module.exports = {
       }
 
       // Get prop name
-      const name = propName(attribute);
-      const normalizedName = name ? name.toUpperCase() : '';
+      const name = propName(attribute).toUpperCase();
```
jessebeach: propName can returned undefined if the type of the AST node is neither JSXIdentifier or JSXNamespacedName.

.travis.yml
```diff
@@ -1,13 +1,31 @@
...
+  - "6"
+  - "5"
+  - "4"
 cache:
   yarn: true
   directories:
     - node_modules
 script:
-  - npm run test:ci
+  - if [ "${FLOW-}" = true ]; then npm run flow; fi
```
ljharb: Yep! It avoids a bash error if the `-x` option is set, and defaults it to the empty string.

src/rules/aria-props.js
ljharb: Given that line 37 means it only runs on `JSXAttribute`, what JSXAttribute could possibly trigger this that isn't one of those two node types?

jessebeach: @ljharb, It feels brittle to me, relying on a side effect of the context rather than responding to the return type of the function. We could either have propName return an empty string as the default case or refactor propName to remove the switch and have it return an indentifier in all cases (make it non-nullable). I'll update that project today and you can include a dependency bump here if you're up for it.

jessebeach: https://github.com/evcohen/jsx-ast-utils/pull/50

src/rules/mouse-events-have-key-events.js
```diff
@@ -45,7 +45,7 @@ module.exports = {
       // Checkout onmouseout / onblur pairing
       const onMouseOut = getProp(attributes, 'onMouseOut');
       const onMouseOutValue = getPropValue(onMouseOut);
-      if (onMouseOut && (onMouseOutValue !== null || onMouseOutValue !== undefined)) {
+      if (onMouseOut && onMouseOutValue != null) {
```
evcohen: what's the benefit of using implicit conversions here vs. explicitly checking for `null` or `undefined`

jessebeach: I do this a lot, too. It's a handy check for null/undefined. The benefit is less code...that's about it :)

evcohen:
> I'd love to get a jsx-ast-utils version bump to go along with these changes once evcohen/jsx-ast-utils#50 is put in a release.

KK - it technically is a breaking change, will publish a major today.

evcohen:
https://github.com/evcohen/jsx-ast-utils/releases/tag/v2.0.0 published

ljharb:
Updated, and audited all uses of `propName` (no further changes were necessary). There were already a plethora of them that were relying on the parser node context never producing a `null` such that I think this was fine as-is, but now they all can rely on it being a string :-)

ljharb:
I'll wait for one more approval before merging.

PR #317, opened 2017-09-04
Add reference to no-access-key

tanipani:
Add WebAIM resource with an explanation why `accesskey` attribute should be avoided.


evcohen:
Thank you!!!

Issue #298, opened 2017-07-19
`label-has-for` false positive when using `allowChildren`


amannn:
When using this rule:

```json
"jsx-a11y/label-has-for": [2, {
  "required": {"every": ["nesting"]},
  "allowChildren": true
}]
```

This code passes the linter check:

```jsx
<label>Firstname</label>
<input type="text" name="firstname" />
```

Currently it passes due to `case 'Literal': return Boolean(child.value);` in [`hasAccessibleChild`](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/src/util/hasAccessibleChild.js) I think.

Or am I missing something?

ljharb:
This should definitely fail. I assume if you set `allowChildren` to false, it fails as expected?

amannn:
Yep, that's correct.

ljharb:
This was added in #259 by @mjaltamirano - do you have any thoughts on how to fix this?

mjaltamirano:
Hey y'all, sorry for the inconvenience. I'll check this first thing tomorrow morning. I suspect it is what @amannn pointed out, in hasAccessibleChild. When I initially wrote #259 I didn't use it and checked specifically for a JSXExpressionContainer. That change might fix it, and I'll add additional test cases to check for that, too.

evcohen:
@mjaltamirano agreed- that's my bad for pointing you in this direction. we should be explicitly checking for a child element that's not `Literal`. sorry about that 😞 and thanks for fixing!

amannn:
No worries, thanks for your help! I'm wondering though what the right heuristic might be.

By looking at `hasAccessibleChild`, I think the following currently pass despite not being accessible (not tested though):

```jsx
// This is the one already mentioned and should be easy to identify
<label>Firstname</label>
<input type="text" name="firstname" />

// Maybe can be prevented by deep traversing the nodes and
// looking if there's an interactive element. If not and there
// is no `htmlFor` prop on the label this is probably an error.
<label><span>Firstname</span></label>
<input type="text" name="firstname" />

// This is pretty hard to identify.
const labels= {firstname: 'Firstname'};
<label>{labels.firstname}</label>
<input type="text" name="firstname" />

// This is also hard to identify
<label><SomeComponent /></label>
<input type="text" name="firstname" />
```

I think the latter two are hard to identify, since it shouldn't lead to false negatives like:

```jsx
<label>{this.renderInput()}</label>
[...]

PR #326, opened 2017-09-13
Introduce the label-has-associated-control rule.

jessebeach:
I've been working on improvements to the `label-has-for` rule and after much toil, I decided to introduce it as a new rule because the logic is sufficiently different. I would also propose that we deprecate `label-has-for` in a future release.

This rule is a little simpler that `label-has-for`. If the `label` wraps an input and has accessible label potential, it passes. If the `label` has an `htmlFor` and has accessible label potential, it passes. "accessible label potential" means the `label` either has text content or `aria-label` or `aria-labelledby` or a custom attribute that will supply a label. The rule recursively looks for these things up to a configured depth. It fails when it can positively determine that any of these conditions are not satisfied.

I've been running this rule over the FB codbase for a month and a half and it's helped me suss out a bevy of issues that I've been fixing. So far it has a zero false positive rate.
 
#298


jessebeach:
@mjaltamirano @evcohen I see this as a replacement for the `label-has-for` rule.

src/util/mayHaveAccessibleLabel.js
```diff
@@ -0,0 +1,88 @@
...
+) {
+  const labellingProps = [
+    'alt', // Assume alt is used correctly on an image
+    'aria-label',
+    'aria-labelledby',
+    ...additionalLabellingProps,
+  ];
+  return openingElement.attributes.some((attribute): boolean => {
+    // We must assume that a spread value contains a labelling prop.
+    if (attribute.type !== 'JSXAttribute') {
```
evcohen: should this be `JSXSpreadAttribute`?

src/util/mayHaveAccessibleLabel.js
```diff
@@ -0,0 +1,88 @@
...
+    // Check for labelling props.
+    if (
+      node.openingElement
+      /* $FlowFixMe */
+      && hasLabellingProp(node.openingElement, additionalLabellingProps)
+    ) {
+      return true;
+    }
+    // Recurse into the child element nodes.
+    if (node.children) {
```
evcohen: again, consider using `some` if possible?

```
const { children = [] } = node;

return children.some((childNode) => checkElement(childNode, depth + 1));
```

src/util/mayHaveAccessibleLabel.js
jessebeach: This guarantees that only `JSXAttribute` values pass rather than just not `JSXSpreadAttribute`.

src/util/mayContainChildComponent.js
```diff
@@ -0,0 +1,52 @@
...
+): boolean {
+  function traverseChildren(
+    node: Node,
+    depth: number,
+  ): boolean {
+    // Bail when maxDepth is exceeded.
+    if (depth > maxDepth) {
+      return false;
+    }
+    if (node.children) {
```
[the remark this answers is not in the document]
jessebeach: I don't understand what you mean here. Wouldn't `traverseChildren` need to be recursive and return `true`/`false`?  So `some` would just be at the top level, right?

src/util/mayHaveAccessibleLabel.js
jessebeach: I don't understand how `some` at the top level makes it more readable. `checkElement` will have a loop that recurses and return true or false. It's easy enough to just return the true when it's found. I think the complexity is O(n) in the worst case but it should short circuit when it finds a true case.

jessebeach:
commit 659082e introduces the `['htmlFor', 'nesting', 'either', 'both']` option under the prop `assert`. 

I still need to address the `allowChildren` option and fail if multiple inputs are found in the same label.

src/util/mayHaveAccessibleLabel.js
evcohen: In my opinion, declarative code more clearly indicates intent. It shouldn't have any impact on runtime complexity, since `Array.some` also short-circuits when it returns a truthy value.

ljharb:
The "assert" option doesn't seem to be in the docs; any chance we could add it?

PR #337, opened 2017-10-04
Refined falsy check of the mixed typed baseAttr.value

ricsam:
Refined the usage of the mixed typed value in the baseAttr object by explicitly wrtiting the expressions equivalent to falsy checking baseAttr.value

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,3 +1,6 @@
+[lints]
+sketchy-null-mixed=warn
+
 [ignore]
 <PROJECT_ROOT>/lib/.*
 <PROJECT_ROOT>/docs/.*
--- a/src/util/attributesComparator.js
+++ b/src/util/attributesComparator.js
@@ -29,12 +29,15 @@ function attributesComparator(
         }
         // Value exists and does not match.
         if (
+          /*
           baseAttr.value !== null &&
           baseAttr.value !== false &&
           baseAttr.value !== undefined &&
           baseAttr.value !== '' &&
           baseAttr.value !== 0 &&
           Object.prototype.hasOwnProperty.call(baseAttr, 'value') &&
+          */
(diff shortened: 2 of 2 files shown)


ricsam:
Motivation: Flow throws error when the type is not explicitly checked

ljharb:
I'm confused; if Flow threw an error, wouldn't the tests on master be failing?

Why can Flow not handle falsiness, which is a common and idiomatic thing to rely on in JS?

ricsam:
@ljharb I got flow errors when working with this file. From the flow [docs](https://flow.org/en/docs/types/mixed/#toc-anything-goes-in-nothing-comes-out)

> When you try to use a value of a mixed type you must first figure out what the actual type is or you’ll end up with an error.

I am using flow version 55

ljharb:
In that case, perhaps before making the change, this PR should update flow, and then making the change actually serves a purpose?

ricsam:
@ljharb I understand the confusion, I was not very clear before, sorry. The error, or rather warning, I received was due to my linting rules. I've included a sketchy-null check linting rule now in .flowconfig and I restored the code to its original [form](https://github.com/evcohen/eslint-plugin-jsx-a11y/pull/337/commits/3daae72adb3da419c26d3a13e5580c59a161c237). Flow will now print the [error message](https://travis-ci.org/evcohen/eslint-plugin-jsx-a11y/jobs/283086719#L1649). Whether this is something important or not to implement is up to you but it would benefit me and others with the same setup while not breaking the functionality for anyone else.

ljharb:
Lint rules should only ever be per-project; you should never have a global linter nor global lint rules installed.

Regardless, even with that configuration change, the old code continues to pass - so I'm still not seeing the issue you have in mind.

Issue #270, opened 2017-06-23
Rule click-events-have-key-events should ignore presentational elements


axelboc:
Rule `click-events-have-key-events` does not behave consistently with rules `no-static-element-interactions` and `no-noninteractive-element-interactions` in that it doesn't ignore elements with `role="none"` or `role="presentation"`.

My use case involves a modal backdrop. The modal already has a close button but I'd also like users to be able to close the modal by clicking on the backdrop. This is a usability enhancement targeted specifically at mouse users, so adding a keyboard listener and using an interactive element instead of a `div` would not make sense.

```
<div
  onClick={this.onBackdropClick}
  role="presentation"
/>
```

jessebeach:
Yes, I agree.

jessebeach:
@axelboc Is this something you'd like to take a

jessebeach:
Fixed in #280 .

axelboc:
Oh thanks!

tbleckert:
When will this be released @jessebeach ? Just curious, would like my lint to report no errors 👀

jessebeach:
Should be in versions 5 and 6: https://github.com/evcohen/eslint-plugin-jsx-a11y/pull/280

PR #347, opened 2017-10-17
The "src" folder is not exist in package

SmirnovW:
Package version 6.0.2 has no the 'src' folder

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -5,6 +5,5 @@
 /flow
 /node_modules
 npm-debug.log
-/src
 /reports
 yarn-error.log


SmirnovW:
@ljharb but this is not cool because for example this issue https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/270 is still in actual :(

ljharb:
@SmirnovW i'm really not sure what you mean; if you're referring to that the fix for #270 isn't released yet, that's not really related to this issue.

jessebeach:
@SmirnovW I'm curious what your use-case is for including `src` in the package?

SmirnovW:
@jessebeach Just look at this:
This is what we have in src '/src/rules/click-events-have-key-events.js'
screenshot from 2017-10-24 10-40-06

And this is what we have in lib
'/lib/rules/click-events-have-key-events.js' after build
screenshot from 2017-10-24 10-40-27

jessebeach:
@SmirnovW certainly, lib contains the processed source. What I'm wondering is what use case you have for poking around the source of a dependency in your project? Are you debugging issues in this plugin? Or working on features for this plugin?

jessebeach:
Closing for now. @SmirnovW, I'm happy to entertain discussion of why the `src` files should be included in the built package.

Just for context, we're waiting on Flow to fix an issue with `node_modules` getting checked that's making it a problem to include `src` in this project: https://github.com/facebook/flow/issues/869

Issue #357, opened 2017-11-20
jsx-a11y/media-has-caption going off for silent video muted property always on


zargold:
In my company we use videos with no possibility of audio so it makes no sense to have a <track> for caption. There is no sound whatsoever and so we even put the muted attribute (without being controlled by a toggle of or off).
Still the media-has-caption error shows up.

Would be a nice feature to disable that for media using the muted attribute (unconditionally)  or have the ability to use role='presentation' or type='silent' or something to have that error go away for those particular videos.

![screen shot 2017-11-20 at 12 29 42 pm](https://user-images.githubusercontent.com/11298496/33032408-24688754-cdef-11e7-9f41-21e6ba2b75e3.png)

> Feature Request

jessebeach:
Great suggestion. We can absolutely check for the `mute` attr in the rule.

It's probably one `if` and a `return` statement. If you put up a PR, we can get it in.

zargold:
Ok everyone is ok with that spec? Specifically:
If muted is there unconditionally then it is ok to not have the ‘<track>’?

ljharb:
I definitely don't see the point of required captions if it's impossible to have audio.

wkentdag:
Is there an ETA for this update to get merged upstream to NPM? I'm on the latest `6.0.3` and still getting the same error.

k80:
👆

PR #358, opened 2017-11-21
allowing for no <track> element (no captions) on muted audio/video

zargold:
This resolves Issue #357: Allows for muted `<video />` `<audio />` aka those with `mute` attribute ALWAYS true to not require captions.


evocateur:
Unfortunately, this didn't fix the issue, as the attribute is named `muted`, _not_ `mute`: https://html.spec.whatwg.org/multipage/media.html#the-video-element

evocateur:
@zargold I opened #416 to fix

PR #359, opened 2017-11-21
adding div & p as non-interactive elements

hozefaj:


jessebeach:
@hozefaj thank you for the PR.

In categorizing the HTML elements, I put them into 3 buckets: static, non-interactive, interactive. The non-interactive and interactive buckets have implicit semantic roles or meaning. Non-interactives either descend from the 'Structure' abstract class in ARIA or from a structure role in the AXObject model.

https://github.com/A11yance/aria-query/blob/master/scripts/roles.json
https://github.com/A11yance/axobject-query/blob/master/scripts/axmodel.json

In the case of `p`, it's structure in the AXObject model; it has no ARIA role -- the net result is non-interactive. In the case of `div`, it has no mapping in either system, so it's just static.

So, would you mind updating the PR to only include `p` in the docs update?

hozefaj:
@jessebeach Made the required info. Thanks for the explanation pretty informative. 👍

Issue #327, opened 2017-09-18
Error dependency `jsx-ast-utils` version 


markduan:
Hi, I saw the version of `jsx-ast-utils` in `package.json` dependencies section of `eslint-plugin-jsx-a11y 6.0.2`  is `2.0.0`, but the actually code I download is `1.4.0`.

I meet the same [issue](https://github.com/evcohen/jsx-ast-utils/issues/33) today, and this [PR](https://github.com/evcohen/jsx-ast-utils/pull/43) fixed it.

Could you please upgrade the version number and republish this repo? Thanks a lot!

hammadzz:
@evcohen this would be nice as currently eslint is breaking because of this problem.

When did try branching off v5.1.1 to fix it (https://github.com/hammadzz/eslint-plugin-jsx-a11y/tree/fix-version-jsx-ast-utils) I ran into a new problem with eslint complaining: 
ESLint couldn't find the plugin "eslint-plugin-jsx-a11y"

Waize:
same here @markduan 
please fix this

cheton:
Hi @evcohen 

The published version of `eslint-plugin-jsx-a11y@6.0.2` on NPM is still using `eslint-plugin-jsx-a11y@^1.4.0`
https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/v6.0.2/package.json#L66
```
"jsx-ast-utils": "^1.4.0"
```

Will be a new version of `eslint-plugin-jsx-a11y` published to NPM to reflect the latest changes of the master branch?

ibiBgOR:
![tweet](https://user-images.githubusercontent.com/6171670/33116353-b9ef6506-cf64-11e7-8447-b5be809abdb5.png)
I tweeted him yesterday.

evcohen:
published [v6.0.3](https://github.com/evcohen/eslint-plugin-jsx-a11y/releases/tag/v6.0.3)

PR #363, opened 2017-12-07
Bumping `jsx-ast-utils` to enable objectRestSpread

jnwng:
[Here's the requisite bump](https://github.com/evcohen/jsx-ast-utils/pull/60) in `jsx-ast-utils` to enable `objectRestSpread`. 

This should address #327 and close https://github.com/evcohen/eslint-plugin-jsx-a11y/pull/356 when published — for some reason the `package.json` showing for `eslint-plugin-jsx-a11y` in NPM has `jsx-ast-utils@^1.4.0` but should be `^2.0.0`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "@coursera/eslint-plugin-jsx-a11y",
+  "name": "eslint-plugin-jsx-a11y",
   "version": "6.0.3-0",
   "description": "Static AST checker for accessibility rules on JSX elements.",
   "keywords": [
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "eslint-plugin-jsx-a11y",
-  "version": "6.0.3-0",
+  "version": "6.0.3",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "eslint-plugin-jsx-a11y",
-  "version": "6.0.2",
+  "version": "6.0.3",
```
ljharb: please revert this; package versions should never be bumped in PRs.

ljharb:
LGTM, but this can't pass until v2.0.2 is available.

Separately, it's not necessary because `^2.0.0` automatically includes v2.0.2.

jnwng:
Just wanted to get a change in to make sure that a new version of
`eslint-plugin-jsx-a11y` gets into npm, so whichever way does it (bumping
this version, or just publishing a new release) works for me!

evcohen:
This should come for free now that [v6.0.3](https://github.com/evcohen/eslint-plugin-jsx-a11y/releases/tag/v6.0.3) is published

PR #376, opened 2018-01-01
chore: pin eslint to 4.13.1

jessebeach:


jessebeach:
> We shouldn't just pin it; we should fix the new linting error.
Righteo, thank you for the insight here.

PR #381, opened 2018-01-02
Introduce utils to get roles from elements

jessebeach:
Provides 3 methods: `getImplicitRole`, `getExplicitRole`, `getComputedRole`.

We retrieve roles from elements in a bunch of rules. We should have utils to do this.

`getImplicitRole` is updated to return `null` when no role is found, rather than empty string.


jessebeach:
Not yet. I know of rules that can use it though. I only updated rules in this PR that absolutely needed to be.

Issue #309, opened 2017-08-16
<img onError={} /> should not trigger no-noninteractive-element-interactions rule


echenley:
The following causes eslint to error on the jsx-a11y/no-noninteractive-element-interactions rule.

```js
<img onError={() => {}} />
```

PR #382, opened 2018-01-02
Allow configuration override on no-noninteractive-element-interactions

jessebeach:
See #309 

Elements like `img`, `iframe` and `body` should allow `onLoad` and `onError` attributes. I've implemented this as configuration for the rule.


jessebeach:
@echenley, feel free to review as well.

evcohen:
from what i can tell, this introduces breaking changes. next release should be a major?

jessebeach:
> from what i can tell, this introduces breaking changes. next release should be a major?

Oh definitely. There are a few more bugs to resolve first, though. I'd like to get the bug count down to zero before another release.

PR #383, opened 2018-01-04
Update dev dependencies to latest

evcohen:
Wanted to put this through a PR since this required allowing failures in node v4  in CI. Will be hard to pick up latest changes in `jest` if node v4 is now unsupported by this dev dependency.


package.json
```diff
@@ -32,31 +32,31 @@
   "devDependencies": {
     "babel-cli": "^6.24.1",
     "babel-core": "^6.25.0",
-    "babel-eslint": "^8.0.1",
-    "babel-jest": "^21.2.0",
+    "babel-eslint": "^8.1.2",
+    "babel-jest": "^22.0.4",
```
ljharb: I would prefer to stay at v21 for the forseeable future rather than upgrading jest.

Jest's (and jsdom's) decision to drop node 4 is massively massively user-hostile, and shouldn't have been made, and tbh any project where it's my decision will be migrating away from either for the forseeable future as a result.

In short, I'm a nonblocking -1 on ever upgrading jest.

evcohen: lol I'm just seeing https://github.com/facebook/jest/issues/5211

I don't have a strong opinion; I agree with you, but April 2018 is a short while away from LTS maintenance ending and we can get around this issue by running jest in the `node` test environment. I think their decision sets a dangerous precedent, but the tradeoff for us is (1) doing the work to migrate and (2) losing out on the simplicity of (what I think is) a really solid testing platform.

ljharb: even when LTS is over for node 4 tho, that doesn't mean we *have* to, nor *should* drop support for it. It's just that dropping support for LTS is unacceptable.

For an eslint plugin, we MUST always support exactly what eslint itself does, and no less - so until eslint drops node 4, we never can.

Zirro: For what it's worth, dropping Node.js v4 support in jsdom 10 was not an arbitrary decision. Several features of the DOM can only be implemented through ES6 Proxies, and they are only available in more recent versions of Node.

jsdom 9.x is still there for those who need Node.js v4 support and can manage without those features.

ljharb: Then those features aren’t things jsdom users on node 4 will need; Proxy can be conditionally checked for. It remains a bad and unnecessary decision.

PR #386, opened 2018-01-07
Allow exception for implicit navigation role in no-redundant-roles rule

calinoracation:
This change adds a configuration option that allows adding an exception for the `nav` element specifying a redundant and explicit `navigation` role as advised at https://www.w3.org/WAI/GL/wiki/Using_HTML5_nav_element#Example:The_.3Cnav.3E_element. This fixes https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/365.

I wasn't sure if the config should be flexible or not so I tried to follow the conventions around other rules. Most of the rules I looked at with options also used `flow`, it wasn't clear if I should use it in the rule so I went ahead and did, can certainly undo it if it's not desired, this is my first real time using it.

### DIFF
--- a/__tests__/src/rules/no-redundant-roles-test.js
+++ b/__tests__/src/rules/no-redundant-roles-test.js
@@ -10,7 +10,6 @@
 // -----------------------------------------------------------------------------
 
 import { RuleTester } from 'eslint';
-import { configs } from '../../../src';
 import parserOptionsMapper from '../../__util__/parserOptionsMapper';
 import rule from '../../../src/rules/no-redundant-roles';
 import ruleOptionsMapperFactory from '../../__util__/ruleOptionsMapperFactory';
@@ -41,26 +40,26 @@ const neverValid = [
   { code: '<button role={`${undefined}button`} />', errors: [expectedError('button', 'button')] },
 ];
 
-const recommendedOptions = (configs.recommended.rules[ruleName][1] || {});
-
 ruleTester.run(`${ruleName}:recommended`, rule, {
   valid: [
     ...alwaysValid,
     { code: '<nav role="navigation" />' },
   ]
-    .map(ruleOptionsMapperFactory(recommendedOptions))
     .map(parserOptionsMapper),
   invalid: neverValid
-    .map(ruleOptionsMapperFactory(recommendedOptions))
     .map(parserOptionsMapper),
 });
 
-ruleTester.run(`${ruleName}:strict`, rule, {
+const noNavExceptionsOptions = { nav: [] };
+
+ruleTester.run(`${ruleName}:recommended`, rule, {
   valid: alwaysValid
+    .map(ruleOptionsMapperFactory(noNavExceptionsOptions))
     .map(parserOptionsMapper),
   invalid: [
     ...neverValid,
     { code: '<nav role="navigation" />', errors: [expectedError('nav', 'navigation')] },
   ]
+    .map(ruleOptionsMapperFactory(noNavExceptionsOptions))
     .map(parserOptionsMapper),
 });
--- a/__tests__/src/rules/no-redundant-roles-test.js
+++ b/__tests__/src/rules/no-redundant-roles-test.js
@@ -10,8 +10,10 @@
 // -----------------------------------------------------------------------------
 
 import { RuleTester } from 'eslint';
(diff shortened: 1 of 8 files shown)


__tests__/src/rules/no-redundant-roles-test.js
```diff
@@ -10,8 +10,10 @@
 // -----------------------------------------------------------------------------
 
 import { RuleTester } from 'eslint';
+import { configs } from '../../../src/index';
```
ljharb: this path can/should just be `../../../src`

src/rules/no-redundant-roles.js
```diff
@@ -24,22 +28,33 @@ module.exports = {
...
+        const explicitRole = getExplicitRole(type, node.attributes);
+
+        if (!implicitRole || !explicitRole) {
+          return;
+        }
+
+        if (implicitRole === explicitRole) {
+          const allowedRoles = (options[0] || {});
+          if (
+            Object.prototype.hasOwnProperty.call(allowedRoles, type) &&
```
ljharb: this is fine, but we can also use the `has` package here

calinoracation: Can definitely add it, I see 3 other rules using the same syntax, would you want me to update those instances as well as part of this PR?

calinoracation:
As far as it being the default, since we're relaxing a rule it seems like we likely could make it the default, perhaps it would be good to hold off on making that change until @jessebeach can take a look as that was the recommended approach in https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/365#issuecomment-351191795?

ljharb:
Yes, agreed on holding off til @jessebeach weighs in

src/rules/no-redundant-roles.js
```diff
@@ -24,22 +29,30 @@ module.exports = {
     schema: [schema],
```
jessebeach: Let's tighten up the schema a little:

```
schema: [{
  type: 'object',
  additionalProperties: {
    type: 'array',
    items: {
      type: 'string',
    },
    uniqueItems: true,
  },
}],
```

ljharb:
I think ideally the default should be the most common configuration, and the "recommended" config should be merely enabling those defaults.

calinoracation:
Is the outcome here to leave as is or to change the default to be the new behavior but allow it to be overridden by config? 

If that was the case then someone would need to add an exception to cause `<nav role="navigation" />` to cause a failure?

jessebeach:
> I think ideally the default should be the most common configuration, and the "recommended" config should be merely enabling those defaults.

@ljharb one consideration that just came to mind. The only way to override this would be to pass in an empty array if one wanted no exceptions to the no redundant roles rule. That seems a little unusual in ESLint land, to pass in empty config to undo a baked-in configuration behavior. What do you think?

> Is the outcome here to leave as is or to change the default to be the new behavior but allow it to be overridden by config?

I think we're still hashing it out :)

ljharb:
@jessebeach that is a fair point, altho this plugin has some pre-existing examples:
 - [anchor-is-valid](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md), "aspects" option
 - [no-distracting-elements](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-distracting-elements.md#rule-details), "elements" option
 - [no-noninteractive-element-interactions](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-noninteractive-element-interactions.md), "handlers" option

etc.

jessebeach:
> that is a fair point, altho this plugin has some pre-existing examples

Indeed. @calinoracation, so let's go with `<nav role="navigation"` passes by default, meaning that if no configuration for overrides is defined, then the default, baked-into-the-rule configuration applies. I would just define a default at the module scope for this.

calinoracation:
Great, will get those changes updated in the next few days!

calinoracation:
Apologies for the delay, this should now be updated based on the discussion above to allow the exception by default, but also allow an empty array to be specified to ignore said exception if desired.

jessebeach:
Huzzah!!! Thank you @calinoracation !!

Let's get this in...

Issue #227, opened 2017-05-06
[v5.0.0] Definition for rule 'jsx-a11y/img-has-alt' was not found (jsx-a11y/img-has-alt)


YUzhva:
Hey hi.

I'm using `create-react-app` as quick starter-kit.

This error appears only with new `v5.0.0` at first line of my `.js` file where I'm importing react, like:
```javascript
import React from 'react';
```

![screenshot](https://lh3.googleusercontent.com/LP-1Pv_Fmt63rmptW2ex967C2dUnDnrTES8eY6vtgMVKCoDMfh9KTq8iTNmpLCwdvF9ei5xHExm1oX0=w2398-h1598-rw)

### How to reproduce
After creating new app `create-react-app my-app` I installed eslint `yarn add -D eslint` and configured it with eslint config script:
```bash
./node_modules/.bin/eslint --init
```

`.eslintrc.json`
```json
{
  "extends": "airbnb",
  "env": {
    "browser": true,
    "jest": true
  },
  "plugins": [
    "react",
    "jsx-a11y",
    "import"
  ],
  "rules": {
    "react/jsx-filename-extension": 0
  }
}
```

devDependencies:
```json
  "devDependencies": {
    "eslint": "^3.19.0",
    "eslint-config-airbnb": "^14.1.0",
    "eslint-plugin-import": "^2.2.0",
    "eslint-plugin-jsx-a11y": "^5.0.0",
    "eslint-plugin-react": "^7.0.0",
```

ljharb:
The Airbnb config currently requires v4, not v5. Run `npm ls` and you'll see that your dependencies are invalid.

PR #407, opened 2018-03-15
fix alt-text code sample

katharinafluch:
addresses #227 
documentation of the "alt-text" rule was not up to date

### DIFF
--- a/docs/rules/alt-text.md
+++ b/docs/rules/alt-text.md
@@ -14,7 +14,7 @@ An `<img>` must have the `alt` prop set with meaningful text or as an empty stri
 
 For images that are being used as icons for a button or control, the `alt` prop should be set to an empty string (`alt=""`).
 
-```js
+```jsx
 <button>
   <img src="icon.png" alt="" />
   Save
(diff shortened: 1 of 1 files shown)


docs/rules/alt-text.md
```diff
@@ -38,7 +38,8 @@ This rule takes one optional object argument of type object:
 ```json
 {
     "rules": {
-        "jsx-a11y/img-has-alt": [ 2, {
+        "jsx-a11y/alt-text": [
+         2, {
```
ljharb: please revert this change

docs/rules/alt-text.md
```diff
@@ -14,7 +14,7 @@ An `<img>` must have the `alt` prop set with meaningful text or as an empty stri
 
 For images that are being used as icons for a button or control, the `alt` prop should be set to an empty string (`alt=""`).
 
-```js
+```html
```
ljharb: this should actually be `jsx`, not `html`.

docs/rules/alt-text.md
katharinafluch: In my version (6.0.3) I fought with img-has-alt for a very long time, because I always got 94 errors with this message that this rule is not defined. With alt-text the whole thing works. 

error_img-has-alt

ljharb: ohhh sorry I missed the rule name change. That's a good change :-) I'm talking about the formatting - can you revert back to the original formatting, but keep your rule name change?

Issue #196, opened 2017-03-24
Add ARIA attributes from ARIA 1.1?


backwardok:
ARIA 1.1 is in Candidate Recommendation (https://www.w3.org/TR/wai-aria-1.1/). Would we be able to add support for the new attributes or is it still too soon?

ljharb:
Given that it's been in that state since October, has there been any movement on it? It'd be ideal to get someone from the W3C to weigh in on how likely it is that this spec will change or be withdrawn prior to full standardization.

wyattdanger:
if @Joanmarie has bandwidth she may be able to weigh in on that 🤞

Joanmarie:
The ARIA Working Group is currently focused on testing so that we can transition ARIA 1.1 to PR. Unfortunately, it turns out that writing and verifying testable statements is not the sort of work that results in hordes of eager volunteers. Who knew?? 😉

Jokes aside, we do not anticipate changes to the 1.1 spec, and I cannot fathom a condition which would cause it to be withdrawn. Authors are already using ARIA 1.1, user agents are already implementing it, and assistive technologies are already adding support for it.

HTH!

ljharb:
@Joanmarie Thanks for confirming!

@backwardok sounds like we should support the new attributes ASAP, as long as it can be done in a backwards-compatible way (with devices that only support 1.0)

wyattdanger:
Oooh do we get this for free with @jessebeach's recent addition of aria-query?

jessebeach:
> Oooh do we get this for free with @jessebeach's recent addition of aria-query

Yes, 😁. The new properties, states and roles are already baked into aria-query

jessebeach:
> Oooh do we get this for free with @jessebeach's recent addition of aria-query?

@wyattdanger yes. If I'm missing anything, add it to this JSON in a PR and I'll rebuild the modules with the updated data: https://github.com/A11yance/aria-query/blob/master/scripts/roles.json

rastersysteme:
Since [aria-query](https://github.com/A11yance/aria-query) published Version 2.0.0 it would be nice to switch so we can benefit from ARIA 1.1 with its changed combobox spec for example.

jessebeach:
@rastersysteme agree. Please propose the PR and we'll get it in.

PR #420, opened 2018-04-11
Update aria-query to 2.0.0

rastersysteme:
This PR allows us to use the updated aria attributes from ARIA 1.1 (see #196).

closes #412


jessebeach:
I will get to this. I'm letting a long queue grow in this repo it seems :) Sorry!

PR #421, opened 2018-04-15
fix(label-has-for): Make error messages for specific

### LINKED ISSUE #419: More specific error messages for 'label-has-for'
To address the confusion voiced in #418, the 'label-has-for' rule should return more specific error messages (especially if the `"every"` option is used)

Currently, this is returned regardless of how the violation is invalid:

https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/1b29f504056a466d08e630b79b3fe68ac0ba433b/src/rules/label-has-for.js#L14

hjylewis:
The general error message 'Form label must have associated control' was
used for every type of violation which was particularly confusing when
the 'every' option was used.

Instead the error messages should specify if the JSX requires
every, some, or a specific type of associated control.

Closes #419

### DIFF
--- a/__tests__/src/rules/label-has-for-test.js
+++ b/__tests__/src/rules/label-has-for-test.js
@@ -19,10 +19,21 @@ import rule from '../../../src/rules/label-has-for';
 
 const ruleTester = new RuleTester();
 
-const expectedError = {
-  message: 'Form label must have associated control',
+const expectedNestingError = {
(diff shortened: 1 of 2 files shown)


hjylewis:
I'll fix the linting error when I get a chance

hjylewis:
@ljharb linting error fixed!

ljharb:
LGTM, i consider this semver-patch and will merge in a few days if no objections.

PR #422, opened 2018-04-17
Fix heading-has-content to ignore headers that are aria-hidden

vdh:
Is it safe to assume that a heading that is `aria-hidden` is excused from the `heading-has-content` rule, since it will be skipped by screen readers?


vdh:
Technically "no point", as it would be for structural or otherwise redundant reasons. Icky usage, I know, but at least an `aria-hidden` heading will no longer cause troubles for screen readers.

jessebeach:
@vdh have you run into a case a header needs to be hidden from assistive technology? I'm ambivalent about the change, though.

jessebeach:
@vdh ya, I would use a div and style it appropriately. The semantic HTML elements shouldn't be transmogrified or hidden.

PR #431, opened 2018-05-22
Fix heading-has-content so it works for custom components

### LINKED ISSUE #429: heading-has-content doesn't work for custom components
Hi there,

Just ran into this issue on v6.0.3. I have a custom heading component and added it to my config like this:

```js
'jsx-a11y/heading-has-content': ['error', { components: ['Title'] }]
```

I'd expect a linting error when trying to use it without content (`<Title />`), but none is given. Still works for the standard h1-h6 elements.

Comparing the code to the `anchor-has-content` rule, it seems the actual `components` array from the options is not what's being appended to the array of headings to check (at [this line of code](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/src/rules/heading-has-content.js#L38)), but rather its parent object. Tweaking it locally to match the pattern in `anchor-has-content` seems to fix it for me:

```js
const options = context.options[0] || {};
const componentOptions = options.components || [];
const typeCheck = headings.concat(componentOptions);
```

pdhoopr:
Fixes #429.

I wasn't totally sure of what to do for tests, and I noticed there weren't any similar `anchor-has-content` tests. So I based mine off of the existing ones for `heading-has-content` and the custom component ones in `anchor-is-valid`. Definitely happy to make changes if something looks off!


ljharb:
The test failure seems like a bug in jest; I'll try to fix it in master.

ljharb:
(filed #432 to fix the tests)

jessebeach:
@pdhoopr thank you for the PR! I'm resolving the merge conflict. I'll merge to master when the Travis run returns green.

PR #437, opened 2018-05-30
added JSXText as a valid accessible child

taddei:
we're using `eslint` and `eslint-plugin-jsx-a11y` here at Ticketmaster and our build failed the other day with this error:

```
{....}/components/sellFlow/index.js
  127:9  error  Headings must have content and the content must be accessible by a screen reader  jsx-a11y/heading-has-content

✖ 1 problem (1 error, 0 warnings)
```

but the incriminating line was this:

```jsx
<h4>Terms and Conditions</h4>
```

I managed to track it down to the fact that the text `Terms and Conditions` is defined as a child element of type `JSXText` and that is not accepted as a valid accessible child. I've added a test to simulate our scenario and added a minor fix. Thanks.

### DIFF
--- a/__tests__/src/util/hasAccessibleChild-test.js
+++ b/__tests__/src/util/hasAccessibleChild-test.js
@@ -36,6 +36,15 @@ describe('hasAccessibleChild', () => {
       expect(hasAccessibleChild(element)).toBe(true);
     });
 
+    it('Returns true for JSXText Element', () => {
+      const child = {
+        type: 'JSXText',
+        value: 'foo',
+      };
+      const element = JSXElementMock('div', [], [child]);
+      expect(hasAccessibleChild(element)).toBe(true);
+    });
+
     it('Returns false for hidden child JSXElement', () => {
       const ariaHiddenAttr = JSXAttributeMock('aria-hidden', true);
       const child = JSXElementMock('div', [ariaHiddenAttr]);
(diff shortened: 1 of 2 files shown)


taddei:
Let me know if there's anything else I can do to get this merged. Thanks.

package.json
```diff
@@ -26,6 +26,7 @@
     "prepublish": "safe-publish-latest && not-in-publish || (npm run lint && npm run flow && npm run jest && npm run build)",
     "pretest": "npm run lint:fix && npm run flow",
     "test": "npm run jest",
+    "test:watcher": "npm run jest -- --watch",
```
ljharb: i don't think this is needed; you can do `npm run jest -- --watch` directly

taddei:
perfect, I removed the package.json update

nwalters512:
When can we expect a new version with this fix to be published? It looks like the last release, 6.0.3, was published 6 months ago.

jessebeach:
@nwalters512 I can get to it this coming Sunday (17 June). This plugin is a labor of love, but it's most definitely also a labor of spare time =D

nwalters512:
Thanks! Really appreciate all the work that's put into it ❤️

mitchellhuang:
@jessebeach any update on this?

jessebeach:
@mitchellhuang just waiting on @evcohen to publish the release: #452

PR #452, opened 2018-06-25
Prepping v6.1.0 release

jessebeach:
- CHANGELOG updates for upcoming version 6.1.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,12 @@
+6.1.0 / 2018-06-26
+==================
+- [new] Support for eslint v5, #451
+- [new] aria-query updated to latest version
+- [new] eslint-config-airbnb-base updated to the latest version
+- [deprecate] The rule label-has-for is deprecated and replaced with label-has-associated-control
+- [fix] heading-has-content updated to work with custom components, #431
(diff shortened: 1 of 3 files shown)


ljharb:
oof, i didn't realize the eslint 5 stuff would require a semver-major :-/

could we maybe backport that into a v6 minor?

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+7.0.0 / 2018-06-25
+==================
+- [new] Support for eslint v5, #451
```
ljharb: i don't see any breaking changes indicated here, "new" means minor?

jessebeach:
@ljharb @evcohen I would like to deprecate `label-has-for` in this release. We'll leave it in the plugin, but add `-deprecated` to the rule name and indicate that it will disappear in a future release. `label-has-for` is superseded by `label-has-associated-control`.

CHANGELOG.md
ljharb: i definitely do think so :-) `v6.1.0` seems ideal here (even if you also want to add `deprecated: true` to `label-has-for`)

PR #460, opened 2018-07-09
Ignore spread props when checking attributes

### LINKED ISSUE #459: Image with expanded props throws `The prop must be a JSXAttribute collected by the AST parser.`
Hi,
I have this wrapper around an image:

```jsx
export default class MaybeImage extends React.Component {
  state = {
    error: false,
  }

  handleError = e => this.setState({ error: true })

  render() {
    const { alt, src, ...props } = this.props
    return !src || this.state.error ? (
      <span {...props}>{alt}</span>
    ) : (
      <img src={src} alt={alt} onError={this.handleError} {...props} />
    )
  }
}
```

But it breaks eslint with an error: **The prop must be a JSXAttribute collected by the AST parser.** It's not just an eslint error, rather unhandled javascript exception. I think #7 might be related, but even when I disable the rule, it still breaks.

[This line](https://github.com/evcohen/jsx-ast-utils/blob/master/src/propName.js#L5) in `jsx-ast-utils` gets the prop name, but it doesn't count with spread operator. This is the content of `prop` which throws an error:

```js
   Node {
     type: 'JSXSpreadAttribute',
     argument:
      Node {
        type: 'Identifier',
        name: 'props',
        _babelType: 'Identifier',
     },
     _babelType: 'JSXSpreadAttribute',
     parent:
      Node {
        type: 'JSXOpeningElement',
        attributes: [Array],
        name: [Node],
        selfClosing: true,
        _babelType: 'JSXOpeningElement',
      } }
```

What's your suggestion here? Pass props explicitely so it can be statically analyzed? Or is it bug in `jsx-ast-utils`? I think it should show an eslint error or warning at best, but not throw an exception.

Cheers 👋

tricoder42:
Fixes #459

### DIFF
--- a/__tests__/src/rules/no-noninteractive-element-interactions-test.js
+++ b/__tests__/src/rules/no-noninteractive-element-interactions-test.js
@@ -116,7 +116,7 @@ const alwaysValid = [
   { code: '<html onClick={() => {}} />;' },
   { code: '<i onClick={() => {}} />;' },
   { code: '<iframe onLoad={() => {}} />;' },
-  { code: '<img onError={() => {}} />;' },
+  { code: '<img {...props} onError={() => {}} />;' },
   { code: '<img onLoad={() => {}} />;' },
   { code: '<ins onClick={() => {}} />;' },
   { code: '<kbd onClick={() => {}} />;' },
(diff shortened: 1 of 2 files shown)


jessebeach:
That'll do it. That rule should probably be obtaining the `attrs` in a way that filters the spread attributes before iterating over the set. This gets us past the error now, though, so I'm merging. 

Thank you @tricoder42 !!

brokentone:
I've come here with this error as well, looks like a regression in `6.1.0`. Would love to see a release with this patch!

jessebeach:
@brokentone I'll put putting up a new version tonight. Just need to do the housekeeping for CHANGELOG and wait for Travis.

Issue #455, opened 2018-07-03
label-has-for: broken in v6.1.0


JustFly1984:
```
/Users/justfly/projects/aggrow-static/src/components/common/input-combobox.js
  265:9   error  The value for aria-controls must be a idlist      jsx-a11y/aria-proptypes
  267:9   error  The value for aria-owns must be a idlist          jsx-a11y/aria-proptypes
  297:11  error  The value for aria-activedescendant must be a id  jsx-a11y/aria-proptypes
  298:11  error  The value for aria-controls must be a idlist      jsx-a11y/aria-proptypes

/Users/justfly/projects/aggrow-static/src/components/common/input-quantity-go-to-cart.js
  29:3  error  Form label must have ALL of the following types of associated control: nesting, id  jsx-a11y/label-has-for

/Users/justfly/projects/aggrow-static/src/components/common/input-quantity-max.js
  28:3  error  Form label must have ALL of the following types of associated control: nesting, id  jsx-a11y/label-has-for

/Users/justfly/projects/aggrow-static/src/components/common/input-quantity-remove.js
  28:3  error  Form label must have ALL of the following types of associated control: nesting, id  jsx-a11y/label-has-for

/Users/justfly/projects/aggrow-static/src/components/common/input-select.js
  249:9   error  The value for aria-controls must be a idlist      jsx-a11y/aria-proptypes
  251:9   error  The value for aria-owns must be a idlist          jsx-a11y/aria-proptypes
  277:11  error  The value for aria-activedescendant must be a id  jsx-a11y/aria-proptypes
  278:11  error  The value for aria-controls must be a idlist      jsx-a11y/aria-proptypes

/Users/justfly/projects/aggrow-static/src/components/common/labeled-checkbox.js
  38:3  error  Form label must have ALL of the following types of associated control: nesting, id  jsx-a11y/label-has-for

/Users/justfly/projects/aggrow-static/src/components/common/labeled-combobox.js
  43:5  error  Form label must have ALL of the following types of associated control: nesting, id  jsx-a11y/label-has-for

/Users/justfly/projects/aggrow-static/src/components/common/labeled-select.js
  39:5  error  Form label must have ALL of the following types of associated control: nesting, id  jsx-a11y/label-has-for

/Users/justfly/projects/aggrow-static/src/components/common/labeled-textarea.js
  38:3  error  Form label must have ALL of the following types of associated control: nesting, id  jsx-a11y/label-has-for

[...]

PR #464, opened 2018-07-10
Change every to some in label-has-for

jessebeach:
The default value for `label-has-for` should be `some`, not `every` to maintain the same behavior as in v6.0.x:

```
const required = options.required || { some: ['nesting', 'id'] };
```


ljharb:
Hmm - i thought the default was always "every"?

ljharb:
This isn't the bug I was experiencing - I was finding that "id" wasn't detecting the link when it was nested.

ljharb:
It's possible #439 is related?

jessebeach:
@ljharb blarg, hmm. Ya, my fix probably isn't right :( Do you have a code example I can put into a test case?

ljharb:
Hmm - so I went to the repo where I'd disabled the rule, re-enabled it, and everything still passes.

However, #455 has one: https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/455#issuecomment-402330447

That example is both nested and id-linked, so it should be passing on any settings.

jessebeach:
> It's possible #439 is related?

You traced the error to the right place, but your changes are a red herring. You simply made the rule more specific...it never handled nesting deeper than children. 

I've added an iterative breadth-first search for `<input />` that validates @JustFly1984 case.

jessebeach:
@ljharb @evcohen I still think we should change the default in the rule from `every` to `some`. `every` is too strict and it can always be overridden via config.

ljharb:
I don't find it too strict; either way, changing the defaults imo is a breaking change, and I think we should get this fix out separately.

jessebeach:
> I don't find it too strict; either way, changing the defaults imo is a breaking change, and I think we should get this fix out separately.

That's reasonable. I'll update the PR.

PR #468, opened 2018-08-03
Update aria-props.md to link to the correct doc

csi-lk:
The current one 404s

Pretty straight forward 😇


evcohen:
Thank you!!!!

PR #483, opened 2018-09-06
Rule label-has-for is deprecated

spyl94:
A deprecated rule should not be present in recommended or strict configurations.


src/index.js
```diff
@@ -73,7 +73,6 @@ module.exports = {
             ],
           },
         ],
-        'jsx-a11y/label-has-for': 'error',
         'jsx-a11y/label-has-associated-control': 'error',
```
ljharb: hmm, the addition of the associated control rule to these configs should have been semver-major

ljharb:
Either way, “deprecated” is a documentation thing to notify people; i don’t see much point in removing the rule from these configs now (as opposed to later, when the actual rule is removed)

spyl94:
Changelog (https://github.com/evcohen/eslint-plugin-jsx-a11y/releases/tag/v6.1.0) explains that this rule has been replaced by `label-has-associated-control` which is already present.

Right now I need to disable it, but I agree this change should bump version or wait 7.0.

evcohen:
> Shall we also document in https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-for.md that `label-has-associated-control` is designed replacement of this rule? I didn't know this until I look into this pull request.

Yes, we should! @franklinyu feel free to post a PR

jessebeach:
@spyl94, thank you for the PR. We'll go with the documentation notice in #493 at this time. We'll remove the config line as you proposed here in the next major release.

Issue #485, opened 2018-10-02
Discussion: Anchor is valid rule used in Create React App


AlmeroSteyn:
From a discussion I had with @gaearon.

Currently this rule is activated in Create React App. Which does a lot of good. However, due to the reach of the CLI tool it can sometimes cause confusion for developers who want to upgrade to newer versions. Developers lacking the a11y knowledge who suddenly get confronted by new errors on the console.

For example: https://github.com/facebook/create-react-app/issues/4141

I want to make a PR proposal to update the docs and, possibly, the error message of this rule to assist these developers to easily find the information in question.

But first I want to know how you feel about such a change.

Consider the situation where a developer HAS to have a clickable thing look like an `anchor` due to design forced upon them, would it be OK to add a paragraph in the docs on how to style a button to look like a link? I know that ideally we would not want that but what is happening now is that developers facing this situation just suppress the rule. Where instead a `button` element styled as a link would be a large improvement already instead of an `anchor` without an `href`.

Also the PR would suggested to change the error message of the `noHref` aspect to something along the lines of:
"The href attribute requires a valid value for the link to be keyboard accessible. Provide a valid, navigable address as the href value. Alternatively, change it to a <button> with appropriate styles. Learn more: <url>"

Please let me know what you think. As I said this rule is activated in every react app started with Create React App and could go a long way to help with education.

ljharb:
I think better documentation is always better, and as long as they stress very heavily that such designs are subpar (giving devs ample ability to push back on design with an authoritative reference), providing instructions of how to have proper semantic elements with the aesthetics they’re required to impose seems like a much better outcome than having them disable the rule.

AlmeroSteyn:
Awesome, I will come up with a PR asap.

AlmeroSteyn:
PR in at #486

evcohen:
Fixed in #486

evcohen:
[v6.1.2](https://github.com/evcohen/eslint-plugin-jsx-a11y/releases/tag/v6.1.2) published

PR #486, opened 2018-10-02
Add link-type styling recommendation to anchor-is-valid

AlmeroSteyn:
As discussed in #485, developers face some situations where it is impossible to change design to the visual appearance of a button, or simply get faced with this rule's message and have no idea how to proceed.

This PR addx a section to the documentation of the rule explaining how a link can be changed into a button and restyled as a link in extreme circumstances.

It also adds clearer error massages.

Summary of changes:

1. Add an explanatory section to the docs:

image

2. Change the error messsage to:

PreferButton message: `Anchor used as a button. Anchors are primarily expected to navigate. Use the button element instead. Learn more: https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md`

noHrefMessage: `The href attribute is required for the link to be keyboard accessible. Provide a valid, navigable address as the href value. Alternatively, use a button and change it with appropriate styles. Learn more: https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md`

invalidHrefMessage: `The href attribute requires a valid value to be accessible. Provide a valid, navigable address as the href value. Alternatively, use a button and change it with appropriate styles. Learn more: https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md`

### DIFF
--- a/__tests__/src/rules/anchor-is-valid-test.js
+++ b/__tests__/src/rules/anchor-is-valid-test.js
@@ -18,11 +18,11 @@ import rule from '../../../src/rules/anchor-is-valid';
 
 const ruleTester = new RuleTester();
 
-const preferButtonErrorMessage = 'Anchor used as a button. Anchors are primarily expected to navigate. Use the button element instead.';
+const preferButtonErrorMessage = 'Anchor used as a button. Anchors are primarily expected to navigate. Use the button element instead. Learn more: https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md';
 
-const noHrefErrorMessage = 'The href attribute is required on an anchor. Provide a valid, navigable address as the href value.';
+const noHrefErrorMessage = 'The href attribute is required for an anchor to be keyboard accessible. Provide a valid, navigable address as the href value. If you cannot provide an href, but still need the element to resemble a link, use a button and change it with appropriate styles. Learn more: https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md';
 
-const invalidHrefErrorMessage = 'The href attribute requires a valid address. Provide a valid, navigable address as the href value.';
+const invalidHrefErrorMessage = 'The href attribute requires a valid value to be accessible. Provide a valid, navigable address as the href value. If you cannot provide a valid href, but still need the element to resemble a link, use a button and change it with appropriate styles. Learn more: https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md';
 
 const preferButtonexpectedError = {
   message: preferButtonErrorMessage,
--- a/docs/rules/anchor-is-valid.md
+++ b/docs/rules/anchor-is-valid.md
@@ -88,6 +88,47 @@ If you need to create an interface element that the user can mouse over or mouse
 
 In the example immediately above an `onClick` event handler was added to provide the same experience mouse users enjoy to keyboard-only and touch-screen users. Never fully rely on mouse events alone to expose functionality.
 
+### Case: I understand the previous cases but still need an element resembling a link that is purely clickable
+
+We recommend, without reserve, that elements resembling anchors should navigate. This will provide a superior user experience to a larger group of users out there.
+
+However, we understand that developers are not always in total control of the visual design of web applications. In cases where it is imperative to provide an element resembling an anchor that purely acts as a click target with no navigation as result, we would like to recommend a compromise.
+
+Again change the element to a `<button>`: 
+
+```jsx
+<button 
+  type="button"
+  className="link-button" 
+  onClick={() => this.setState({showSomething: true})}>
+    Press me, I look like a link
+</button>
+``` 
+
+Then use styling to change its appearance to that of a link:
+
+```css
+.link-button {
+  background-color: transparent;
+  border: none;
+  cursor: pointer;
+  text-decoration: underline;
+  display: inline;
+  margin: 0;
+  padding: 0;
+}
+
+.link-button:hover,
+.link-button:focus {
+text-decoration: none;
+}
+```
+
+This button element can now also be used inline in text.
+
+Once again we stress that this is an inferior implementation and some users will encounter difficulty to use your website, however, it will allow a larger group of people to interact with your website than the alternative of ignoring the rule's warning.
+
+
 ### References
   1. [WebAIM - Introduction to Links and Hypertext](http://webaim.org/techniques/hypertext/)
   1. [Links vs. Buttons in Modern Web Applications](https://marcysutton.com/links-vs-buttons-in-modern-web-applications/)
--- a/src/rules/anchor-is-valid.js
+++ b/src/rules/anchor-is-valid.js
@@ -15,11 +15,11 @@ import { generateObjSchema, arraySchema, enumArraySchema } from '../util/schemas
 
 const allAspects = ['noHref', 'invalidHref', 'preferButton'];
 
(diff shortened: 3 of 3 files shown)


__tests__/src/rules/anchor-is-valid-test.js
```diff
@@ -18,11 +18,11 @@ import rule from '../../../src/rules/anchor-is-valid';
 
 const ruleTester = new RuleTester();
 
-const preferButtonErrorMessage = 'Anchor used as a button. Anchors are primarily expected to navigate. Use the button element instead.';
+const preferButtonErrorMessage = 'Anchor used as a button. Anchors are primarily expected to navigate. Use the button element instead. Learn more: https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md';
 
-const noHrefErrorMessage = 'The href attribute is required on an anchor. Provide a valid, navigable address as the href value.';
+const noHrefErrorMessage = 'The href attribute is required for the link to be keyboard accessible. Provide a valid, navigable address as the href value. Alternatively, use a button and change it with appropriate styles. Learn more: https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md';
```
ljharb: It's not really a link unless it's an `<a>` (an anchor) with an href; "on an anchor" is correct here. What about that is unclear?

Separately, if there's an href it should only ever be an `<a>` - the instructions should be to help people style the `<a>` as a button, not to use the wrong tag.

__tests__/src/rules/anchor-is-valid-test.js
```diff
@@ -18,11 +18,11 @@ import rule from '../../../src/rules/anchor-is-valid';
...
 
-const invalidHrefErrorMessage = 'The href attribute requires a valid address. Provide a valid, navigable address as the href value.';
+const invalidHrefErrorMessage = 'The href attribute requires a valid value to be accessible. Provide a valid, navigable address as the href value. Alternatively, use a button and change it with appropriate styles. Learn more: https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md';
```
ljharb: maybe instead of "alternatively", say "if you need something that looks like a link but is not one, use a button…"?

docs/rules/anchor-is-valid.md
```diff
@@ -88,6 +88,46 @@ If you need to create an interface element that the user can mouse over or mouse
...
+### Case: I understand the previous cases but still need an element resembling a link that is purely clickable
+
+We recommend, without reserve, that elements resembling anchors should navigate. This will provide a superior user experience to a larger group of users out there.
+
+However, we understand that developers are not always in total control of the visual design of web applications. In cases where it is imperative to provide an element resembling an anchor that purely acts as a click target with no navigation as result, we would like to recommend a compromise.
+
+Again change the element to a `<button>`: 
+
+```jsx
+<button
```
ljharb: all `<button>` elements should have an explicit `type`; let's be sure our examples do too. In this case, `type="button"` (or else it'll be a submit button).

docs/rules/anchor-is-valid.md
```diff
@@ -88,6 +88,46 @@ If you need to create an interface element that the user can mouse over or mouse
...
+
+We recommend, without reserve, that elements resembling anchors should navigate. This will provide a superior user experience to a larger group of users out there.
+
+However, we understand that developers are not always in total control of the visual design of web applications. In cases where it is imperative to provide an element resembling an anchor that purely acts as a click target with no navigation as result, we would like to recommend a compromise.
+
+Again change the element to a `<button>`: 
+
+```jsx
+<button 
+  className="link-button"
```
ljharb: i'm not sure if it's better to recommend inline styles or a className here, given the complexity of CSS methods.

docs/rules/anchor-is-valid.md
```diff
@@ -88,6 +88,46 @@ If you need to create an interface element that the user can mouse over or mouse
...
+
+.link-button:hover,
+.link-button:focus {
+text-decoration: none;
+}
+```
+
+This button element can now also be used inline in text.
+
+Once again we stress that this is an inferior implementation and some users will encounter difficulty to use your website, however, it will allow a larger group of people to interact with your website than the alernative.
```
ljharb: "than the alternative of ignoring the rule's warning"?

docs/rules/anchor-is-valid.md
AlmeroSteyn: I'm used to CSS so this did not strike me as complex. Can you even specify pseudoclasses in inline styles?

docs/rules/anchor-is-valid.md
AlmeroSteyn: Changed.

AlmeroSteyn:
I have addressed all the review comments aside from the comment of CSS vs inline styles. 

Let me know what you think.

docs/rules/anchor-is-valid.md
ljharb: true, perhaps not - i guess this is fine

gaearon:
Can we get this in ASAP? Since this warning is showing up in CRA 2.0 apps, it's important that people understand what it's about before they disable it. This PR helps explain it.

src/rules/anchor-is-valid.js
```diff
@@ -15,11 +15,11 @@ import { generateObjSchema, arraySchema, enumArraySchema } from '../util/schemas
...
 
-const noHrefErrorMessage = 'The href attribute is required on an anchor. Provide a valid, navigable address as the href value.';
+const noHrefErrorMessage = 'The href attribute is required for an anchor to be keyboard accessible. Provide a valid, navigable address as the href value. If you cannot provide an href, but still need the element to resemble a link, use a button and change it with appropriate styles. Learn more: https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md';
 
-const invalidHrefErrorMessage = 'The href attribute requires a valid address. Provide a valid, navigable address as the href value.';
+const invalidHrefErrorMessage = 'The href attribute requires a valid value to be accessible. Provide a valid, navigable address as the href value. If you cannot provide a valid href, but still need the element to resemble a link, use a button and change it with appropriate styles. Learn more: https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md';
```
Timer: Could we run these links through a url shortener? 😄 i.e. https://goo.gl/s89kbn

ljharb: what's the benefit of that, versus the loss of clarity on where it will redirect to?

Timer: It'd reduce console noise for users who run into this a lot but know how to handle it.

AlmeroSteyn: I can do this, if required :-)

@ljharb Do you think shortening the link will cause enough clarity loss not to cut the verbosity?

Timer: We could use an aliased shortlink for improved clarity, e.g. https://tinyurl.com/anchor-is-valid

ljharb: I don’t think using a shortlink is a good idea, unless it’s one provided by Github.

AlmeroSteyn: @Timer, as you raised it, is this something you can live with in CRA? Would be awesome to keep this rule alive in there 😄

Timer: Yeah, we'll probably just replace the link ourselves in the output formatter. Thanks!

AlmeroSteyn: Awesome!

PR #487, opened 2018-10-05
Make aria-role case sensitive

### LINKED ISSUE #472: aria-role rule: only lowercase values should be allowed.
Currently, the following "pass", but should be considered invalid.

```
role="Button"
role="Presentation"
```

qw-in:
Fixes #472 

Perhaps this is a naive solution but the tests pass :smile:

### DIFF
--- a/src/rules/aria-role.js
+++ b/src/rules/aria-role.js
@@ -54,7 +54,7 @@ module.exports = {
       // value isn't in the form of a literal.
       if (value === undefined || value === null) { return; }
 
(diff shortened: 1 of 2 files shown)


src/rules/aria-role.js
```diff
@@ -54,7 +54,7 @@ module.exports = {
       // value isn't in the form of a literal.
       if (value === undefined || value === null) { return; }
 
-      const normalizedValues = String(value).toLowerCase().split(' ');
```
evcohen: s/normalizedValues/values since we aren't normalizing anything.

PR #488, opened 2018-10-05
Fix error when type attribute is truthy

### LINKED ISSUE #400: Crash when typing <input type/>
I am very new to node.js and React.
I am using ESLint in Visual Studio Code.

Typing `<input type/>` in my jsx file alone is enough to trigger this error.

I was actually entering `<input />` first, then trying to fill back `type=...` later, and this error shows up.

```
[Error - 11:08:25] ESLint stack trace:
[Error - 11:08:25] TypeError: value.toUpperCase is not a function
    at Object.getImplicitRoleForInput [as input] (proj\node_modules\eslint-plugin-jsx-a11y\lib\util\implicitRoles\input.js:19:19)
    at getImplicitRole (proj\node_modules\eslint-plugin-jsx-a11y\lib\util\getImplicitRole.js:24:41)
    at JSXOpeningElement (proj\node_modules\eslint-plugin-jsx-a11y\lib\rules\role-supports-aria-props.js:47:108)
    at listeners.(anonymous function).forEach.listener (proj\node_modules\eslint\lib\util\safe-emitter.js:47:58)
    at Array.forEach (native)
    at Object.emit (proj\node_modules\eslint\lib\util\safe-emitter.js:47:38)
    at NodeEventGenerator.applySelector (proj\node_modules\eslint\lib\util\node-event-generator.js:251:26)
    at NodeEventGenerator.applySelectors (proj\node_modules\eslint\lib\util\node-event-generator.js:280:22)
    at NodeEventGenerator.enterNode (proj\node_modules\eslint\lib\util\node-event-generator.js:294:14)
    at CodePathAnalyzer.enterNode (proj\node_modules\eslint\lib\code-path-analysis\code-path-analyzer.js:608:23)
```

// package.json
```
...
    "eslint": "^4.15.0",
    "eslint-config-airbnb": "^16.1.0",
    "eslint-plugin-import": "^2.8.0",
    "eslint-plugin-jsx-a11y": "^6.0.3",
    "eslint-plugin-react": "^7.6.1",
...
```

// .eslintrc.json
```
...
  "extends": [
    "airbnb",
    "eslint:recommended",
    "plugin:react/recommended"
  ],
...
```

qw-in:
Fixes #400 

The null check was not sufficient as jsx can pass non string truthy values. For example:
```jsx
<input type />
```

### DIFF
--- a/src/util/implicitRoles/input.js
+++ b/src/util/implicitRoles/input.js
@@ -9,7 +9,7 @@ export default function getImplicitRoleForInput(attributes) {
   if (type) {
     const value = getLiteralPropValue(type) || '';
 
-    switch (value.toUpperCase()) {
+    switch (typeof value === 'string' && value.toUpperCase()) {
       case 'BUTTON':
       case 'IMAGE':
       case 'RESET':
--- a/__tests__/src/util/implicitRoles/input-test.js
+++ b/__tests__/src/util/implicitRoles/input-test.js
@@ -29,4 +29,7 @@ describe('isAbstractRole', () => {
   it('works for the default case', () => {
     expect(getImplicitRoleForInput([JSXAttributeMock('type', '')])).toBe('textbox');
   });
+  it('works for the true case', () => {
+    expect(getImplicitRoleForInput([JSXAttributeMock('type', true)])).toBe('textbox');
+  });
 });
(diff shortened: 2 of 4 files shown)


src/util/implicitRoles/input.js
```diff
@@ -9,7 +9,7 @@ export default function getImplicitRoleForInput(attributes) {
   if (type) {
     const value = getLiteralPropValue(type) || '';
 
-    switch (value.toUpperCase()) {
+    switch (typeof value === 'string' && value.toUpperCase()) {
```
ljharb: in both of these cases, if it's not a string, prior to your change it hits the same return value in the default case, and after your change it hits the same return value outside the switch statement. It doesn't seem like your changes are needed to make the tests pass.

PR #489, opened 2018-10-06
Special case for checkbox with role switch in role-has-required-aria-props

### LINKED ISSUE #484: `<input type="checkbox" role="switch" />` should pass jsx-a11y/role-has-required-aria-props
`<input type="checkbox" role="switch" />` should pass jsx-a11y/role-has-required-aria-props

qw-in:
Fixes #484 

This passes the tests but I could use some context. Is it only this exact case that should pass?

Is it still valid with multiple roles?
```html
<input type="checkbox" role="switch checkbox" />
```


qw-in:
@jessebeach that makes sense 😊

PR #491, opened 2018-10-15
[484] Fix role-has-required-aria-props for semantic elements like input[checkbox]

jessebeach:
Fix for #484 .

I introduced a util that identifies semantic elements with a role that "modifies" their role, casting it up. So,

```
<input type="checkbox" role="switch" />
```

becomes a switch component and it includes the `aria-checked` concept already.


evcohen:
Nice, thanks!

PR #504, opened 2018-11-23
No implicit role for `<img>` with `alt=""`

othree:
Hi, I found the implicit role of `` with `alt=""` is incorrect. 
The document from W3C says it should have the corresponding role.

Ref: <https://www.w3.org/TR/html-aria/#docconformance>

### DIFF
--- a/__tests__/src/rules/role-supports-aria-props-test.js
+++ b/__tests__/src/rules/role-supports-aria-props-test.js
@@ -141,23 +141,8 @@ ruleTester.run('role-supports-aria-props', rule, {
     // this will have global
     { code: '<link aria-checked />' },
 
-    // IMG TESTS - implicit role is `presentation`
-    { code: '<img alt="" aria-atomic />' },
-    { code: '<img alt="" aria-busy />' },
-    { code: '<img alt="" aria-controls />' },
-    { code: '<img alt="" aria-describedby />' },
-    { code: '<img alt="" aria-disabled />' },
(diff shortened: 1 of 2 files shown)


othree:
I didn't update the test because I am not sure how to test `role=''`

jessebeach:
@othree I see. Yes, that's not technically correct. I agree with the change.

othree:
I updated the test. And found if no 'role' provided. The aria attribute check will always pass.
So I update the invalid case to `alt='foobar'` instead of `alt=''`.

__tests__/src/rules/role-supports-aria-props-test.js
```diff
@@ -141,23 +141,8 @@ ruleTester.run('role-supports-aria-props', rule, {
     // this will have global
     { code: '<link aria-checked />' },
 
-    // IMG TESTS - implicit role is `presentation`
-    { code: '<img alt="" aria-atomic />' },
-    { code: '<img alt="" aria-busy />' },
```
ljharb: why remove all these?

othree: Because the implicit role is not `presentation`.
The only way to change the role to `presentation` is by assigning it in the attribute.
But I didn't see any test case in this file have an assigned role.
All cases are using other ways to switch role, e.g. change value of `type`

jessebeach:
@othree this is great, thank you for the test fixes. I appreciate the punctilious attention to spec detail :)

PR #508, opened 2018-12-10
Disable jsx-a11y/label-has-for in recommended

greypants:
Since there's [issues with label-has-for](https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/455)  and it's [deprecated](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-for.md) , it shouldn't be a recommended rule, right?

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -76,7 +76,7 @@ module.exports = {
             ],
           },
         ],
-        'jsx-a11y/label-has-for': 'error',
+        'jsx-a11y/label-has-for': 'off',
         'jsx-a11y/label-has-associated-control': 'error',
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -76,7 +76,7 @@ module.exports = {
             ],
           },
         ],
-        'jsx-a11y/label-has-for': 'error',
+        'jsx-a11y/label-has-for': 'off',
         'jsx-a11y/label-has-associated-control': 'error',
```
ljharb: enabling `label-has-associated-control` should have been a semver-major; removing `label-has-for` is not, so this seems fine.

PR #515, opened 2018-12-27
Control has associated label

jessebeach:
This rule is the inverse of `label-has-associated-control`.

Check the Markdown file for details of the rule.

It's intended to catch unlabelled buttons, checkboxes, radios, menuitems, options, etc. I did a bunch of QA work to finagle the configs so that the output is actionable.

The danger here is that we encourage developers to put labels on things like containers, which really shouldn't be labelled. The exceptions to that rule are dialogs and alerts, but I opened a new issue to write a rule for those: #516 

There really isn't a difference between the behavior of recommended and strict, so I made the configs the same for both.


__tests__/src/rules/control-has-associated-label-test.js
```diff
@@ -0,0 +1,50 @@
...
+    { code: '<button><span aria-labelledby="js_1" /></button>' },
+    { code: '<button>{sureWhyNot}</button>' },
+    // Interactive Roles
+    { code: '<div role="button">Save</div>' },
+  ].map(parserOptionsMapper),
+  invalid: [
+    { code: '<button />', errors: [expectedError] },
+    { code: '<button><span /></button>', errors: [expectedError] },
+    { code: '<button><img /></button>', errors: [expectedError] },
+    { code: '<button><span title="This is not a real label" /></button>', errors: [expectedError] },
```
ljharb: should there be tests for all the other input types as well? It would ensure your intentions are unambiguous :-)

src/index.js
```diff
@@ -58,10 +59,12 @@ module.exports = {
         'jsx-a11y/aria-role': 'error',
         'jsx-a11y/aria-unsupported-elements': 'error',
         'jsx-a11y/click-events-have-key-events': 'error',
+        'jsx-a11y/control-has-associated-label': 'error',
```
ljharb: this needs to be set to “off”, or else the change is semver-major (which we might as well avoid as long as possible)

jessebeach:
Should we enforce a label on `canvas`, `audio` and `video` tags?

jessebeach:
The results from my internal tests look good. About 250 errors over ~70K files. Keep in mind that we've done a TON of work to ferret out unlabelled elements. These represent the ones that we had no signal for.

I just need to fill out the `control-has-associated-label.md` file and this rule should be good to go.

jessebeach:
I'm finding lots of juicy examples where all the rules in the plugin have been followed **except** that the element doesn't have a label. This is the last step!

jessebeach:
@ljharb @evcohen thoughts on merging this new rule in? I ran it over the FB codebase and the results are great. Fixable and accurate.

keithmadden-toast:
@jessebeach Wasn't sure if I should open an issue for this but this PR adds `control-has-associated-label` to recommended & strict modes but only disables `label-has-for` for recommended. Is that the intended behaviour? I'm happy to create a follow-up PR if not 🙂

backwardok:
Hey @jessebeach ! I was looking through some of the rules that exist and I noticed that this rule isn't enabled for either recommended or strict. Was that intended? I mostly ask because the [Airbnb config uses it](https://github.com/airbnb/javascript/blob/711aeb650db14c3b8cc329511ac508ee05bc1af6/packages/eslint-config-airbnb/rules/react-a11y.js#L70-L97) and I was surprised it wasn't listed on the README as an available rule

ljharb:
@backwardok adding a new rule to a builtin config is a breaking change, so it's planned for whenever we end up doing a semver-major bump.

PR #522, opened 2019-01-09
Fix documentation #510

### LINKED ISSUE #510: Description of role="presentation" in docs is misleading
The no-static-element-interactions docs [currently state](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/e53906d0b2a9402b019625349ed2d58d178b3239/docs/rules/no-static-element-interactions.md#case-this-element-is-not-a-button-link-menuitem-etc-it-is-catching-bubbled-events-from-elements-that-it-contains):

### Case: This element is not a button, link, menuitem, etc. It is catching bubbled events from elements that it contains

If your element is catching bubbled click or key events from descendant elements, then the proper role for this element is `presentation`.

```
<div
  onClick="onClickHandler"
  role="presentation">
  <button>Save</button>
</div>
```

Marking an element with the role `presentation` indicates to assistive technology that this element should be ignored; it exists to support the web application and is not meant for humans to interact with directly.

---

There are a couple of things wrong with this:

1. The example implies that a `div` element should have `role="presentation"`. But setting `role="presentation"` on a `div` is redundant. The definition of `role="presentation"` in [WAI-ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/#presentation) is:

    > An element whose implicit native role semantics will not be mapped to the accessibility API.

    Elements like `div` and `span` have no implicit native role semantics in the first place. See Example 8 in that section for further clarification.

1. The paragraph after the example suggests that assistive technology ignores elements with `role="presentation"` entirely, which has never been the case. In fact, this is a common misconception. Per the note in that same section [WAI-ARIA 1.1](https://www.w3.org/TR/wai-aria-1.1/#presentation):

    > Many individuals erroneously consider `role="presentation"` to be synonymous with aria-`hidden="true"`…

You might rewrite the section to explain that the rule is prone to false positives, and that `role="presentation"` is just an escape hatch to make the linter ignore a particular element. However, that advice could lead to unintended consequences. For example, consider the case where someone is using a `ul` element to handle events that bubble up from its descendants:

```jsx
<ul onClick={this.handleItemClick}>
  <li><button>Click me 1</button></li>
  <li><button>Click me 2</button></li>
  <li><button>Click me 3</button></li>
</ul>
```

Adding `role="presentation"` to the `ul` element would remove the list semantics.

So I think you should consider adding a different escape hatch for the event bubble case. There's no appropriate ARIA attribute, because such an element doesn't need any special handling from assistive technology. Maybe you could check for an attribute called something like `data-only-handles-propagated-events`. 🤷‍♂️

n-chardon:
Fixes #510 
Removed mention of the role attribute. Suggestion for fixing the error by disabling the rule for the element.


docs/rules/no-static-element-interactions.md
```diff
@@ -40,19 +40,21 @@ Note: Adding a role to your element does **not** add behavior. When a semantic H
 
 ### Case: This element is not a button, link, menuitem, etc. It is catching bubbled events from elements that it contains
 
-If your element is catching bubbled click or key events from descendant elements, then the proper role for this element is `presentation`.
+If your element is catching bubbled click or key events from descendant elements, there are no appropriate roles for your element : you will have to deactivate the rule. Consider explaining the reason for disabling the rule as well.
```
ljharb:
```suggestion
If your element is catching bubbled click or key events from descendant elements, there are no appropriate roles for your element: you will have to deactivate the rule. Consider explaining the reason for disabling the rule as well.  
```

n-chardon:
Role=none does not exist in the ARIA specification. You are suggesting
using ARIA roles to fix a false positive in a linting tool, and the example
in the current documentation leads to accessibility issues : it is the
reason I took the time to make modifications to the documentation. There
are at least two legitimate use cases I encountered in my job for using a
event handler on a static element : I will provide them quickly.

Le jeu. 10 janv. 2019 à 02:00, J. Renée Beach <notifications@github.com> a
écrit :

n-chardon:
With the ARIA combobox design pattern, there are two subcomponents: the input, and the listbox component. Keyboard interactions are managed on the input, but mouse interactions are managed on the listbox. If you add a wrapper `<div>` with a mouse event handler around the listbox, it will trigger the rule. If you add `role="presentation"` to the wrapper `<div>`, VoiceOver and JAWS will stop reading the listbox, making it impossible for the user to browse options. 

Second example: you want to make a whole  `<div>` clickable, which contains lots of information and lots of markup. To make it usable for screenreader users , you should not wrap  the whole block in a `<button>` tag. Instead, you will create a button inside the block, that will work with event bubbling : no actual event handler is needed for the  `<button>` itself. Being natively focusable, the `<button>` element will also cater for the keyboard users' needs. See a  quick example : https://codepen.io/n-chardon/pen/XoPYPR

jessebeach:
@n-chardon you convinced me. Please update the PR and I'll merge it. Thank you for taking the time to explain your reasoning.

n-chardon:
@jessebeach thanks for taking the time to read me. I have updated the PR, please let me know if there is anything else to do on my side.

docs/rules/no-static-element-interactions.md
```diff
@@ -17,49 +17,53 @@ Indicate the element's role with the `role` attribute:
...
+  /* The <div> element has a child <button> element that allows keyboard interaction */
+}
+{
+  /* eslint-disable-next-line no-static-element-interactions */
+}
+<div onClick={this.handleButtonClick}>
   <button>Save</button>
   <button>Cancel</button>
-</div>
+</div>;
```
ljharb: this semi isn’t needed here

docs/rules/no-static-element-interactions.md
```diff
@@ -17,49 +17,53 @@ Indicate the element's role with the `role` attribute:
...
-If your element is catching bubbled click or key events from descendant elements, there are no appropriate roles for your element : you will have to deactivate the rule. Consider explaining the reason for disabling the rule as well.  
+If your element is catching bubbled click or key events from descendant elements, there are no appropriate roles for your element : you will have to deactivate the rule. Consider explaining the reason for disabling the rule as well.
 
-```
-{/* The <div> element has a child <button> element that allows keyboard interaction */}
-{/* eslint-disable-next-line no-static-element-interactions */}
-<div
-  onClick="onClickHandler">
+```jsx
+{
```
ljharb: please revert these jsx comment changes

jessebeach:
@n-chardon as @ljharb mentioned, please propose the changes without the formatting changes. Sorry for the churn!

docs/rules/no-static-element-interactions.md
n-chardon: Sorry, I am a bit lost here: I added the "jsx" format indicator, because it was added in the upstream branch. Shall I remove it?

ljharb: if you rebase on latest master, it might help clean up the diff

n-chardon: Sorry about the delayed answer. @ljharb Would you like me to clean history of my branch? It might help with readability.

ljharb: @n-chardon please don't delete the branch; that will close the PR - but sure, feel free to reset it to latest master and reapply your changes!

n-chardon:
@ljharb eveything is up to date, and hopefully, much cleaner.

ljharb:
@n-chardon thanks; I rebased it again and reverted a few more unrelated formatting changes.

@jessebeach i think this is ready for your review again :-)

jessebeach:
I updated ESLint to the latest version locally on master and started getting the same errors as the Travis run. We'll get this committed once I get the ESLint issue sorted, which may be in a week because I go on vacation in 2 days I won't be looking at a computer for a week :)

jessebeach:
https://github.com/eslint/eslint/issues/12119

TrevorBurnham:
@n-chardon 👋 I'm just now seeing this. Re:

> If you add `role="presentation"` to the wrapper `<div>`, assistive technology may fail on children elements.

My understanding is that `role="presentation"` is a no-op on a `<div>`. E.g. in the markup

```js
<div role="presentation">
  <button>Click me</button>
</div>
```

the computed role for the button is `"button"` (according to the Accessibility inspector in Chrome). Can you show an example where setting `role="presentation"` on a `<div>` makes a difference?

n-chardon:
@TrevorBurnham you are right: in your example, the button will still be exposed as a button. However, I remember encountering an issue where `role="presentation"` erased the semantics of children. I will get back to you with an example.

PR #539, opened 2019-01-27
(chore) - upgrade and simplify dev dependencies.

evcohen:
### Problem
some dev dependencies were pretty far behind. it's important to stay up to date on these so we can pick up new features & improve performance and security of this package.

### Solution
upgrade ~everything that's outdated. simplify the babel setup by using `babel-preset-airbnb` + an upgrade `strip-flow-types` plugin.

i did not upgrade `flow-bin` because we have some flow errors. this will be tackled in a separate branch.

note: i decided to use `babel-preset-airbnb` to stay in line with our choice for eslint config. we definitely don't need everything it supports for compliation, but it does help us stay up to date with future needs, if necessary.


evcohen:
this breaks node 4 -- going to dig in a little when i get a chance.

ljharb:
No; if anything I’d suggest dropping jest on favor of the only test runner that supports older engines, tape. Alternatively we can just pin jest to the last version that supports node 4.

evcohen:
> No; if anything I’d suggest dropping jest on favor of the only test runner that supports older engines, tape. Alternatively we can just pin jest to the last version that supports node 4.

We'd also have to pin babel, because babel 7 dropped support for node 4. I'm not sure I see an upgrade path if we continue to support unmaintained node versions.

ljharb:
If we're not testing on node 4, then effectively they can't use it on node 4 :-/

The babel 7 part is indeed a problem; it's got all of my babel-using projects stuck on babel 6 for the forseeable future. That, however, can be addressed by *building* on latest node and then *running the tests* on all versions; the test runner can't be bypassed in that way. I think it's incredibly unfortunate and user-hostile that both jest and mocha have dropped support for node < 6, and it's why I now recommend using jest only for apps, and tape for everything else.

jessebeach:
Let's get this update. Not updating is causing us all kinds of grief with parser errors. The ecosystem is moving forward.

PR #541, opened 2019-02-01
Allow aria- props to be null/undefined

TrevorBurnham:
Currently, the [aria-proptypes](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/aria-proptypes.md) rule has the odd property that it won't allow `aria-` props to be reset by setting them to `null` or `undefined`. For example, these are invalid:

```jsx
<Heading aria-level={null} />
<Heading aria-level={undefined} />
```

I'm guessing that this behavior is unintended. There's currently a check for `value === null`, where `value` comes from `getPropLiteralValue()`; but that doesn't work for `null` itself, because `getPropLiteralValue()` [returns the string `"null"`](https://github.com/evcohen/jsx-ast-utils/blob/e4c031233d6bf220bcf15b4c0d80367b2c340e0e/__tests__/src/getPropLiteralValue-test.js#L110-L117) in that case.


jessebeach:
@TrevorBurnham thank you!

PR #542, opened 2019-02-01
Allow role-incompatible props to be null/undefined

TrevorBurnham:
Related to #541. Currently, the [role-supports-aria-props](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/role-supports-aria-props.md) rule won't allow `aria-` props that are unsupported by the element's `role` to be unset with `null` or `undefined`. This PR fixes that.


jessebeach:
@TrevorBurnham thank you!

PR #551, opened 2019-02-07
docs: add proper title to links to axe website for media-has-caption

hozefaj:
while reading the docs for `media-has-caption`, was confused as to why there are two links that point to the same URL. Until you click on its not possible to know they point to different URL's.

This change makes it more obvious. Updated the text for the anchors.


hozefaj:
@ljharb Man, you are quick 👍

PR #558, opened 2019-02-19
Fix typo: defintions > definitions

olitreadwell:
This corrects a minor spelling error I found within the documentation relating to the `aria-role` rule. 

Please let me know if there is more that you need from me to consider this file change. 

Thanks for the work y'all do on this eslint plugin for accessibility 🥇


evcohen:
thanks!

PR #563, opened 2019-03-07
Scaffold RN ruleset by creating directory structure for web and native.

evcohen:


ljharb:
It might be a breaking change to move the existing paths, ftr. Is all that necessary right now for web, or can native just start out in a subdir?

PR #569, opened 2019-03-16
[fix] control-has-associated-label should allow generic links

### LINKED ISSUE #552: control-has-associated-label false positive on link elements
Greetings!

The website I am working on uses Next.js as a framework, and as such all html, including the page header, is defined in JSX. I've found that the rule `control-has-associated-label` is incorrectly flagging link elements in our header as interactive, and expecting labels.

For example:
```jsx
<html lang="en">
  <NextHead>
    ...
    // ERROR: A control must be associated with a text label.
    <link rel="apple-touch-icon-precomposed" sizes="120x120" href="/static/favicon/apple-touch-icon-120x120.png" /> 
    ...
  </NextHead>
  ...
</html>
```

I can get around this by adding `"ignoreElements": ["link"],` to my eslint config for now, but I don't think this should be default behavior.

Thanks for the great plugin btw ✌️

skylarpeterson:
Fixes #552


src/rules/control-has-associated-label.js
```diff
@@ -21,6 +21,8 @@ import mayHaveAccessibleLabel from '../util/mayHaveAccessibleLabel';
 
 const errorMessage = 'A control must be associated with a text label.';
 
+const blacklist = ['link'];
```
ljharb:
```suggestion
const ignoreList = ['link'];
```
although why not make this a Set?

src/rules/control-has-associated-label.js
```diff
@@ -48,13 +50,15 @@ module.exports = {
       ignoreRoles = [],
     } = options;
 
+    const newIgnoreElements = ignoreElements.concat(blacklist);
```
ljharb: then this can be 
```suggestion
    const newIgnoreElements = new Set([...ignoreElements, ...ignoreList]);
```

src/rules/control-has-associated-label.js
```diff
@@ -48,13 +50,15 @@ module.exports = {
...
+
     const rule = (node: JSXElement) => {
       const tag = elementType(node.openingElement);
       const role = getLiteralPropValue(getProp(node.openingElement.attributes, 'role'));
       // Ignore interactive elements that might get their label from a source
       // that cannot be discerned from static analysis, like
       // <label><input />Save</label>
-      if (includes(ignoreElements, tag)) {
+      if (includes(newIgnoreElements, tag)) {
```
ljharb: and this can be
```suggestion
      if (newIgnoreElements.has(tag)) {
```

Issue #399, opened 2018-02-13
Allow object as href/specialLink values 


caub:
We get this error `Cannot read property 'expression' of undefined` as eslint output when using

```js
const a = {};
const foo = () => (
	<Link
		to={{
			pathname: '/bar',
			query: { ...a },
		}}
	/>
);
```

and .eslintrc:
```json
{
	"extends": [
		"eslint:recommended",
	],
	"plugins": ["jsx-a11y"],
	"env": {
		"browser": true,
		"node": true
	},
	"parserOptions": {
		"ecmaVersion": 2018,
		"sourceType": "module",
		"experimentalObjectRestSpread": true,
		"ecmaFeatures": {
			"jsx": true
		}
	},
	"rules": {
		"no-unused-vars": "warn",
		"jsx-a11y/anchor-is-valid": ["warn", {
			"components": ["Link"],
			"specialLink": [ "to" ],
			"aspects": ["noHref", "invalidHref", "preferButton"]
		}]
	}
}
```

https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/9b9745cec0003b2dcd61fe9153c516b75fd5b7a4/src/rules/anchor-is-valid.js#L104

I reported the error in eslint originally: https://github.com/eslint/eslint/issues/9960#issuecomment-365179752

stefanmaric:
I'm also experiencing this.

ljharb:
This is an interesting request; a url is a string, and that’s what this is meant to enforce.

I think if react-router’s Link is sufficiently different from an `<a>`, then it warrants its own react-router-specific rule.

AlmeroSteyn:
@caub Wow sorry somehow missed the @ CC

Yes, as @ljharb mentions, it was not built to cater for objects in the link prop. 

The React Router `Link` component renders to an accessible `<a>` tags and the `to` prop is guarded with a `prop-type` isRequired declaration and, when omitted, it will be reported on by React itself as can be seen on the following console capture:

![image](https://user-images.githubusercontent.com/5063473/39565525-9c9f4178-4eb8-11e8-86aa-bfcd65f6e808.png)

I am not sure if this should be double checked by this linter as when the React error is dealt with, an accessible `<a>` will be rendered.

Thoughts?

justinanastos:
I created a repo repo with this issue before I realized there was already any issue. Here's the link if it helps anyone:

[...]

PR #573, opened 2019-03-16
[#399] Account for spread in parser options

jessebeach:
Use `experimentalObjectRestSpread` in the ESLint test parser.


__tests__/__util__/parserOptionsMapper.js
```diff
@@ -1,6 +1,7 @@
 const defaultParserOptions = {
-  ecmaVersion: 6,
+  ecmaVersion: 2018,
   ecmaFeatures: {
+    experimentalObjectRestSpread: true,
```
ljharb: i believe in 2018+, this ecmaFeature is no longer required because it’s not experimental anymore.

ljharb: ping?

PR #588, opened 2019-05-04
add plugin:jsx-a11y/disabled

justinhelmer:
Addresses https://github.com/airbnb/javascript/issues/2032


__tests__/index-test.js
```diff
@@ -21,9 +21,17 @@ describe('all rule files should be exported by the plugin', () => {
...
 
 describe('configurations', () => {
+  it('should export a \'disabled\' configuration', () => {
+    assert(plugin.configs.disabled);
+  });
+
   it('should export a \'recommended\' configuration', () => {
     assert(plugin.configs.recommended);
   });
+
```
justinhelmer: `strict` was missing from the test so I thought I would backfill it

package.json
```diff
@@ -27,7 +27,7 @@
     "pretest": "npm run lint:fix && npm run flow",
     "test": "npm run jest",
     "test:ci": "npm run jest -- --ci --runInBand",
-    "jest": "jest --coverage __tests__/**/*"
+    "jest": "jest --coverage __tests__/**"
```
justinhelmer: The test path pattern that was used did not include tests from the root of `__tests__/`; only from sub-directories within `__tests/`. For this reason, `__tests__/index-test.js` was not being run. This new pattern includes the `__tests__/` directory, plus all subdirectories, recursively.

scripts/addRuleToIndex.js
```diff
@@ -35,8 +35,13 @@ export default function transformer(file, api, options) {
         ));
         path.parentPath.value.value.properties.sort(nameSort);
       }
-      // Set default reporting to error.
+      // Set default reporting to error (for disabled plugin).
```
justinhelmer: updated `addRuleToIndex` so that new rules created with the `create-rule` script will get added to the `disabled` configuration and turned `off`.

src/index.js
```diff
@@ -39,6 +39,45 @@ module.exports = {
     'tabindex-no-positive': require('./rules/tabindex-no-positive'),
   },
   configs: {
+    disabled: {
```
justinhelmer: The current ruleset

ljharb:
Thanks for the PR!

Closing, however, per https://github.com/airbnb/javascript/issues/2032#issuecomment-489292508

PR #590, opened 2019-05-27
Remove rootDir from Jest path configs

jessebeach:
Jest was failing to exclude the `__util__` directory from test runs.


jessebeach:
> ¯\_(ツ)_/¯

same same

PR #594, opened 2019-05-31
Allow select as a valid child of label.

mikecaulley:
Currently an error is thrown for select within a label.

For example:

```
<label htmlFor="selectInput">
  Favorite ice creams
  <select id="selectInput" />
</label>
```

throws an error but should be okay.


jessebeach:
Thanks @mikecaulley !!

PR #602, opened 2019-06-23
Update ESLint to v6

jessebeach:
V6 just landed yesterday! 

https://eslint.org/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,9 +4,6 @@ node_js:
   - "9"
   - "8"
   - "7"
-  - "6"
-  - "5"
-  - "4"
 cache:
   yarn: true
   directories:
@@ -25,9 +22,7 @@ env:
   global:
     - TEST=true
   matrix:
-    - ESLINT=5
-    - ESLINT=4
-    - ESLINT=3
+    - ESLINT=6
 matrix:
   fast_finish: true
   include:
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -25,26 +22,11 @@ env:
   global:
     - TEST=true
   matrix:
-    - ESLINT=5
-    - ESLINT=4
```
ljharb: If we still claim to support v4 and v5, we need to keep testing on them.

package.json
```diff
@@ -67,7 +68,7 @@
     "jsx-ast-utils": "^2.0.1"
   },
   "peerDependencies": {
-    "eslint": "^3 || ^4 || ^5"
+    "eslint": "^4 || ^5 || ^6"
```
ljharb: Is it necessary to make this a breaking change by dropping v3?

.travis.yml
jessebeach: Good catch.

.travis.yml
```diff
@@ -3,10 +3,6 @@ node_js:
   - "10"
   - "9"
   - "8"
-  - "7"
-  - "6"
-  - "5"
-  - "4"
```
ljharb: eslint 4 works on node 4, doesn’t it?

jessebeach: @ljharb no, we were allowing failures for that combo:

```
- node_js: "4" # not sure why this is failing; probably something in jest
      env: ESLINT=4
```

package.json
jessebeach: Yes, I'm running into issues like this one:

```
https://github.com/airbnb/javascript/issues/1580
```

.travis.yml
ljharb: we were still testing it tho - allowed failures are because it was temporary, and we need to fix it eventually.

package.json
ljharb: we may need to adjust the babel-eslint version to keep it compatible with v3, or we can skip those tests - there’s still the default parser.

PR #603, opened 2019-06-23
Update ESLint to v6

jessebeach:
Let's try just the addition of v6.

(ref #602)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: node_js
 node_js:
+  - "12"
   - "10"
   - "9"
   - "8"
@@ -28,6 +29,7 @@ env:
   global:
     - TEST=true
   matrix:
+    - ESLINT=6
     - ESLINT=5
     - ESLINT=4
     - ESLINT=3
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -40,6 +42,10 @@ matrix:
       env: ESLINT=5
     - node_js: "4"
       env: ESLINT=5
+    - node_js: "4"
+      env: ESLINT=6
```
ljharb: i suspect this won’t work; i think eslint 6 requires node 6 or 8.

package.json
```diff
@@ -36,7 +36,7 @@
     "babel-jest": "^24.0.0",
     "babel-preset-airbnb": "^3.2.0",
     "coveralls": "^3.0.1",
-    "eslint": "^5.15.2",
+    "eslint": "^6.0.0",
```
ljharb: ideally the dev dep range exactly matches the peer dep range

jessebeach: @ljharb I don't know what change you're suggesting that I make. Spell it out for me please :)

ljharb: sorry :-)
```suggestion
    "eslint": "^3 || ^4 || ^5 || ^6",
```

jessebeach: @ljharb  WHOA!!! Today I learned!

ljharb: The benefit is that in CI, `npm ls` will remain valid for all supported versions in the peer dep range (since whenever `npm ls` exits nonzero, the dep graph is invalid and nothing can be expected to work)

jessebeach:
@ljharb ok, I think this is it. Just blocked on https://github.com/airbnb/javascript/issues/2036

ljharb:
@jessebeach try with v2.18.0 of the import plugin.

jessebeach:
That worked!

PR #604, opened 2019-06-23
Only run branch test coverage on the master branch

jessebeach:
We're running Travis twice on each PR. We should only be running Travis on the `master` branch. The PR config will run Travis on PRs.


jessebeach:
I'll revert this if isn't the right thing to do. I think it'll speed up our PR test runs.

ljharb:
nah this is fine. typically one only makes PRs from forks instead of branches directly on the repo, so this doesn't come up.

PR #616, opened 2019-06-30
Reinstate tests

jessebeach:
Some node/eslint combos have begun passing again with the improvements to jsx-ast-utils

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,10 +50,15 @@ matrix:
     - node_js: "6"
       env: ESLINT=6
   allow_failures:
-    - node_js: "9"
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "eslint-plugin-jsx-a11y",
-  "version": "6.2.1",
+  "version": "6.2.2",
```
ljharb: versions shouldn’t be bumped in PRs, generally?

jessebeach: Ack, this was a mistake. I'll rebase.

PR #619, opened 2019-07-01
Add @babel/runtime to the dependencies

jessebeach:
#617 , thanks to @davidje13 for reporting.

Moving this dependency to `dependencies` so it's available at runtime.


jessebeach:
This is pretty clear that we need to include `@babel/runtime` in the runtime dependencies, given require statements in the `lib` files like this:

```
var _interopRequireDefault = require("@babel/runtime/helpers/interopRequireDefault");
```

I'm going to merge this and cut a new patch release.

PR #624, opened 2019-07-18
Add new autocomplete-valid rule using axe-core

WilcoFiers:
This PR introduces the autocomplete-valid rule that was added into axe-core 3.2.0:

https://dequeuniversity.com/rules/axe/3.3/autocomplete-valid/


jessebeach:
@WilcoFiers why can't the eslint rule just call is-valid-autocomplete directly? The function returns a Boolean. I still contend that the trip through an opaque tree structure, these virtual nodes, is a layer of complexity that is unnecessary here. Given a set a scalar values, the rule should returns a Boolean. 

```
const (isValid: boolean) = isValidAutocomplete(attrValue);
```
It's (mostly) the same result without a trip through a blackbox custom tree. These virtual nodes strike me as an artifact of Axe's architecture and a unnecessary complication for the purpose here.

WilcoFiers:
@jessebeach You're right for testing autocomplete attribute itself. `axe.commons.text.isValidAutocomplete(autocompleteValue);` is enough there. But the rule also checks that the element is appropriate for the autocomplete value. Autocomplete doesn't work if you put it on an element that doesn't allow that value. Something like `<input type="date" autocomplete="email" />` (one of the tests) doesn't work because a `type=date` element can never contain an email address.

One thing that we could consider is splitting this over two rules, one that just looks at the value, and the second one to check if the value is appropriate for the element it's on. With axe we didn't do that because they share some code, and it's largely just two different ways to fail at the same thing.

src/rules/autocomplete-valid.js
```diff
@@ -0,0 +1,58 @@
...
+        attrs: node.attributes,
+      });
+
+      const { violations } = runVirtualRule('autocomplete-valid', vNode);
+      if (violations.length === 0) {
+        return;
+      }
+      context.report({
+        node,
+        message: violations[0].nodes[0].all[0].message,
```
ljharb: why only the first violation?

WilcoFiers: `runVirtualRule` returns a standard axe-core data structure, but since we only ran one rule, we only have one violation to return.

WilcoFiers: Put in a comment

jessebeach:
> But the rule also checks that the element is appropriate for the autocomplete value

@WilcoFiers ok then what do you think about about exposing a function like this?

```
const (isValid: boolean) = isValidAutocomplete(attrValue, inputType);
```

The function would have this signature:

```
isValidAutocomplete(attrValue: string, inputType:string) => boolean
```
This avoids the trip through a Virtual Node structure and returns the same information. The lint plugin can get these string values in whatever manner it deems appropriate. Are these two strings values all that are needed to run the rule?

WilcoFiers:
@jessebeach Axe-core doesn't have such a method. We'd either have to add such a method to axe-core, or we'd have to duplicate it for JSX-A11y.

It seems to me there are two differences between what you're suggesting and what I've proposed.

1. The output is a boolean rather than a results object with the message baked into it. You can pull those messages from axe-core. That's what I did for the tests. That's not a problem, although it seems to me things are a little easier if axe-core just gives the messages back. You can get localisation that way too.

2. The call signature of the method. Essentially, the call signature is this:

```js
isValidAutocomplete({
  nodeName,
  attr: (attrName) => ({
    accesskey: attrValue,
    type: inputType
  }[attrName])
}) => AxeResult
```

Rather than this:

```js
isValidAutocomplete(attrValue: string, inputType:string) => boolean
```

The biggest benefit to the first one is that it's consistent. This isn't so much apparent for a one-off rule, but axe-core has dozens of rules. If they didn't have the same call signature, things would get messy pretty quickly.

I could remove the virtual node class and pass the nodeName, autocomplete and type in directly. I'm not sure if your concern is with actually doing an abstraction, or if it's a performance concern. I can run some performance tests to see what the overhead of wrapping this data would be.

jessebeach:
> I could remove the virtual node class and pass the nodeName, autocomplete and type in directly. I'm not sure if your concern is with actually doing an abstraction, or if it's a performance concern. I can run some performance tests to see what the overhead of wrapping this data would be.

Yes! I'm concerned with both abstraction and performance. The Virtual Node introduces a vector for bugs because it creates an intermediate structure. The only purpose for this structure is to satisfy aXe's interface for its rules. I'd rather pass the scalar values that are necessary for evaluation directly to the evaluations and not go through the Node transformation. This will be both less abstract and probably more performant because we're simply doing less work.

jessebeach:
@WilcoFiers I see your point with returning `AxeResult`. Does it return `null` if there is no violation? Or do we need to inspect the object to know the result? And if we need to inspect the object, is there an exported `enum` of statuses to reference or is it a contract we'll need to keep on the plugin side? I quite like a bash approach here, where `null` means no errors and the presence of a message indicates errors.

WilcoFiers:
@jessebeach Thanks for the feedback! That's a little different from what I understood at CSUN, but that is a while ago :-). I'll push an update ASAP. Regarding the violations, it is an empty array when there are none.

WilcoFiers:
Updated :-) Please take a look.

src/rules/autocomplete-valid.js
```diff
@@ -0,0 +1,63 @@
...
+    },
+    schema: [schema],
+  },
+
+  create: context => ({
+    JSXOpeningElement: (node) => {
+      // Determine if ignoreNonDOM is set to true
+      // If true, then do not run rule.
+      const options = context.options[0] || {};
+      const ignoreNonDOM = !!options.ignoreNonDOM;
```
jessebeach: @WilcoFiers, this seems like an unnecessary flag, perhaps a bit confusing, too. What is an example case where `ignoreNonDOM` is set to `false`. What would the JSX be in this case?

src/rules/autocomplete-valid.js
```diff
@@ -0,0 +1,63 @@
...
+      const elType = elementType(node);
+      const isDOMNode = dom.get(elType);
+      if (ignoreNonDOM && !isDOMNode) {
```
jessebeach: This rule should apply to any element or component that is either an `input` or semantically an `input`, right? Rather than asking the implementer to make a global config choice about ignoring non-DOM, you can assert that the element is either semantically an `input` or that it has a role that would make it semantically an `input`, then either bail if both are false or evaluate if either are true.

src/rules/autocomplete-valid.js
WilcoFiers: Thanks for the feedback. My thinking was that by default, if `autocomplete` was used on something that isn't an HTML element, we'd assume it will be passed onto an `input[type=text]` element, and so you'd want to have it tested by default. (i.e. `<MyInput autocomplete="foo" />` should fail by default).

If you used `autocomplete` for some other purpose on a component, you could turn that off with `ignoreNonDom`. That seemed like the way that setting was used in other rules as well. Happy to use something specific to this rule if you think that's better. I could also flip the default, so that it doesn't test `autocomplete` on components and you have to turn it on if you want that.

With regard to what the rule applies to. It applies to all elements (except for `form`) that in HTML are allowed an `autocomplete` attribute. So that's `input` (except for certain types like buttons and hidden), `select`, and `textarea`. If you put `autocomplete` on anything else, axe-core ignores it, since browsers do as well. That includes things like `div[role=textbox]`.

jessebeach: >My thinking was that by default, if autocomplete was used on something that isn't an HTML element, we'd assume it will be passed onto an input[type=text] element, and so you'd want to have it tested by default. (i.e. <MyInput autocomplete="foo" /> should fail by default).

What do you think of simply ignoring non-DOM all the time? If the element isn't an `input`, then just bail out.

src/rules/autocomplete-valid.js
```diff
@@ -0,0 +1,63 @@
...
+
+      const attr = (attrName) => {
+        const value = getLiteralPropValue(getProp(node.attributes, attrName));
+        return (typeof value === 'string' ? value : null);
+      };
+      const props = {
+        nodeName: isDOMNode ? elType : 'input',
+        nodeType: 1,
+      };
+      const hasAttr = attrName => attr(attrName) !== null;
```
jessebeach: Why pass `hasAttr` and `attr` to the aXe call if `hasAttr` also calls `attr`. This is leaking implementation details from aXe. `hasAttr` seems unnecessary.

src/rules/autocomplete-valid.js
```diff
@@ -0,0 +1,63 @@
...
+      // If true, then do not run rule.
+      const options = context.options[0] || {};
+      const ignoreNonDOM = !!options.ignoreNonDOM;
+      const elType = elementType(node);
+      const isDOMNode = dom.get(elType);
+      if (ignoreNonDOM && !isDOMNode) {
+        return;
+      }
+
+      const attr = (attrName) => {
```
jessebeach: The `attr` prop below should simply be a `string` and not nullable. If this rule doesn't have a string value in the attribute, then it won't call aXe. That way, we don't need to leak an implementation detail of aXe to the rule.

src/rules/autocomplete-valid.js
```diff
@@ -0,0 +1,63 @@
...
+        const value = getLiteralPropValue(getProp(node.attributes, attrName));
+        return (typeof value === 'string' ? value : null);
+      };
+      const props = {
```
jessebeach: This is leaking aXe implementation details. The rule shouldn't need to know the shape of this `props` object. The value passed to this argument should just be a string.

src/rules/autocomplete-valid.js
```diff
@@ -0,0 +1,63 @@
...
+        nodeName: isDOMNode ? elType : 'input',
+        nodeType: 1,
+      };
+      const hasAttr = attrName => attr(attrName) !== null;
+
+      const { violations } = runVirtualRule('autocomplete-valid', { attr, props, hasAttr });
```
jessebeach: This should be the interpolated call to aXe:

```
// Pass
const { violations } = runVirtualRule('autocomplete-valid', { tag: 'input', value: 'on'});
// Fail
const { violations } = runVirtualRule('autocomplete-valid', { tag: 'div', value: 'on'});
// Pass, because the rule would just ignore non-DOM elements. Or the plugin wouldn't call aXe in this case if it would fail.
const { violations } = runVirtualRule('autocomplete-valid', { tag: 'MyButton', value: 'on'});
```

Would this work?

src/rules/autocomplete-valid.js
ljharb: fwiw, I agree with that suggestion, except that there needs to be a way to specify an input-like custom component - like `Input` etc.

(Ideally that'd be in settings, for all rules to use, not just per-rule)

WilcoFiers: Sounds good, except for the minor caviat that it also needs to look at `select` and `textarea`. But yeah, let me try to get both of those points in.

src/rules/autocomplete-valid.js
WilcoFiers: Role doesn't work. You can't do `<div role="textbox" autocomplete="given-name">`, or at least, browsers don't support you doing that. `autocomplete` only works on native text fields and on `select`.

src/rules/autocomplete-valid.js
WilcoFiers: @jessebeach That would require an API change to axe-core. The one thing I don't think works here is the `value`. We need two attributes, `type` and `autocomplete` for this rule. I also would want a call signature that can work for other rules as well. How about the following:

```js

const { violations } = runVirtualRule('autocomplete-valid', {
  nodeName: 'input',
  attributes: {
    type: 'text',
    autocomplete: 'given-name'
  }
});
```

jessebeach: @WilcoFiers yes, that works! It keeps the interface simple and declarative.

WilcoFiers: I like it. This is going to require an update to axe-core. Let me see if I can try to convince the powers that be to let me put this in a patch, so that I don't have to wait 2 months for a release. This may take me a little while anyway (busy busy busy!). I'll close the PR for now, and will reopen once I've got an update to axe-core in place. If it's alright with you, I'll include you in the axe-core PR as well Jesse. I want to make sure I get it right this time around.

jessebeach: @WilcoFiers sure thing, include me on the review!

src/rules/autocomplete-valid.js
WilcoFiers: I was totally wrong here. This should be input, plus whatever is on inputComponents, which I'm adding in.

Issue #576, opened 2019-03-19
jsx-a11y/no-noninteractive-tabindex should not flag `tooltip` role


jessebeach:
```
<div
  aria-describedby={labelID}
  role="tooltip"
  tabIndex={0}>
```

A tooltip is meant to be tabbable without being actionable.

jessebeach:
Hmm, here's better guidance. The `tooltip` role is non-interactive and it does not receice focus.

```
Tooltip widgets do not receive focus. A hover that contains focusable elements can be made using a non-modal dialog.
```

https://www.w3.org/TR/wai-aria-practices-1.1/#tooltip

jessebeach:
There's nothing to do here. Tooltips should not receive focus. They are content. At Meta, we associate them with their related element by using aria-describedby on the source element and this points to the tooltip's ID. Only interactive element should receive focus and tooltips are static content. If your tooltip contains interactive content, then it is a dialog and not a tooltip.

PR #640, opened 2019-10-12
fix(no-noninteractive-tabindex): should not flag `tooltip`

jjharriso:
Fixes issue #576  
Adds `tooltip` as allowed role and adds `alwaysValid` code examples for testing.

#hacktoberfest

### DIFF
--- a/__tests__/src/rules/no-noninteractive-tabindex-test.js
+++ b/__tests__/src/rules/no-noninteractive-tabindex-test.js
@@ -35,6 +35,8 @@ const alwaysValid = [
   { code: '<div />' },
   { code: '<div tabIndex="-1" />' },
   { code: '<div role="button" tabIndex="0" />' },
+  { code: '<div role="tooltip" tabIndex="0" />' },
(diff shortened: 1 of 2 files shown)


ljharb:
Same as #641? Let’s keep both open and keep them in sync before merging.

jjharriso:
Thanks @ljharb, yes, they are the same. I made the mistake of PR'ing my fork's master branch, I should have created a branch for this instead.

jessebeach:
@jjharriso My guess here is that you want to be able to tab into the tooltip, right?

In that case, what you really have in a non-modal dialog, not a tooltip. The content of a tooltip should be associated with the triggering element via `aria-describedby` on the triggering element. If focus moves into the tooltip, then it is not a tooltip, it's a dialog.

> Authors SHOULD ensure that elements with the role tooltip are referenced through the use of aria-describedby before or at the time the tooltip is displayed.
https://www.w3.org/TR/wai-aria-1.1/#tooltip

I'm going to reject this change, although I'm open to discussion on the topic.

PR #641, opened 2019-10-12
fix(no-noninteractive-tabindex): should not flag `tooltip`

jjharriso:
Fixes issue #576 
Adds `tooltip` as allowed role and adds `alwaysValid` code examples for testing.

#hacktoberfest


jessebeach:
Duplicate of #640

PR #653, opened 2019-12-02
Update aria-query and axobject-query to the latest versions

jessebeach:
aria-query to version 4.0.1

axobject-query to version 2.1.1


jessebeach:
I need to run this version against the current version on the FB codebase to compare the output.

PR #660, opened 2020-01-29
fix(mouse-event): remove check from custom elements

### LINKED ISSUE #657: jsx-element triggers rule
```
<MyElement onMouseOver={() => {}} />
```
The example above is triggering the rule `jsx-a11y/mouse-events-have-key-events` because it is missing onFocus. But should this rule be applied to custom jsx-elements? I thought it was supposed to only handle native html elements like `div`.

tanmoyopenroot:
#### PR checklist

- [x] Addresses an existing issue: fixes #657 
- [x] bug
- [x] Includes tests
- [ ] Documentation update

#### Overview of change:
Check for a custom element using `dom.get` function from `aria-query` package.


__tests__/src/rules/mouse-events-have-key-events-test.js
```diff
@@ -31,22 +31,35 @@ const mouseOutError = {
 ruleTester.run('mouse-events-have-key-events', rule, {
   valid: [
     { code: '<div onMouseOver={() => void 0} onFocus={() => void 0} />;' },
-    {
-      code: '<div onMouseOver={() => void 0} onFocus={() => void 0} {...props} />;',
-    },
+    { code: '<div onMouseOver={() => void 0} onFocus={() => void 0} {...props} />;' },
```
ljharb: please revert these unrelated formatting changes

tanmoyopenroot: Done

PR #662, opened 2020-01-31
Update focus/hover style for link-button

mdvorscak:
To match the default browser style


jessebeach:
> To match the default browser style

@mdvorscak thank you for the PR. Could you link me to the default stylesheet for Chrome with this value?

jessebeach:
I found the reference: https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/core/html/resources/controls_refresh.css;l=25

jessebeach (approved):
lgtm

PR #665, opened 2020-02-23
Fix autocomplete-valid-test

matijs:
`data` was never passed and `.fail` is not a function.

@WilcoFiers `._audit` looks like it's axe-core internal and it's the only reason `axe-core` is used in `eslint-plugin-jsx-a11y`.  Are there plans to make this part of the public axe-core API?

### DIFF
--- a/__tests__/__util__/axeMapping.js
+++ b/__tests__/__util__/axeMapping.js
@@ -1,6 +1,6 @@
 /* eslint-disable import/prefer-default-export, no-underscore-dangle */
 import * as axe from 'axe-core';
 
-export function axeFailMessage(checkId, data) {
-  return axe._audit.data.checks[checkId].messages.fail(data);
+export function axeFailMessage(checkId) {
+  return axe._audit.data.checks[checkId].messages.fail;
 }


matijs:
Cloning the repo and running `npm install` results in `axe-core` v3.5.1 being installed. Tests fail using this version.

Fixing the tests makes them pass locally but then they fail in CI (Does Travis use a cached version of node_modules somehow?).

Experimenting with manually downgrading `axe-core` has these results:

| branch                             | `axe-core` | result |
| :--------------------------------- | ---------- | ------ |
| master                             | v3.5.1     | fail   |
| master                             | v3.4.2     | fail   |
| master                             | v3.4.1     | pass   |
| matijs:fix/autocomplete-valid-test | v3.5.1     | pass   |
| matijs:fix/autocomplete-valid-test | v3.4.2     | pass   |
| matijs:fix/autocomplete-valid-test | v3.4.1     | fail   |

It looks like [this change](https://github.com/dequelabs/axe-core/commit/a2ddba305003db0540c093f4ad9015c8854cd98d) that ended up in `axe-core` v.3.4.2 is responsible for the tests failing, but since `_audit` suggests it's internal… I'm not sure what to do.

matijs:
Looked into why there was no lockfile and found [this PR](https://github.com/evcohen/eslint-plugin-jsx-a11y/pull/255). @ljharb I was wondering if you still feel the same way.

ljharb:
Always and forever, only apps should have lockfiles.

ljharb:
Observable is public, so this looks like a breaking change in a patch version of axe-core. We can peg the version in package.json to get tests passing again.

matijs:
Closed in favour of #666

PR #670, opened 2020-03-13
Fixes #669: use the `language-tags` package to check the `lang` rule

### LINKED ISSUE #669: js-a11y/lang not recognizing script subtag
According to [the documentation for lang attribute on MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang), a valid HTML lang attribute is composed of three parts: 

language-[script]-[region]

where the script subtag defines the writing system used for the language, and is always 4 characters long, with the first letter capitalized. This subtag is particularly useful as there are two different writing systems in Chinese: simplified (`Hans`) and traditional (`Hant`).

For example, I can have each of the following valid lang attributes:

- `zh-CN`
- `zh-Hant`
- `zh-Hans-CN`

But now the js-a11y/lang rule is rejecting the last two.

Reference:
- [lang - HTML: Hypertext Markup Language | MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang)

imtsuki:
The `language-tags` package provides language tag validation functionality based on the latest IANA language subtag registry.

Instead of manually parsing tags like before, this PR makes use of this package and makes the `lang` rule more exhaustive and more robust. File `src/util/attributes/ISO.json` is removed then.

Signed-off-by: imtsuki <me@qjx.app>


imtsuki:
As there's no response for a while, I'll close this PR.

ljharb:
Let’s please leave it open; there’s no expiration date on PRs and I’d still like to get it in :-)

jessebeach:
> As there's no response for a while, I'll close this PR.

@imtsuki apologies for the delay in response. This is a great change; thank you for proposing it!

package.json
```diff
@@ -68,6 +68,7 @@
     "emoji-regex": "^9.0.0",
     "has": "^1.0.3",
     "jsx-ast-utils": "^2.4.1"
+    "language-tags": "^1.0.5"
```
ljharb, line 71:
```suggestion
    "language-tags": "^1.0.5",
    "jsx-ast-utils": "^2.4.1"
```

PR #671, opened 2020-03-13
Add the jsx-a11y/lang rule to the recommended and strict presets

astorije:
It turns out that `jsx-a11y/lang` is the only rule not enabled in either recommended or strict presets.
I wasn't sure if it was intentional or accidental but I figured it would take me as much time to open an issue asking about it than opening a PR to suggest it. Do what you please with it :)


astorije:
Absolutely agree with that, feel free to keep this PR open until the next major version, I'm in no rush :)

PR #682, opened 2020-04-30
[Refactor] removing a use of `some`

andrelmlins:
I believe that it is not necessary to use the add function. Leaving the whole rule to the Regex.

### DIFF
--- a/src/rules/img-redundant-alt.js
+++ b/src/rules/img-redundant-alt.js
@@ -59,8 +59,7 @@ module.exports = {
       const redundantWords = REDUNDANT_WORDS.concat(words);
 
       if (typeof value === 'string' && isVisible) {
-        const hasRedundancy = redundantWords
-          .some((word) => Boolean(value.match(new RegExp(`(?!{)\\b${word}\\b(?!})`, 'i'))));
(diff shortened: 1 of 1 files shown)


andrelmlins:
Hi @jessebeach , I was implementing accessibility rules for SvelteJs. When comparing my code with this one, I noticed this point.

src/rules/img-redundant-alt.js
```diff
@@ -59,8 +59,7 @@ module.exports = {
       const redundantWords = REDUNDANT_WORDS.concat(words);
 
       if (typeof value === 'string' && isVisible) {
-        const hasRedundancy = redundantWords
-          .some((word) => Boolean(value.match(new RegExp(`(?!{)\\b${word}\\b(?!})`, 'i'))));
+        const hasRedundancy = Boolean(value.match(new RegExp(`(?!{)\\b(${redundantWords.join('|')})\\b(?!})`, 'i')));
```
ljharb, line 62:
```suggestion
        const hasRedundancy = new RegExp(`(?!{)\\b(${redundantWords.join('|')})\\b(?!})`, 'i').test(value);
```

jessebeach:
@ljharb, @andrelmlins made the change. Could you approve and I'll merge it?

PR #685, opened 2020-05-09
Feat/aria 12

jessebeach:


package.json
```diff
@@ -59,7 +59,7 @@
   "license": "MIT",
   "dependencies": {
     "@babel/runtime": "^7.4.5",
-    "aria-query": "^4.0.1",
+    "aria-query": "https://pkg.csb.dev/eps1lon/aria-query/commit/64c91301/aria-query",
```
ljharb, line 62: what is "csb.dev"?

(either way i hope that we'd revert this back to a proper npm dep before merging)

PR #695, opened 2020-06-11
Account for the new `generic` role in ARIA 1.2

jessebeach:
https://www.w3.org/TR/wai-aria-1.2/#generic

>A nameless container element that has no semantic meaning on its own.
>
>Contrast with group, which semantically groups its descendants in a named container.
>
>A generic can provide a limited number of accessible states and properties for its descendants, such as aria-live attributes. This differentiates it from the presentation role.
>
>The generic role is intended for implementors of User Agents. Authors SHOULD NOT use this role in content.

Without this change, `div` and `span` elements are erroneously considered non-interactive elements. A basic assumption in the plugin up to this point is that all roles have meaning, and all roles that don't have interactive semantics, have non-interactive semantics. The `generic` role is a role without meaning, like `void`,

### DIFF
--- a/src/util/isNonInteractiveElement.js
+++ b/src/util/isNonInteractiveElement.js
@@ -26,6 +26,9 @@ const nonInteractiveRoles = new Set(roleKeys
         // 'toolbar' does not descend from widget, but it does support
         // aria-activedescendant, thus in practice we treat it as a widget.
         && name !== 'toolbar'
+        // This role is meant to have no semantic value.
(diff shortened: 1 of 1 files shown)


ljharb:
Might this fix a few open issues?

jessebeach:
> Might this fix a few open issues?

I don't think so. I'm prepping for this change: https://github.com/A11yance/aria-query/pull/46

`generic` isn't supposed to be used by Authors.

PR #696, opened 2020-06-12
Update to ARIA 1.2; fix test failures

### LINKED ISSUE #644: jsx-a11y/aria-props flags aria-details as invalid
`aria-details` is a valid attribute. Debug in what cases it's being flagged as invalid.

jessebeach:
See aria-query's update to ARIA 1.2: https://github.com/A11yance/aria-query/pull/46

- [x] Include the next release of [aria-query](https://github.com/A11yance/aria-query)

### The changes

1. `scrollbar` becomes an interactive element
1. `slider` and `spinbutton` lose some required props
1. `aria-selected `, `aria-expanded`, `aria-invalid` and `aria-haspopup` are no longer allowed on several roles.
1. Fixes #644


__tests__/src/rules/control-has-associated-label-test.js
```diff
@@ -192,7 +192,6 @@ const alwaysValid = [
   { code: '<div role="presentation" />' },
   { code: '<div role="region" />' },
   { code: '<div role="rowgroup" />' },
-  { code: '<div role="scrollbar" />' },
```
ljharb, line 195: is this a breaking change? or would you consider it a bugfix?

jessebeach: @ljharb  Breaking change in behavior, but not API. This will be a Minor release. Some new violations will probably pop up for folks when they run their linter. Does this sound right?

ljharb: semver for eslint rules is tricky; “more errors” is definitely at least a minor, but the real question is whether they have a reasonable expectation that the code shouldn’t be warned on. If this code getting a new warning will help them catch a bug instead of surprising them, then that seems nonmajor?

jessebeach: > If this code getting a new warning will help them catch a bug instead of surprising them, then that seems nonmajor?

Thinking on this more, Authors will most likely be removing overrides put in place because scrollbar wasn't considered an interactive element before now, so putting a click handler on one would have raised a lint warning. That'll be good to note in the release notes.

jessebeach:
In testing the ARIA 1.2 upgrade in `aria-query`, I found a bug that `aria-level` wasn't included as an allowed (and required) prop on the `heading` role.

I found one instance of `The attribute aria-haspopup is not supported by the role link. This role is implicit on the element a`, which is neat.

I found one instance of `Elements with the ARIA role "option" must have the following attributes defined: aria-selected`.

And, `The attribute aria-disabled is not supported by the role listitem. This role is implicit on the element li`

And, `The attribute aria-haspopup is not supported by the role tooltip`

And, `Elements with the ARIA role "checkbox" must have the following attributes defined: aria-checked`

Errors/Warnings were virtually the same over 120,000 files. I'm confident in this upgrade.

ljharb:
woot, ship it (once tests pass)

PR #705, opened 2020-06-19
Fix aria-level error for the heading role (#704)

### LINKED ISSUE #704: Contradicting linting issue with jsx-a11y/role-supports-aria-props
Since the latest update, getting linting issue:

``Elements with the ARIA role "heading" must have the following attributes defined: aria-level  jsx-a11y/role-has-required-aria-props``

After adding the aria-level, getting a contradicting error:

``The attribute aria-level is not supported by the role heading  jsx-a11y/role-supports-aria-props``

jessebeach:
* Fixes #704


jessebeach:
Here's the failing test. I'm updating aria-query now.

PR #708, opened 2020-06-20
Negative tabindex is allowed in aria-activedescendant-has-tabindex rule

### LINKED ISSUE #673: jsx-a11y/aria-activedescendant-has-tabindex rule is incompatible with w3c wai-aria examples
Hi,

I'm trying to implement wai-aria compatible components following w3c wai-aria-practices examples and the following rule causes me problems :
https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/7bcea2006f1d79b399e4aaec82e9cbbbe6ef8cc4/src/rules/aria-activedescendant-has-tabindex.js#L46-L49

This rule doesn't validate w3c [following example](https://www.w3.org/TR/wai-aria-practices/examples/listbox/listbox-collapsible.html) as it uses negative tabIndex 

regards,
Nicolas

jessebeach:
Fixes #673

### DIFF
--- a/src/rules/aria-activedescendant-has-tabindex.js
+++ b/src/rules/aria-activedescendant-has-tabindex.js
@@ -13,7 +13,7 @@ import isInteractiveElement from '../util/isInteractiveElement';
 // Rule Definition
 // ----------------------------------------------------------------------------
 
-const errorMessage = 'An element that manages focus with `aria-activedescendant` must be tabbable';
+const errorMessage = 'An element that manages focus with `aria-activedescendant` must have a tabindex';
 
 const schema = generateObjSchema();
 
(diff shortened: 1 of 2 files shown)


src/rules/aria-activedescendant-has-tabindex.js
```diff
@@ -43,21 +43,17 @@ module.exports = {
...
-          tabIndex === undefined
-          || tabIndex >= 0
-        )
+        && tabIndex === undefined
       ) {
         return;
       }
 
-      if (tabIndex >= 0) {
+      if (tabIndex >= 0 || tabIndex <= -1) {
```
ljharb, line 56: does it need to allow -2 and beyond, or is just -1 sufficient?

jessebeach: Let's just make this `>= -1` to preserve the previous behavior and expand it to -1.

PR #713, opened 2020-06-22
Deprecate the accessible-emoji rule

### LINKED ISSUE #627: The advice from jsx-a11y/accessible-emoji appears to lead to a degraded experience with assistive technologies
From some testing on os x with VoiceOver, it seems the guidance provided in the rule jsx-a11y/accessible-emoji leads to a degraded experience with the label not being announced to the user.

I've setup a codepen which demonstrates this at the following link.

https://codepen.io/anon/pen/ymoBgb

The rule seemed to have been created based on the advice in a blog post from a few years ago now, so perhaps browsers and assistive technologies have advanced since then.

https://github.com/evcohen/eslint-plugin-jsx-a11y/pull/136

jessebeach:
Fixes #627 

@luke-john's codepen reveals how our rule produces a worse experience than the native emoji support.

https://codepen.io/anon/pen/ymoBgb


evcohen:
Thanks 😄

PR #721, opened 2020-06-30
Add failing test for autocomplete with dynamic type

lencioni:
This will cause axe to throw an error like:

> TypeError: Cannot read property 'replace' of null

I believe that the problem might be related to getLiteralPropValue
returning `null` instead of `undefined` for dynamic props.


ljharb:
I pushed up a fix, but this is really a workaround for what's likely a bug in axe caused by using default values and forgetting that that doesn't handle `null`. Would you mind filing an issue upstream for it?

lencioni:
Happy to file a bug upstream, but I'm not sure I understand what you mean by this:
> caused by using default values and forgetting that that doesn't handle `null`

ljharb:
hmm, the stack trace points to https://github.com/dequelabs/axe-core/blob/8a4d9b0970ad909100856faf102f14fea4d6c268/lib/commons/text/sanitize.js - but that would fail on undefined too, so my assumption (which may be incorrect) is that they've got code like `function foo(x = '') { return x.replace(…); }`, which would work for undefined or a string but not for null.

It's really hard to figure out where the error is, though, since they apparently bundle the library :-/

PR #727, opened 2020-07-08
Allow `title` attribute or `aria-label` attribute instead of accessible child in the "anchor-has-content" rule

### LINKED ISSUE #423: anchor-has-content and aria-label / title tags?
JAWS 11+ and ChromeVox has the ability to read aria-label as well as title tags on anchors.  Wouldn't this satisfy the "anchor-has-content" requirement?  Example usecase is a background image for "scroll to top".

Thank you in advance!

gorohash:
Fixes #423 

Allowed `title` and `aria-label` attributes instead of text content in the "anchor-has-content" rule.

### DIFF
--- a/__tests__/src/rules/anchor-has-content-test.js
+++ b/__tests__/src/rules/anchor-has-content-test.js
@@ -36,6 +36,9 @@ ruleTester.run('anchor-has-content', rule, {
       code: '<Link>foo</Link>',
       settings: { 'jsx-a11y': { components: { Link: 'a' } } },
     },
+    { code: '<a title={title} />' },
+    { code: '<a aria-label={ariaLabel} />' },
+    { code: '<a title={title} aria-label={ariaLabel} />' },
   ].map(parserOptionsMapper),
   invalid: [
     { code: '<a />', errors: [expectedError] },
--- a/docs/rules/anchor-has-content.md
+++ b/docs/rules/anchor-has-content.md
@@ -6,6 +6,8 @@
 
 Enforce that anchors have content and that the content is accessible to screen readers. Accessible means that it is not hidden using the `aria-hidden` prop. Refer to the references to learn about why this is important.
 
+Alternatively, you may use the `title` prop or the `aria-label` prop.
+
 ## Rule options
 
 This rule takes one optional object argument of type object:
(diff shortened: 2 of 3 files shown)


jessebeach:
@gorohash are you thinking that using both `title` and `aria-label` provides visual and non-visual label access?

gorohash:
@jessebeach 
I think it provides non-visual label access, but it doesn't necessarily provide enough visual label access.
However, as discussed in #423, icons can provide the meaning visually.

backwardok:
@gorohash Requiring both `title` and `aria-label` seems like you're trying to use the `title` as a hover tooltip for visual affordance, but I don't think that's necessary. I would update this to allow for passing for having a `title` or `aria-label` but not forcing them both to be present.

gorohash:
@backwardok I agree with you. A tooltip shown by `title` is not always the best way to provide visual affordance. I look forward to your update!

backwardok:
Hi @gorohash ! I may have been confusing with my language. I meant I recommend that it be updated to be an OR instead of an AND 🙂  Would you want to make that update to your own code?

gorohash:
@backwardok Thanks for your advice! I updated my code

backwardok (approved):
Thanks for adding in this support!

backwardok:
Hey @jessebeach ! Was wondering if you could review this PR? Thanks! 😄

ljharb:
@gorohash I rebased this, but some tests are failing - there's a chance i did the merge wrong. Mind taking a look before I ping jesse again?

ljharb:
@jessebeach this is freshly rebased; mind giving it a review?

jessebeach:
I'll have a look tonight after the kiddo goes to sleep.

jessebeach (approved):
This is a good improvement. 

In the future, we will want to incorporate the `mayHaveAccessibleLabel` utility and somehow port a consideration for `title` back into that utility, so that we don't have so much logic directly in a rule, where it might diverge from other similar rules.

But for now, this is an improvement to this rule.

jessebeach:
Thank you @backwardok !!

jessebeach:
ack, apologies @ljharb! I hit the Merge button on this one. I know you like to merge the commits in a different way.

ljharb:
No worries, it’s fine :-) thanks for reviewing!

Issue #720, opened 2020-06-29
label-has-associated-control controlComponents should allow regex


travigd:
See the title. :^)

Ideally, I'd like to be able to specify `"controlComponents": "Input$"` to match any component that ends with `Input` (e.g., `<TextInput />`, `<EmailInput />`, `<BirthdateInput />`, etc.).

ljharb:
I'd prefer a glob to a regex string; ie, `*Input`.

travigd:
2 cents: I would vote for regex over glob. Regexes are more powerful and I think they're more common in this kind of thing (I've only ever seen globs used for file paths). How would you express `.+(Input|Area|Box)` using a glob (e.g., if we wanted to allow `TextInput` or `TextBox` or `TextArea` as valid "input" components)? The downside is that regexes are a bit harder to use. ¯\_(ツ)_/¯

ljharb:
"More powerful" is a downside, not an upside.

I'd expect that example to be `["*Input", "*Area", "*Box"]`.

travigd:
You say potato, I say potato. 😉

I don't have super strong feelings either way, but the eslint core rules use regexes over globs.
e.g. https://eslint.org/docs/rules/no-unused-vars

ljharb:
Depends on which rules you look at. 

Accepting a regex string opens us up to a host of potential CVEs id rather avoid.

hallzac2:
I created https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/pull/749 to add glob support 😄

PR #735, opened 2020-08-18
adding glob to label-has-associated-control

lilia-paz:
fixes issue #720 

Request to add glob or regex to controlComponents in label-has-associated-control. Added glob to components

### DIFF
--- a/src/rules/label-has-associated-control.js
+++ b/src/rules/label-has-associated-control.js
@@ -58,12 +58,12 @@ module.exports = {
         return;
       }
       const controlComponents = [
-        'input',
-        'meter',
-        'output',
-        'progress',
-        'select',
-        'textarea',
+        '*Input',
+        '*Meter',
+        '*Output',
+        '*Progress',
+        '*Select',
+        '*Textarea',
       ].concat((options.controlComponents || []));
       // Prevent crazy recursion.
       const recursionDepth = Math.min(


src/rules/label-has-associated-control.js
```diff
@@ -58,12 +58,12 @@ module.exports = {
...
-        'output',
-        'progress',
-        'select',
-        'textarea',
+        '*Input',
+        '*Meter',
+        '*Output',
+        '*Progress',
+        '*Select',
+        '*Textarea',
```
ljharb, line 66: Thanks for the PR! I'm not sure how this actually resolves the issue tho - these are just the defaults (which should not have asterisks). `mayContainChildComponent` is what likely needs to be tweaked to support stars.

lilia-paz: Hi Jordan, thank you for your comments. I'm looking at modifying <mayContainChildComponent> and including the glob package, picoMatch. Does this sound good?

lilia-paz: Hi Jordan, thanks for your comment. I'm going to modify `mayContainChildComponent` and was going to include the package, picoMatch, for globs. Does this work? Thanks again.

ljharb: I'd prefer using minimatch, but sounds great thanks

jessebeach:
Thank you for the PR @lilia-paz . Please reopen this PR when you plan to continue working on it.

Issue #743, opened 2020-09-17
fieldset should allow role="radiogroup"


wjmao88:
I've encountered an error trying to put `radiogroup` as a role on a `fieldset`
![image](https://user-images.githubusercontent.com/5618503/93527683-3bb49880-f907-11ea-820d-6b4b10edc637.png)

but according to MDN, radiogroup is a permitted role on `fieldset` https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset

![image](https://user-images.githubusercontent.com/5618503/93527748-525aef80-f907-11ea-8167-e8f68e956e74.png)

hallzac2:
I created https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/pull/746 to address this 😄

PR #746, opened 2020-10-05
Update recommended config to allow fieldset to have the radiogroup role

hallzac2:
###  Description
Update the recommended config to allow `fieldset` to have the `radiogroup` role. Also update the test for this rule to have an entry for this change and one for the `presentation` role.

[MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset) lists both `radiogroup` and `presentation` as valid options for the role attribute on a `fieldset`.

This fixes https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/743.

### Testing
1. Pull this branch (or apply these changes as a patch)
2. Run `npm i`
3. Run `npm run test`, no tests should fail


hallzac2:
I only added this to the `recommended` config for now. Is this something that should be added to the `strict` config as well?

ljharb (approved):

jessebeach (approved):
Looks good, thank you for the PR!

PR #754, opened 2020-10-26
Upgrade jsx-ast-utils dependency to v3.1.0

### LINKED ISSUE #747: Upgrade jsx-ast-utils to support Optional Chaining (tracking issue)
I opened an issue upstream on jsx-ast-utils: https://github.com/jsx-eslint/jsx-ast-utils/issues/103 . The fix will have to occur there, and then this package should be updated to depend on (or at least accept in the semver range) the new version.

## Symptom:
Began seeing `The prop value with an expression type of ChainExpression could not be resolved. Please file issue to get this fixed immediately.` in eslint output after upgrading.

ECMAscript Optional Chaining shipped, resulting in various changes to parsers and javascript/typescript tools to support it. Initially there were different, non-standard names for the AST nodes to represent it. Over this year, consensus has settled out, including in the parser that eslint uses, estree (see https://github.com/estree/estree/pull/204 ).

## Versions:
eslint@7.10.0
jsx-ast-utils@2.4.1 (via eslint-plugin-react@7.21.3, eslint-plugin-jsx-a11y@6.3.1)

## Minimal test case
Component.jsx
```
import React from 'react';

export function Component() {
  const x = {};
  return (
    <div role="button" key={x.y?.z}>
      hi
    </div>
  );
}

```

.eslintrc
```
module.exports = {
  extends: [
    'plugin:react/recommended',
    'plugin:jsx-a11y/recommended'
  ],
  plugins: ['react', 'jsx-a11y'],
}
```

run
```
eslint ./Component.jsx
```

jessebeach:
I totally missed that jsx-ast-utils had a minor version bump.

Fixes #747


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "eslint-plugin-jsx-a11y",
-  "version": "6.4.0",
+  "version": "6.5.0",
```
ljharb, line 3: i don't think this bump is necessary - 3.1 is already included with the semver range, so this is just a patch.

ljharb (changes requested):

ljharb (approved):

PR #757, opened 2020-11-17
Remove no-onchange from recommended and strict

### LINKED ISSUE #398: Is no-onchange still a relevant rule?
Hey, all. Like the title says, is this still a necessary rule? `onchange` is a pretty ubiquitous technique. [React](https://reactjs.org/docs/forms.html#the-select-tag) suggests using it for `<select>` elements to set state. Using `onblur` instead of `onchange` in this case means a user could potentially select an option with their keyboard and submit the form without ever losing focus, and the state wouldn't be updated. 

Of course, not everyone is using React and I could just ignore the rule, but I wanted to throw this question out. [W3C](https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/SCR19.html) even has a suggested technique for their Web Content Accessibility Guidelines 2.0 that shows how to use `onchange` with a `<select>` element.

If this is still relevant, would we be able to get some more current references in the [docs](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-onchange.md) that support this rule? Of the two current citations, one is from 2004 and the other is from 2005. Considering this is an eslint plugin, it might be nice to have documentation that's not from a time when IE6 dominated the web 😝

karlhorky:
**TODO**

- [x] Determine which browsers exhibit behavior tested by @diverent2 in https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/398#issuecomment-685442841

See https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/398#issuecomment-729113622

Closes #398

### DIFF
--- a/docs/rules/no-onchange.md
+++ b/docs/rules/no-onchange.md
@@ -1,4 +1,6 @@
-# no-onchange
+# [Deprecated] no-onchange
+
+⚠️ **Deprecated:** This rule is based on reports of behavior of [old browsers (eg. IE 10 and below)](https://www.quirksmode.org/dom/events/change.html#t05). In the meantime, this behavior has been corrected, both in newer versions of browsers as well as [in the DOM spec](https://bugzilla.mozilla.org/show_bug.cgi?id=969068#c2).
 
 Enforce usage of `onBlur` over/in parallel with `onChange` on select menu elements for accessibility. `onBlur` **should** be used instead of `onChange`, unless absolutely necessary and it causes no negative consequences for keyboard only or screen reader users. `onBlur` is a more declarative action by the user: for instance in a dropdown, using the arrow keys to toggle between options will trigger the `onChange` event in some browsers. Regardless, when a change of context results from an `onBlur` event or an `onChange` event, the user should be notified of the change unless it occurs below the currently focused element.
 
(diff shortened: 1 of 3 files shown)


ljharb (changes requested):
Before landing this, I'd like to understand which browsers it's relevant for, so we can note it in the documentation.

karlhorky:
QuirksMode to the rescue!

Screen Shot 2020-11-17 at 20 58 44

> When you focus on a select box and go through its options with the keyboard, IE and Opera fire the change event every time you select a new option. The other browsers fire the event only when you confirm the new option by leaving the select box.

So it looks like the bug was there until at least IE10 pr2.

@ljharb does that address the question of compatibility?

https://www.quirksmode.org/dom/events/change.html#t05

karlhorky:
Also interesting to note is that it was actually written into the DOM spec a long time ago this way: https://bugzilla.mozilla.org/show_bug.cgi?id=969068#c2

karlhorky:
Ok, I've added a deprecation warning following the other PRs such as #713, along with an explanation. Is it acceptable like this?

karlhorky:
@ljharb What do you think about my new changes? Is this good to get merged?

ljharb (approved):

karlhorky:
Thanks for the review and merge @ljharb !

Issue #402, opened 2018-02-25
Add option for passed hrefs (as in Next.js) (anchor-is-valid)


sk22:
I'd suggest adding an option for `a` tags that don't have an `href` because its containing component passes the `href` to it.

For example, Next.js' Link component takes the `href` but [requires an `a` tag as a child](https://github.com/zeit/next.js#with-link):

```javascript
<Link href="/about">
  <a>About</a>
</Link>
```

Creating such structure triggers the anchor-is-valid rule:

```
[eslint] The href attribute is required on an anchor. Provide a valid, navigable address as the href value. (jsx-a11y/anchor-is-valid)
```

At the moment, I don't see any practical usage of the anchor-is-valid rule that allows passing the href to the child.

A possible fix could be to check if the parent component (i.e. Link) has the `passHref` prop set (which is [optional](https://github.com/zeit/next.js#using-a-component-that-supports-onclick) in Next.js when the child is an `a` tag without an own `href`).
Other than that, the rule could just check for the `<Link href><a></Link>` structure.

sk22:
Here's a workaround (i.e., disabling the rule's `noHref` aspect):

```json
"jsx-a11y/anchor-is-valid": [ "error", {
  "components": [ "Link" ],
  "specialLink": [ "hrefLeft", "hrefRight" ],
  "aspects": [ "invalidHref", "preferButton" ]
}]
```

ljharb:
That sounds like a really strange API; if it’s a Link why would it need an `a` tag to be passed in? All that does is force a cloneElement, needlessly

sk22:
It's indeed a strange API. The reason they did it like this is probably so the Link element can be rendered as any, not only an anchor. But there's an `as` prop on Link, I'll just check out if this can be used instead.

sk22:
Btw; yep, what you said is literally what the Next.js Link does... https://github.com/zeit/next.js/blob/canary/lib/link.js#L149

ljharb:
I’m not sure why they wouldn’t accept a component (or string “a”) so they can render the element directly rather than cloning :-/

Either way I’m not sure that supporting this generically is something either possible or reasonable to do.

sk22:
I see... Thanks!

ljharb:
I’ll leave this open if somebody has an idea of how to handle it, but hardcoding an arbitrary library’s strange convention is a nonstarter.

trevordmiller:
[...]

PR #769, opened 2021-02-03
Add Next.js case to anchor-is-valid rule docs

zackdotcomputer:
Adds up-front documentation of the issue and possible workarounds for users encountering #402


ljharb (approved):
Thanks, this is great!

PR #778, opened 2021-03-02
[Docs] Document the similarity between html-has-lang and lang

### LINKED ISSUE #777: html-has-lang rule seems redundant with lang
The rule [html-has-lang](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/html-has-lang.md) is fully superseded by the [lang](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/lang.md) rule. Would we consider marking the former as deprecated in favor of the latter one?

As I was reviewing the list of rules, I got confused and had to figure out whether I misunderstood something. Having one marked as deprecated would really help. I don’t see any particular reason for html-has-lang to exist. Alternatively, if there is an important case that I miss, we could add explanation to the docs as to the similarity of the two.

Related: #750.

denis-sokolov:
The simplest way to resolve #777.

### DIFF
--- a/docs/rules/html-has-lang.md
+++ b/docs/rules/html-has-lang.md
@@ -1,6 +1,6 @@
 # html-has-lang
 
-<html> elements must have the lang prop.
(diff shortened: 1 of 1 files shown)


jessebeach (approved):
Thank you @denis-sokolov !

PR #779, opened 2021-03-02
[Docs] Simplify the differences table in main doc

denis-sokolov:
The heading of the section says “difference”, yet most of the rows have no difference, leading to a confusion.

Alternatively, I would suggest renaming the section to “Exhaustive detail of recommended and strict configurations”.

Related: #238.


jessebeach:
> Alternatively, I would suggest renaming the section to “Exhaustive detail of recommended and strict configurations”.

@denis-sokolov this is a workable suggestion. Could you PR this?

denis-sokolov:
@jessebeach, sure! #786

PR #782, opened 2021-03-25
Fully deprecate accessible-emoji rule

backwardok:
The [accessible-emoji rule got marked as deprecated](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/pull/713) to address https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/627, but was never removed from the rule set.

This removes the rule from the strict and recommended lists and also updates the readme to reflect it being deprecated.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -50,7 +50,6 @@ module.exports = {
         },
       },
       rules: {
-        'jsx-a11y/accessible-emoji': 'error',
         'jsx-a11y/alt-text': 'error',
         'jsx-a11y/anchor-has-content': 'error',
         'jsx-a11y/anchor-is-valid': 'error',
@@ -206,7 +205,6 @@ module.exports = {
         },
       },
       rules: {
-        'jsx-a11y/accessible-emoji': 'error',
         'jsx-a11y/alt-text': 'error',
         'jsx-a11y/anchor-has-content': 'error',
         'jsx-a11y/anchor-is-valid': 'error',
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -50,7 +50,7 @@ module.exports = {
         },
       },
       rules: {
-        'jsx-a11y/accessible-emoji': 'error',
+        'jsx-a11y/accessible-emoji': 'off',
```
ljharb, line 53: any reason to keep the line at all, given that it provides no config?
```suggestion
```

src/index.js
```diff
@@ -206,7 +206,7 @@ module.exports = {
...
```
ljharb, line 209: same here?
```suggestion
```

ljharb (approved):

src/index.js
backwardok: Wasn't sure if it was something that should be there or not! Happy to remove them :)

backwardok:
@ljharb is there a way to retry a test or do I have to rebase and push again to trigger the tests to rerun?

ljharb:
@backwardok i think only a maintainer can retry one; otherwise the typical way is to close and reopen the PR. In this case, the task is failing because when rebased, there's merge conflicts. Feel free to ignore this tho; we can rebase it for you when it's ready to land.

jessebeach:
Hi @backwardok ! Could you rebase this and re-push it? I'd love to get it merged.

backwardok:
Thanks @ljharb !

PR #798, opened 2021-06-04
Update anchor-is-valid.md

sosualfred:


docs/rules/anchor-is-valid.md
```diff
@@ -46,7 +46,7 @@ Valid `href` attributes values are:
 <a href="#top">Navigate to internal page location</a>
 ```
 
-### Case: I need the HTML to be interactive, don't I need to use an a tag for that?
+### Case: I need the HTML to be interactive, don't I need to use a tag for that?
```
ljharb, line 49: this is correct as is. it's "an anchor tag", not "a tag".

jessebeach:
`a` is a noun here, not an indefinite article. Oh English ;p

PR #807, opened 2021-07-18
fix: remove incorrect implicit role from dl

chrisgrabinski:
## Description

When assigning a `role` attribute to `dl`, eslint throws the following error:

```
The element dl has an implicit role of list. Defining this explicitly is redundant and should be avoided.
jsx-a11y/no-redundant-roles
```

However, based on documentation from W3 and MDN, while the `dl` element accepts a role of `list`, it's **not implicit**.

## Solution

- Remove the code that states that `dl` has an implicit role of `list`.
- Remove test for `role-supports-aria-props`

## Resources

- **W3:** [Document conformance requirements for use of ARIA attributes in HTML](https://www.w3.org/TR/html-aria/#docconformance)
- **MDN:** [\<dl\>: The Description List element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl)

### DIFF
--- a/__tests__/src/rules/role-supports-aria-props-test.js
+++ b/__tests__/src/rules/role-supports-aria-props-test.js
@@ -425,10 +425,6 @@ ruleTester.run('role-supports-aria-props', rule, {
       errors: [errorMessage('aria-expanded', 'dialog', 'dialog', true)],
     },
     {
-      code: '<dl aria-expanded />',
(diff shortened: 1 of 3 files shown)


ljharb:
In the intervening time, this commit's changes seem to have already landed on main. Sorry for the delay, and I hope you'll still contribute again!

PR #808, opened 2021-07-20
fix: tabIndex="0" -> tabIndex={0}

DoctorDerek:
```
Type 'string' is not assignable to type 'number | undefined'.ts(2322)
index.d.ts(1835, 9): The expected type comes from property 'tabIndex' which is declared here on type 'DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, HTMLDivElement>'
(JSX attribute) HTMLAttributes<HTMLDivElement>.tabIndex?: number | undefined
```

### DIFF
--- a/docs/rules/no-static-element-interactions.md
+++ b/docs/rules/no-static-element-interactions.md
@@ -22,6 +22,7 @@ Indicate the element's role with the `role` attribute:
   onKeyPress={onKeyPressHandler}
   role="button"
   tabindex="0">
+  tabindex={0}>
   Save
 </div>
 ```


ljharb:
This is a TypeScript error, and in fact all HTML attributes are all strings - so React will turn the number into a string when rendering.

It's not wrong to make this change, but it's not wrong to leave it alone either.

DoctorDerek:
Good point @ljharb , and in fact build tools will error on the opposite case, e.g.
```
preload={true} // OMG A BOOLEAN DIDN'T YOU MEAN ... STRING!!!?????? 🧵🧵🧵🧵
```
vs.
```
preload="true"
```

One would think, though, that even though `tabIndex` gets coerced to a string, it should actually be numeric.

To quote MDN:
> The tabindex global attribute indicates that its element can be focused, and where it participates in sequential keyboard navigation (usually with the Tab key, hence the name).
> It accepts an integer as a value, with different results depending on the integer's value:
src: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

DoctorDerek:
> That it accepts an integer doesn't change that it's an integer in string form. TS can't type "integer" anyways.

Absolutely, `tabIndex={0.1}` is way more wrong than `tabIndex="0"` -- as the latter is not only correct, it's strictly correct.

Nevertheless, it's jarring to be following the recommended ESLint example and get another error. Thus the PR.

I doubt the TypeScript people are going to be fans of changing their type to `string` since TypeScript can't type `integer`... 😀

ljharb:
They can type `number`, though - i think that's worth filing an issue or a PR on TS itself.

PR #814, opened 2021-10-02
Upgrade aria-query and axobject-query libraries, with changes to remove Map and Set usage

jessebeach:
Closes #875.

### DIFF
--- a/package.json
+++ b/package.json
@@ -69,11 +69,11 @@
   "license": "MIT",
   "dependencies": {
     "@babel/runtime": "^7.19.0",
-    "aria-query": "^4.2.2",
+    "aria-query": "^5.1.1",
     "array-includes": "^3.1.5",
     "ast-types-flow": "^0.0.7",
     "axe-core": "^4.4.3",
-    "axobject-query": "^2.2.0",
+    "axobject-query": "^3.1.1",
     "damerau-levenshtein": "^1.0.8",
     "emoji-regex": "^9.2.2",
     "has": "^1.0.3",
--- a/__tests__/src/rules/role-supports-aria-props-test.js
+++ b/__tests__/src/rules/role-supports-aria-props-test.js
@@ -94,6 +94,7 @@ ruleTester.run('role-supports-aria-props', rule, {
     { code: '<a href="#" aria-disabled />' },
     { code: '<a href="#" aria-dropeffect />' },
     { code: '<a href="#" aria-flowto />' },
+    { code: '<a href="#" aria-haspopup />' },
     { code: '<a href="#" aria-grabbed />' },
     { code: '<a href="#" aria-hidden />' },
     { code: '<a href="#" aria-label />' },
@@ -115,6 +116,7 @@ ruleTester.run('role-supports-aria-props', rule, {
     { code: '<area href="#" aria-dropeffect />' },
     { code: '<area href="#" aria-flowto />' },
     { code: '<area href="#" aria-grabbed />' },
+    { code: '<area href="#" aria-haspopup />' },
     { code: '<area href="#" aria-hidden />' },
     { code: '<area href="#" aria-label />' },
     { code: '<area href="#" aria-labelledby />' },
@@ -136,6 +138,7 @@ ruleTester.run('role-supports-aria-props', rule, {
     { code: '<link href="#" aria-flowto />' },
     { code: '<link href="#" aria-grabbed />' },
     { code: '<link href="#" aria-hidden />' },
+    { code: '<link href="#" aria-haspopup />' },
     { code: '<link href="#" aria-label />' },
     { code: '<link href="#" aria-labelledby />' },
     { code: '<link href="#" aria-live />' },
@@ -553,20 +556,8 @@ ruleTester.run('role-supports-aria-props', rule, {
       errors: [errorMessage('aria-invalid', 'link', 'link', true)],
     },
     {
-      code: '<link href="#" aria-haspopup />',
-      errors: [errorMessage('aria-haspopup', 'link', 'link', true)],
-    },
-    {
       code: '<area href="#" aria-invalid />',
       errors: [errorMessage('aria-invalid', 'link', 'area', true)],
-    },
-    {
-      code: '<area href="#" aria-haspopup />',
-      errors: [errorMessage('aria-haspopup', 'link', 'area', true)],
-    },
-    {
-      code: '<a href="#" aria-haspopup />',
-      errors: [errorMessage('aria-haspopup', 'link', 'a', true)],
     },
     {
       code: '<a href="#" aria-invalid />',


__tests__/src/rules/aria-props-test.js
```diff
@@ -19,7 +19,7 @@ import getSuggestion from '../../../src/util/getSuggestion';
 // -----------------------------------------------------------------------------
 
 const ruleTester = new RuleTester();
-const ariaAttributes = [...aria.keys()];
+const ariaAttributes = aria.keys();
```
ljharb, line 22: still seems like we should leave these as they are?

src/util/isSemanticRoleElement.js
```diff
@@ -38,25 +39,21 @@ const isSemanticRoleElement = (
...
-          roles.forEach((role) => {
-            if (res === true) {
-              return;
-            }
+          for (let m = 0; m < roles.length; m += 1) {
+            const role = roles[m];
             if (role.name === roleAttrValue) {
-              res = true;
+              return true;
             }
```
ljharb, line 50: this seems like it'd be better as a `.some` than a for loop?

ljharb:
Marked as draft, since the dep updates aren't published yet.

__tests__/src/rules/aria-props-test.js
jessebeach: `Map.prototype.keys` returns an `Iterator` and now this new method returns an `Array`. So we don't need to "transform" an iterator into an array via the spread hand-wave. This is the breaking in the breaking change :/

ljharb: right - but it's not really breaking, because we *could* still spread the array, since arrays are iterable just like iterators are (spread syntax works on iterables, not iterators; it's just that builtin iterators are iterable by convention)

jessebeach:
I'm getting back to this this week.

src/util/isSemanticRoleElement.js
```diff
@@ -11,12 +11,13 @@ const isSemanticRoleElement = (
...
-  elementAXObjects.forEach((axObjects, concept) => {
-    if (res) {
-      return;
-    }
+  const entries = elementAXObjects.entries();
+  for (let i = 0; i < entries.length; i += 1) {
+    const [
+      concept,
+      axObjects,
+    ] = entries[i];
```
ljharb, line 20: any reason this can't use `elementAXObjects.forEach`?

src/util/isSemanticRoleElement.js
```diff
@@ -38,25 +39,17 @@ const isSemanticRoleElement = (
...
+        )) {
+          return true;
         }
-      });
+      }
     }
-  });
-  return res;
+  }
+  return false;
```
ljharb, line 52: this kind of seems like `axObjects.some` would work?

__tests__/src/rules/aria-props-test.js
jessebeach: It's been reverted to the original.

src/util/isSemanticRoleElement.js
jessebeach: It's easier to just return from the for loop, rather than creating a closure to capture the result of the loop in a forEach.

src/util/isSemanticRoleElement.js
jessebeach: Updated to use `.some`.

src/util/isSemanticRoleElement.js
```diff
@@ -38,25 +39,18 @@ const isSemanticRoleElement = (
...
-            }
-          });
+      return axObjects.some((obj) => {
+        const roles = AXObjectRoles.get(obj);
+        if (roles && roles.some(
+          (role) => role.name === roleAttrValue,
+        )) {
+          return true;
         }
+        return false;
```
ljharb, line 49:
```suggestion
        return roles && roles.some((role) => role.name === roleAttrValue);
```

src/util/isNonInteractiveElement.js
```diff
@@ -78,24 +80,28 @@ const interactiveElementRoleSchemas = elementRoleEntries
...
+const nonInteractiveElementAXObjectSchemas = elementAXObjects.entries()
   .reduce((
     accumulator,
     [
       elementSchema,
       AXObjectSet,
     ],
   ) => {
-    if ([...AXObjectSet].every((role): boolean => nonInteractiveAXObjects.has(role))) {
+    if (AXObjectSet.every((role): boolean => nonInteractiveAXObjects.findIndex(
```
ljharb, line 102: this looks like typescript or flow, but this is a JS file. how is this parsing? (same question in lots of places)

styfle: Looks like Flow as seen at the top of the file https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/18266282424c546f47a74f842a7bf62bdde1d5e8/src/util/isInteractiveElement.js#L1-L3

And also in package.json https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/18266282424c546f47a74f842a7bf62bdde1d5e8/package.json#L22

Flow uses `.js` file extension.

ljharb: ah right.

we should really move this project over to TS; flow's basically dead outside of FB.

matthieujabbour:
@jessebeach thanks for your work! Any chance this PR will be merged in the coming weeks?

jessebeach:
I'll have time in September while I'm on parental leave to get this landed.

__tests__/src/rules/aria-role-test.js
```diff
@@ -23,7 +23,7 @@ const errorMessage = {
   type: 'JSXAttribute',
 };
 
-const roleKeys = [...roles.keys()];
+const roleKeys = roles.keys();
 
 const validRoles = roleKeys.filter((role) => roles.get(role).abstract === false);
```
ljharb, line 28: `.keys()` returns an iterator - why would an iterator have `.filter` or `.map` etc on them (at least prior to iterator helpers hitting stage 3 and being shipped or polyfilled)?

jessebeach: @ljharb apologies here, I'm in the middle updating this PR and I shouldn't have pushed it to Github yet. I'll take the feedback into account!

jessebeach: @ljharb maybe I flubbed this up. `keys` isn't the built-in method for Set or Map. It's just a method I defined on an interface to the objects in this library and it returns an array, for example: https://github.com/A11yance/aria-query/blob/main/src/domMap.js#L420

What do you think, too confusing to use the same method name as a similar built-in? I was trying to simplify the interface to aria-query and axobject-query while keeping them relatively similar to the Sets/Maps that were used before (but that required the babel runtime).

ljharb: Yes, I think that's pretty confusing - you could make them return `array.keys()` tho, and then it'd be an iterator?

ljharb: (if you don't want to depend on an engine implementing that, you could use https://npmjs.com/array.prototype.keys)

jessebeach: > Yes, I think that's pretty confusing - you could make them return `array.keys()` tho, and then it'd be an iterator?

ya, that makes sense. That means a major release of both of those dependencies though :) Alright, I'll get on that. Good news is it should me no major upgrades for this project. Bad news is anyone who already adjusted to the v5 release. But them's the brakes in software.

ljharb: why does it mean a major release?

ljharb: oh, i suppose if someone was depending on it being an array, yeah :-/

.eslintrc
```diff
@@ -1,13 +1,8 @@
 {
-  "root": true,
```
ljharb, line 2: this definitely needs to stay here, otherwise eslint configs in higher directories can pollute things.

.gitignore
```diff
@@ -8,5 +8,3 @@ npm-shrinkwrap.json
 package-lock.json
 yarn-error.log
 yarn.lock
-
-.npmignore
```
ljharb, line 12: it seems like maybe this PR needs a rebase :-)

jessebeach: > it seems like maybe this PR needs a rebase :-)

Indeed. I put it into draft mode because I'm prepping it for the release of the new minor versions to `aria-query` and `axobject-query`. I had to brush off my git revert/rebase/restore knowledge to figure out how to squash all the changes, get them back to main, then apply just the changes I'll need to support the upcoming releases of the two dependencies. That's all done now. I'll be releasing the updated dependency versions soon and then I'll set this one back to active for review.

.eslintrc
jessebeach: Yes, I just pushed changes with a rebase. This is in draft. Please don't bother reviewing until I set it back for review. Thank you for the quick attention all the same!

package.json
```diff
@@ -69,11 +69,11 @@
   "license": "MIT",
   "dependencies": {
     "@babel/runtime": "^7.19.0",
-    "aria-query": "^4.2.2",
+    "aria-query": "^5.0.0",
```
ljharb, line 72: hm, interestingly both v4 and v5 claim to require node 6, but this package requires node 4.

what in aria-query requires node 6?

package.json
```diff
@@ -69,11 +69,11 @@
...
     "array-includes": "^3.1.5",
     "ast-types-flow": "^0.0.7",
     "axe-core": "^4.4.3",
-    "axobject-query": "^2.2.0",
+    "axobject-query": "^3.0.1",
```
ljharb, line 76: v2 has no engines field, but v3 requires node 6, so - same question as above

package.json
jessebeach: Nothing, it's a superfluous line. Good catch, I'll remove it.

package.json
jessebeach: Same as above.

jessebeach:
Well, it only took a year, but here we go, ready to commit :) 

@ljharb for the final review.

ljharb (approved):

PR #827, opened 2021-11-27
test: align usage of jest expect across tests

armano2:


package.json
```diff
@@ -45,7 +46,6 @@
     "eslint-plugin-flowtype": "^5.8.0 || ^8.0.2",
     "eslint-plugin-import": "^2.25.3",
     "estraverse": "^5.3.0",
-    "expect": "^24.9.0",
```
ljharb, line 48: Using the package and the explicit imports are quite intentional; being more coupled to jest is undesirable.

__tests__/src/util/isNonInteractiveRole-test.js
```diff
@@ -1,5 +1,4 @@
-/* eslint-env mocha */
-import expect from 'expect';
+/* eslint-env jest */
```
ljharb, line 1: This should really be done with overrides in eslintrc, not in a comment per file.

ljharb (changes requested):

package.json
armano2: ok, previously `expect` was not always imported, now it should be consistent

__tests__/src/util/isNonInteractiveRole-test.js
armano2: done

jessebeach:
@ljharb looks like your requested changes were made. Shall we get this landed?

jessebeach (approved):
lgtm, thanks @armano2 !!

ljharb (approved):

PR #828, opened 2021-12-02
Added allowedInvalidRoles option to aria-role

### LINKED ISSUE #574: role="text" should be valid
<div role="text"></div>

Throws `error  Elements with ARIA roles must use a valid, non-abstract ARIA role  jsx-a11y/aria-role`

It's a valid role plz see: https://axesslab.com/text-splitting/

JoshuaKGoldberg:
Adds an optional `allowedInvalidRoles: string[]` option for an allowlist of roles to no longer consider invalid.

Fixes #574.

### DIFF
--- a/docs/rules/aria-role.md
+++ b/docs/rules/aria-role.md
@@ -10,11 +10,14 @@ This rule takes one optional object argument of type object:
 {
     "rules": {
         "jsx-a11y/aria-role": [ 2, {
+            "allowedInvalidRoles": ["text"],
             "ignoreNonDOM": true
         }],
     }
 }
 ```
+
+`allowedInvalidRules` is an optional string array of custom roles that should be allowed in addition to the ARIA spec, such as for cases when you [need to use a non-standard role](https://axesslab.com/text-splitting).
 
 For the `ignoreNonDOM` option, this determines if developer created components are checked.
 
--- a/src/rules/aria-role.js
+++ b/src/rules/aria-role.js
@@ -14,6 +14,13 @@ import { generateObjSchema } from '../util/schemas';
 const errorMessage = 'Elements with ARIA roles must use a valid, non-abstract ARIA role.';
 
 const schema = generateObjSchema({
+  allowedInvalidRoles: {
+    items: {
+      type: 'string',
+    },
+    type: 'array',
+    uniqueItems: true,
+  },
   ignoreNonDOM: {
     type: 'boolean',
     default: false,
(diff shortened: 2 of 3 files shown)


src/rules/aria-role.js
```diff
@@ -28,42 +35,44 @@ export default {
...
-  create: (context) => ({
-    JSXAttribute: (attribute) => {
-      // Determine if ignoreNonDOM is set to true
-      // If true, then do not run rule.
-      const options = context.options[0] || {};
-      const ignoreNonDOM = !!options.ignoreNonDOM;
+  create: (context) => {
+    const options = context.options[0] || {};
+    const ignoreNonDOM = !!options.ignoreNonDOM;
+    const allowedInvalidRoles = new Set(options.allowedInvalidRoles || [])
```
JoshuaKGoldberg, line 41: The diff is a little unfriendly 😞 but all I did for most lines was move `const options` parsing to the top of `create` instead of running it inside each `JSXAttribute`.

src/rules/aria-role.js
```diff
@@ -28,42 +35,44 @@ export default {
...
+        // If value is undefined, then the role attribute will be dropped in the DOM.
+        // If value is null, then getLiteralAttributeValue is telling us that the
+        // value isn't in the form of a literal.
+        if (value === undefined || value === null) { return; }
 
-      const values = String(value).split(' ');
-      const validRoles = [...roles.keys()].filter((role) => roles.get(role).abstract === false);
-      const isValid = values.every((val) => validRoles.indexOf(val) > -1);
+        const values = String(value).split(' ');
+        const validRoles = [...roles.keys()].filter((role) => roles.get(role).abstract === false);
```
JoshuaKGoldberg, line 66: Related: is there a reason to do this inside each `JSXAttribute`, rather than once per rule (in the top of `create`)? If not, I'd be happy to move it up in this PR.

docs/rules/aria-role.md
```diff
@@ -10,12 +10,15 @@ This rule takes one optional object argument of type object:
...
     "rules": {
         "jsx-a11y/aria-role": [ 2, {
+            "allowedInvalidRoles": ["text"],
             "ignoreNonDOM": true
         }],
     }
 }
 ```
 
+`allowedInvalidRules` is an optional string array of custom roles that should be allowed in addition to the ARIA spec.
```
ljharb, line 20: it might be nice to include an example of "text", and why that's reasonable?

src/rules/aria-role.js
ljharb: nope, looks like this should be hoisted to the top of the module, since it doesn't depend on anything except `roles`.

ljharb (commented):
LGTM with minor comments

JoshuaKGoldberg:
Ping @ljharb - could I request a re-review please?

ljharb (approved):
LGTM!

coreyward:
Any idea when this will be published?

ljharb:
Nope, no idea. It will be included in the next release, whenever that is.

johan:
It would be lovely to get a `v6.6.0` published from main with this, so we can use npm version number dependencies rather than git urls to use the feature.

(+ A bit confusing to find the option in the official docs but not be able to use it in the latest published version, three weeks after it landed.)

ljharb:
Re the last part, for any GitHub repo, docs on the default branch often contain unpublished changes - you need to check under the tag.

coreyward:
I encourage maintainers to circle back to issues and PRs with a comment mentioning the version that includes a fix or merge. Since a lot of users will find issue pages while trying to self serve, the additional information helps reduce confusion and create awareness without too much extra work.

ljharb:
That’s unsustainable to do manually, and my experience with bots that auto comment on release is that they’re excessively noisy.

Unfortunately the best outcome would come with user education about how to understand when a closed issue/PR is, or isn’t, released.

coreyward:
In practice, it's not bad. I also release when there are changes that fix issues. I find that proactive communication cuts down on the volume of pings, asks, and follow-up issues. A little extra work to solve a lot of extra work makes sense to me, but obviously you are welcome to do things however works best for you, I'm just sharing what I've learned.

PR #833, opened 2022-02-24
New rule: prefer-tag-over-role

### LINKED ISSUE #101: rules to warn on redundant roles
`role` is probably the most misused `aria` attribute, for example #88, #91.

Adding `role=button` to a `div` does not make it a button. 
One has to script the behavior, i.e. if `enter`/`space` was pressed invoke the `click` handler. 
Also add `tabindex=0` to make it focusable. 

However if one had used `<button>` attaching a click handler would be sufficient.

`<button>` is focusable/tabable by default (no need of tabindex=0), and it triggers a `click` event when `enter`/`space` is pressed (no need for extra keypress event listeners).

Same for `link`, `checkbox`, `radio`, `radiogroup`, `textbox` and the proposed [`password`](http://tink.uk/proposed-aria-password-role/) roles. There are already corresponding HTML alternatives available, and one should use them instead of just adding roles.

I guess It is better to call out (error, warn) in such situation.

JoshuaKGoldberg:
Fixes #101

### DIFF
--- a/__tests__/src/rules/prefer-tag-over-role-test.js
+++ b/__tests__/src/rules/prefer-tag-over-role-test.js
@@ -1,0 +1,66 @@
+import { RuleTester } from 'eslint';
+import parserOptionsMapper from '../../__util__/parserOptionsMapper';
+import rule from '../../../src/rules/prefer-tag-over-role';
+
+const ruleTester = new RuleTester();
+
+const expectedError = (role, tag) => ({
+  message: `Use ${tag} instead of the "${role}" role to ensure accessibility across all devices.`,
+  type: 'JSXOpeningElement',
+});
+
+ruleTester.run('element-role', rule, {
+  valid: [
+    { code: '<div />;' },
+    { code: '<div role="unknown" />;' },
+    { code: '<div role="also unknown" />;' },
+    { code: '<other />' },
+    { code: '<img role="img" />' },
+    { code: '<input role="checkbox" />' },
+  ].map(parserOptionsMapper),
+  invalid: [
+    {
+      code: '<div role="checkbox" />',
+      errors: [expectedError('checkbox', '<input type="checkbox">')],
+    },
+    {
+      code: '<div role="button checkbox" />',
+      errors: [expectedError('checkbox', '<input type="checkbox">')],
+    },
+    {
+      code: '<div role="heading" />',
+      errors: [
+        expectedError('heading', '<h1>, <h2>, <h3>, <h4>, <h5>, or <h6>'),
+      ],
+    },
+    {
+      code: '<div role="link" />',
+      errors: [
+        expectedError(
+          'link',
+          '<a href=...>, <area href=...>, or <link href=...>',
+        ),
+      ],
+    },
+    {
+      code: '<div role="rowgroup" />',
+      errors: [expectedError('rowgroup', '<tbody>, <tfoot>, or <thead>')],
+    },
+    {
+      code: '<span role="checkbox" />',
+      errors: [expectedError('checkbox', '<input type="checkbox">')],
+    },
+    {
+      code: '<other role="checkbox" />',
+      errors: [expectedError('checkbox', '<input type="checkbox">')],
+    },
+    {
+      code: '<other role="checkbox" />',
+      errors: [expectedError('checkbox', '<input type="checkbox">')],
+    },
+    {
+      code: '<div role="banner" />',
+      errors: [expectedError('banner', '<header>')],
+    },
+  ].map(parserOptionsMapper),
+});
--- a/src/rules/prefer-tag-over-role.js
+++ b/src/rules/prefer-tag-over-role.js
@@ -1,0 +1,87 @@
+import { roleElements } from 'aria-query';
+import { getProp, getPropValue } from 'jsx-ast-utils';
+
+import getElementType from '../util/getElementType';
+import { generateObjSchema } from '../util/schemas';
+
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -28,7 +28,7 @@
     "test": "npm run jest",
     "posttest": "aud --production",
     "test:ci": "npm run jest -- --ci --runInBand",
-    "jest": "jest --coverage __tests__/**/*"
+    "jest2": "jest --coverage __tests__/**/*"
```
JoshuaKGoldberg, line 31: whoops 😄 

```suggestion
    "jest": "jest --coverage __tests__/**/*"
```

__tests__/src/rules/prefer-tag-over-role-test.js
```diff
@@ -0,0 +1,43 @@
+import { RuleTester } from 'eslint';
+import parserOptionsMapper from '../../__util__/parserOptionsMapper';
+import rule from '../../../src/rules/prefer-tag-over-role';
+
+const ruleTester = new RuleTester();
+
+const expectedError = (role, tag) => ({
+  message:
+    `Use <${tag}> instead of the "${role}" role to ensure accessibility across all devices.`,
```
ljharb, line 9:
```suggestion
  message: `Use <${tag}> instead of the "${role}" role to ensure accessibility across all devices.`,
```
i smell in-editor prettier :-p

ljharb (approved):
Seems pretty straightforward

jessebeach (changes requested):
It's a good rule. Thank you for proposing it. I just have a few improvements that will use our utilities and provide more robustness.

src/rules/prefer-tag-over-role.js
```diff
@@ -0,0 +1,59 @@
...
+
+      const matchedTags = roleElements.get(role);
+      if (!matchedTags) {
+        return;
+      }
+
+      const matchedNames = Array.from(
+        new Set(Array.from(matchedTags).map((matchedTag) => matchedTag.name))
+      );
+      if (matchedNames.some((matchedTag) => matchedTag === node.name.name)) {
```
jessebeach, line 36: Use the `getElementType` util here instead of referencing `node.name.name` directly. `getElementType` deals with configured Component names as well as namespaces: 

1. Implementation example: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/7f6463e5cffd1faa5cf22e3b0d33465e22bd10e1/src/rules/no-static-element-interactions.js#L56
2. Definition: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/7f6463e5cffd1faa5cf22e3b0d33465e22bd10e1/src/util/getElementType.js
3. jsx-ast-utils source for reference: https://github.com/jsx-eslint/jsx-ast-utils/blob/main/src/elementType.js

src/rules/prefer-tag-over-role.js
```diff
@@ -0,0 +1,59 @@
+import { roleElements } from "aria-query";
+import { getProp, getPropValue } from "jsx-ast-utils";
+
+import { generateObjSchema } from "../util/schemas";
+
+const errorMessage =
+  'Use {{tag}} instead of the "{{role}}" role to ensure accessibility across all devices.';
```
jessebeach, line 7: `tag` should include attributes if necessary. For example, `<a>` is semantically an anchor on a page, but `<a href>` is a link. `<input>` default to `textbox`, whereas `<input type="checkbox`>` is a checkbox. The recommendation here should be as specific as possible. `roleElements` will give you the `attributes` associated with the tag name, so you can build this string with maximum detail.

__tests__/src/rules/prefer-tag-over-role-test.js
```diff
@@ -0,0 +1,52 @@
...
+import rule from "../../../src/rules/prefer-tag-over-role";
+
+const ruleTester = new RuleTester();
+
+const expectedError = (role, tag) => ({
+  message: `Use ${tag} instead of the "${role}" role to ensure accessibility across all devices.`,
+  type: "JSXOpeningElement",
+});
+
+ruleTester.run("element-rol", rule, {
```
jessebeach, line 12: nit: "element-role"

jessebeach, line 12: Please add tests for multiple roles, such as `<div role="button checkbox">`. I think your rule will fail in this case. The last role in a space-delimited list will be the one that applies.

jessebeach (changes requested):
This is really close!! Just a few polishing tweaks to get it over the line! I think we'll be able to commit it after these few changes.

src/rules/prefer-tag-over-role.js
JoshuaKGoldberg: Love it, thanks!

__tests__/src/rules/prefer-tag-over-role-test.js
JoshuaKGoldberg: TIL!

jessebeach (approved):
Thank you for the updates!

@ljharb looks good to ship!

jessebeach:
There's still a little work to expose the rule through the plugin, but this is good enough to get the rule logic into the code.

jessebeach:
Thanks @JoshuaKGoldberg for proposing and writing this rule!

JoshuaKGoldberg:
Thanks for the helpful+informative reviews @jessebeach!

And oops, I didn't think to add to the plugin rules & recommended sets. I can send a PR now.

ljharb:
@JoshuaKGoldberg somehow we all totally missed the docs for this rule - could you send a PR to add them?

PR #835, opened 2022-02-24
Ignore contenteditable elements in no-noninteractive-element-interactions

### LINKED ISSUE #525: Should contentEditable elems be excempt from no-noninteractive-element-interactions?
I have a `pre` element set as `contentEditable` for reasons. Thoughts on removing `contentEditable` elements from the [no-noninteractive-element-interactions](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-noninteractive-element-interactions.md) rule?

I also added `role="textbox"` since that's really what it is, but I have other `input`s inserting content in specific areas of the `contentEditable` elem; this is why it's _not_ an actual `textarea`. However, doing so comes back with the [no-noninteractive-element-to-interactive-role](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-noninteractive-element-to-interactive-role.md) error.

Here's the snippet in question:

```html
<pre
  aria-label="Edit description"
  aria-multiline="true"
  contentEditable
  onKeyUp={this.doSomething}
  role="textbox"
>
  <!-- 
    Other HTML which builds the content of this container
    which, in turn, can also be edited directly.
  -->
</pre>
```

My thoughts are since this has `contentEditable` it's now an interactive element and could be exempt from this rule.

JoshuaKGoldberg:
Fixes #525

### DIFF
--- a/__mocks__/JSXAttributeMock.js
+++ b/__mocks__/JSXAttributeMock.js
@@ -24,6 +24,8 @@ export default function JSXAttributeMock(prop: string, value: mixed, isExpressio
   let attributeValue = astValue;
   if (isExpressionContainer || astValue.type !== 'Literal') {
     attributeValue = JSXExpressionContainerMock(astValue);
+  } else if (attributeValue.type === 'Literal' && !('raw' in (attributeValue: any))) {
+    (attributeValue: any).raw = JSON.stringify((attributeValue: any).value);
   }
 
   return {
--- a/src/util/isContentEditable.js
+++ b/src/util/isContentEditable.js
@@ -1,0 +1,7 @@
+import { getProp } from 'jsx-ast-utils';
+
+export default function isContentEditable(tagName, attributes) {
+  const prop = getProp(attributes, 'contentEditable');
+
+  return prop?.value?.raw === '"true"';
+}
--- a/__tests__/src/rules/no-noninteractive-element-interactions-test.js
+++ b/__tests__/src/rules/no-noninteractive-element-interactions-test.js
@@ -337,6 +337,9 @@ const neverValid = [
   { code: '<time onClick={() => {}} />;', errors: [expectedError] },
   { code: '<ol onClick={() => {}} />;', errors: [expectedError] },
   { code: '<ul onClick={() => {}} />;', errors: [expectedError] },
+  { code: '<ul contentEditable="false" onClick={() => {}} />;', errors: [expectedError] },
+  { code: '<article contentEditable onClick={() => {}} />;', errors: [expectedError] },
+  { code: '<div contentEditable role="article" onKeyDown={() => {}} />;', errors: [expectedError] },
   /* HTML elements attributed with a non-interactive role */
   { code: '<div role="alert" onClick={() => {}} />;', errors: [expectedError] },
   { code: '<div role="alertdialog" onClick={() => {}} />;', errors: [expectedError] },
--- a/__tests__/src/util/isContentEditable-test.js
+++ b/__tests__/src/util/isContentEditable-test.js
@@ -1,0 +1,51 @@
+import expect from 'expect';
+import isContentEditable from '../../../src/util/isContentEditable';
+import JSXAttributeMock from '../../../__mocks__/JSXAttributeMock';
+
+describe('isContentEditable', () => {
+  describe('HTML5', () => {
+    describe('content editable', () => {
+      it('should identify HTML5 contentEditable elements', () => {
+        const attributes = [
+          JSXAttributeMock('contentEditable', 'true'),
+        ];
+        expect(isContentEditable('some tag', attributes))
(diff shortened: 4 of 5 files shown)


src/util/isContentEditable.js
```diff
@@ -0,0 +1,8 @@
+import { getProp } from "jsx-ast-utils";
+
+const isContentEditable = (tagName, attributes) => {
+  const prop = getProp(attributes, "contentEditable");
+  return prop && prop.value?.raw !== '"false"';
```
ljharb, line 5:
```suggestion
  return getProp(attributes, "contentEditable")?.value?.raw !== '"false"';
```
?

ljharb (approved):
LGTM

src/util/isContentEditable.js
JoshuaKGoldberg: 👌 snazzy

src/util/isContentEditable.js
```diff
@@ -0,0 +1,7 @@
+import { getProp } from "jsx-ast-utils";
+
+const isContentEditable = (tagName, attributes) => {
+  return getProp(attributes, "contentEditable")?.value?.raw !== '"false"';
+};
+
+export default isContentEditable;
```
ljharb, line 7:
```suggestion
export default function isContentEditable(tagName, attributes) {
  return getProp(attributes, 'contentEditable')?.value?.raw !== '"false"';
};
```

JoshuaKGoldberg: Love it -- I was just going off the existing sibling util files. +1

src/util/isContentEditable.js
```diff
@@ -0,0 +1,5 @@
+import { getProp } from 'jsx-ast-utils';
+
+export default function isContentEditable(tagName, attributes) {
+  return getProp(attributes, 'contentEditable')?.value?.raw !== '"false"';
```
JoshuaKGoldberg, line 4: Hmm, actually, what if it's `undefined` for not existing... will update.

src/util/isContentEditable.js
```diff
@@ -0,0 +1,7 @@
+import { getProp } from 'jsx-ast-utils';
+
+export default function isContentEditable(tagName, attributes) {
```
ljharb, line 3: this one needs tests (i'll add them shortly)

ljharb:
@JoshuaKGoldberg so, i rebased this, and tweaked the logic slightly (from "not false" to "is true") based on my testing - iow, that browsers normalize all values outside the enum of "true", "false", or "inherit" to "false" (and the JS setter refuses any value but those three), and now some tests are failing.

Mind resetting your local branch, and taking a look?

src/util/isContentEditable.js
```diff
@@ -0,0 +1,11 @@
...
+  const prop = getProp(attributes, 'contentEditable');
+
+  if (!prop || !prop.value || typeof prop.value.raw !== 'string') {
```
JoshuaKGoldberg, line 6:
```suggestion
  if (!prop?.value || typeof prop.value.raw !== 'string') {
```

src/util/isContentEditable.js
```diff
@@ -0,0 +1,11 @@
...
+    return false;
+  }
+
+  return prop.value.raw === '"true"';
```
JoshuaKGoldberg, line 10: Hmm, is this a result of the way JSX attributes get parsed? In my Chrome 104, `<div contenteditable>` works the same as `<div contentEditable="true">`.

Screenshot of adding contenteditable to a div in Chrome dev tools, and the div being contenteditable

ljharb: yes, if you check out that HTML element and look at `.contentEditable` you'll get the string true.

__mocks__/JSXAttributeMock.js
```diff
@@ -24,6 +25,8 @@ export default function JSXAttributeMock(prop: string, value: mixed, isExpressio
   let attributeValue = astValue;
   if (isExpressionContainer || astValue.type !== 'Literal') {
     attributeValue = JSXExpressionContainerMock(astValue);
+  } else if (attributeValue.type === 'Literal' && !('raw' in attributeValue)) {
+    attributeValue.raw = JSON.stringify(attributeValue.value);
```
JoshuaKGoldberg, line 29: Sorry, I am a Flow noob and it exits with `Unknown system error -86` locally. I don't know how to fix the type errors here. 🙃 

...can we switch to TypeScript? 😄

ljharb: I'll take a look at it.

Separately, flow's been dead for years, but a) i'm not a fan of TS, and b) the work involved is likely not small.

JoshuaKGoldberg:
I'm working on the Flow issues now, will rebase once I'm done...

PR #843, opened 2022-03-16
[Docs] Update error message for no-static-element-interactions rule

### LINKED ISSUE #761: `no-static-element-interactions` message suggests adding a role, but not changing the element.
(sorry about the title, this is a very tough title to write)

The [docs for the no-static-element-interactions rule](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/no-static-element-interactions.md) cover two common cases of needing a static element.

The message on error:
```
Static HTML elements with event handlers require a role.
```

This typically happens in pairs with `click-events-have-key-events`:
```
Static HTML elements with event handlers require a role. (jsx-a11y/no-static-element-interactions)
Visible, non-interactive elements with click handlers must have at least one keyboard listener. (jsx-a11y/click-events-have-key-events)
```

However, both the docs and message are dancing around an issue: _people use clickable divs because they don't know any better_. Making an interactive `div` accessible requires a minimum of 3 extra properties, plus all the keyboard event handlers on top of click handlers (and all the `keyCode` listening necessary to match the role). The best fix for this rule is almost always to change the static element into an interactive element, but the message sends people down a more difficult and verbose path.

Would you be willing to accept a change to have the message reflect this? I'm struggling with grammar consistency, but something like:
```
A static HTML element with an event handler requires a role, or should be replaced with an interactive element.
```

uncommon-type:
## Description

This PR updates error message for `no-static-element-interactions rule`. It recommends using native HTML whenever possible. If for some reason it is not, this PR suggests adding an appropriate role along with support for mouse, keyboard, and touch inputs to all interactive non-semantic elements.

## Related Issue

Closes #761   

## Acceptance Criteria
- [x] Recommend using native HTML whenever possible
- [x] Recommend adding to all interactive non-semantic elements an appropriate role as well as support for mouse, keyboard and touch input

## Type of Changes

|     | Type                       |
| --- | -------------------------- |
|    | :bug: Bug fix              |
|    | :sparkles: New feature     |
|    | :hammer: Refactoring       |
|    | :100: Add tests            |
|    | :link: Update dependencies |
|  ✓  | :scroll: Docs              |

### DIFF
--- a/__tests__/src/rules/no-static-element-interactions-test.js
+++ b/__tests__/src/rules/no-static-element-interactions-test.js
@@ -19,7 +19,7 @@ import ruleOptionsMapperFactory from '../../__util__/ruleOptionsMapperFactory';
 
 const ruleTester = new RuleTester();
 
-const errorMessage = 'Static HTML elements with event handlers require a role.';
+const errorMessage = 'Avoid non-native interactive elements. If using native HTML is not possible, add an appropriate role and support for tabbing, mouse, keyboard, and touch inputs to an interactive content element.';
 
 const expectedError = {
   message: errorMessage,
--- a/src/rules/no-static-element-interactions.js
+++ b/src/rules/no-static-element-interactions.js
@@ -29,7 +29,7 @@ import isNonInteractiveRole from '../util/isNonInteractiveRole';
 import isNonLiteralProperty from '../util/isNonLiteralProperty';
 import isPresentationRole from '../util/isPresentationRole';
 
-const errorMessage = 'Static HTML elements with event handlers require a role.';
+const errorMessage = 'Avoid non-native interactive elements. If using native HTML is not possible, add an appropriate role and support for tabbing, mouse, keyboard, and touch inputs to an interactive content element.';
 
 const domElements = [...dom.keys()];
 const defaultInteractiveProps = [


__tests__/src/rules/no-static-element-interactions-test.js
```diff
@@ -19,7 +19,7 @@ import ruleOptionsMapperFactory from '../../__util__/ruleOptionsMapperFactory';
 
 const ruleTester = new RuleTester();
 
-const errorMessage = 'Static HTML elements with event handlers require a role.';
+const errorMessage = 'Avoid non-native interactive elements. If using native HTML is not possible, add an appropriate role and support for tabbing, mouse, keyboard and touch inputs to an interactive content element.';
```
ljharb, line 22:
```suggestion
const errorMessage = 'Avoid non-native interactive elements. If using native HTML is not possible, add an appropriate role and support for tabbing, mouse, keyboard, and touch inputs to an interactive content element.';
```

src/rules/no-static-element-interactions.js
```diff
@@ -29,7 +29,7 @@ import isNonInteractiveRole from '../util/isNonInteractiveRole';
 import isNonLiteralProperty from '../util/isNonLiteralProperty';
 import isPresentationRole from '../util/isPresentationRole';
 
-const errorMessage = 'Static HTML elements with event handlers require a role.';
+const errorMessage = 'Avoid non-native interactive elements. If using native HTML is not possible, add an appropriate role and support for tabbing, mouse, keyboard and touch inputs to an interactive content element.';
```
ljharb, line 32:
```suggestion
const errorMessage = 'Avoid non-native interactive elements. If using native HTML is not possible, add an appropriate role and support for tabbing, mouse, keyboard, and touch inputs to an interactive content element.';
```

ljharb (approved):
Seems great!

Issue #174, opened 2017-02-04
Implement global settings


beefancohen:
Discussed in #65. We can provide global settings API for users to specify mappings from low-level DOM elements to their associated wrapper component. This allows us to better apply the ruleset over a codebase. It can also have configuration options for `ARIA` convention for props (i.e. camel-casing like `ariaExpanded` or just `expanded`) that map to respective `ARIA` properties. Also contains a special property for `role` since that sometimes has conflicts (as seen in #166).

Can look something like this:

```json
{
	"settings": {
		"components": {
			"a": ["Link"],
			"abbr": [],
			"address": [],
			"area": [],
			"article": [],
			"aside": [],
			"audio": [],
			"b": [],
			"base": [],
			"bdi": [],
			"bdo": [],
			"big": [],
			"blockquote": [],
			"body": [],
			"br": [],
			"button": ["Button"],
			"canvas": [],
			"caption": [],
			"cite": [],
			"code": [],
			"col": [],
			"colgroup": [],
			"data": [],
			"datalist": [],
			"dd": [],
			"del": [],
			"details": [],
			"dfn": [],
			"dialog": [],
			"div": [],
			"dl": [],
			"dt": [],
			"em": [],
			"embed": [],
			"fieldset": [],
			"figcaption": [],
			"figure": [],
			"footer": ["Footer"],
			"form": [],
			"h1": [],
			"h2": [],
			"h3": [],
			"h4": [],
			"h5": [],
			"h6": [],
			"head": [],
			"header": ["Header"],
			"hgroup": [],
			"hr": [],
			"html": [],
			"i": [],
			"iframe": [],
			"img": ["Image"],
			"input": ["Input"],
			"ins": [],
			"kbd": [],
			"keygen": [],
			"label": ["Label"],
			"legend": [],
			"li": [],
			"link": [],
			"main": [],
			"map": [],
			"mark": [],
			"menu": [],
			"menuitem": [],
			"meta": [],
			"meter": [],
			"nav": [],
			"noscript": [],
			"object": [],
			"ol": [],
			"optgroup": [],
			"option": [],
			"output": [],
			"p": [],
			"param": [],
			"picture": [],
			"pre": [],
			"progress": [],
			"q": [],
			"rp": [],
			"rt": [],
			"ruby": [],
			"s": [],
			"samp": [],
			"script": [],
			"section": [],
			"select": [],
			"small": [],
			"source": [],
			"span": [],
[...]

PR #844, opened 2022-03-24
[New] Introduce a plugin-wide setting for custom components.

pcorpet:
For #174

This is an early version of the change I want to introduce: there are still some rules that I haven't updated.
Can you start a review on the API and the implementation to see if I should finish it please?

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,24 @@ Add `plugin:jsx-a11y/recommended` or `plugin:jsx-a11y/strict` in `extends`:
 }
 ```
 
+To enable your custom components to be checked as DOM elements, you can set global settings in your
+configuration file by mapping each custom component name to a DOM element type.
+
+```json
+{
+  "settings": {
+    "jsx-a11y": {
+      "components": {
+        "CityInput": "input",
+        "CustomButton": "button",
+        "MyButton": "button",
+        "RoundButton": "button"
+      }
+    }
+  }
+}
+```
+
 ## Supported Rules
 
 <!-- AUTO-GENERATED-CONTENT:START (LIST) -->
--- a/__tests__/src/rules/accessible-emoji-test.js
+++ b/__tests__/src/rules/accessible-emoji-test.js
@@ -37,6 +37,11 @@ ruleTester.run('accessible-emoji', rule, {
     { code: '<span aria-hidden="true">🐼</span>' },
     { code: '<span aria-hidden>🐼</span>' },
     { code: '<div aria-hidden="true">🐼</div>' },
+    { code: '<input type="hidden">🐼</input>' },
+    {
+      code: '<CustomInput type="hidden">🐼</CustomInput>',
+      settings: { 'jsx-a11y': { components: { CustomInput: 'input' } } },
+    },
   ].map(parserOptionsMapper),
   invalid: [
     { code: '<span>🐼</span>', errors: [expectedError] },
@@ -46,5 +51,6 @@ ruleTester.run('accessible-emoji', rule, {
     { code: '<i role="img" aria-labelledby="id1">🐼</i>', errors: [expectedError] },
     { code: '<Foo>🐼</Foo>', errors: [expectedError] },
     { code: '<span aria-hidden="false">🐼</span>', errors: [expectedError] },
+    { code: '<CustomInput type="hidden">🐼</CustomInput>', errors: [expectedError] },
   ].map(parserOptionsMapper),
 });
--- a/__tests__/src/rules/no-autofocus-test.js
+++ b/__tests__/src/rules/no-autofocus-test.js
@@ -28,6 +28,14 @@ const ignoreNonDOMSchema = [
   },
 ];
 
+const componentsSettings = {
+  'jsx-a11y': {
+    components: {
+      Button: 'button',
+    },
+  },
+};
+
 ruleTester.run('no-autofocus', rule, {
   valid: [
     { code: '<div />;' },
@@ -36,6 +44,8 @@ ruleTester.run('no-autofocus', rule, {
     { code: '<Foo bar />' },
     { code: '<Foo autoFocus />', options: ignoreNonDOMSchema },
     { code: '<div><div autofocus /></div>', options: ignoreNonDOMSchema },
+    { code: '<Button />', settings: componentsSettings },
+    { code: '<Button />', options: ignoreNonDOMSchema, settings: componentsSettings },
   ].map(parserOptionsMapper),
   invalid: [
     { code: '<div autoFocus />', errors: [expectedError] },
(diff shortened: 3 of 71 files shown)


jessebeach:
Love it, thank you for starting this. 

I have a 10 week old at home and he's been a bit of a drag on time and energy for Open Source :) I'll try to find time to review, but if I'm a little slow, it's cause I'm changing diapers ;p

__tests__/src/rules/accessible-emoji-test.js
```diff
@@ -37,6 +37,11 @@ ruleTester.run('accessible-emoji', rule, {
     { code: '<span aria-hidden="true">🐼</span>' },
     { code: '<span aria-hidden>🐼</span>' },
     { code: '<div aria-hidden="true">🐼</div>' },
+    { code: '<input type="hidden">🐼</input>' },
+    {
+      code: '<CustomInput type="hidden">🐼</CustomInput>',
+      settings: { 'jsx-a11y': { components: { CustomInput: 'input' } } },
```
ljharb, line 43: should this be a map of custom to html, or a map of html to a unique array of customs? seems like that'd make it easier to audit "all the custom x elements" in a project that has a lot.

src/util/getElementType.js
```diff
@@ -0,0 +1,19 @@
...
+ */
+
+import type { JSXOpeningElement } from 'ast-types-flow';
+import { elementType } from 'jsx-ast-utils';
+
+import type { ESLintContext } from '../../flow/eslint';
+
+const getElementType = (context: ESLintContext): ((node: JSXOpeningElement) => string) => {
+  const { settings } = context;
+  const componentMap = (settings['jsx-a11y'] || {}).components || {};
```
ljharb, line 12:
```suggestion
  if (!settings['jsx-a11y']?.components) {
    return elementType;
  }
  const componentMap = settings['jsx-a11y'].components;
```

__tests__/src/rules/accessible-emoji-test.js
pcorpet: I was worried to have components added in several html lists. However I've changed it to your suggestion and checked the setting when loading it.

src/util/getElementType.js
pcorpet: Of course! Thanks for the simplification.

__tests__/src/rules/accessible-emoji-test.js
ljharb: Hm, that’s actually a really good point :-) maybe we could validate that at runtime, with either configuration we choose?

pcorpet: Yes, please check how I've implemented in my latest version.

src/util/getElementType.js
```diff
@@ -0,0 +1,33 @@
...
+  Object.keys(componentMap).forEach((domElement) => {
+    const customComponents = componentMap[domElement];
+    customComponents.forEach((customComponent) => {
+      const existingDomElement = reverseComponentMaps[customComponent];
+      if (existingDomElement && existingDomElement !== domElement) {
+        throw new Error(`${customComponent} was specified twice.`);
+      }
+      reverseComponentMaps[customComponent] = domElement;
+    });
+  });
```
ljharb, line 26: seems like this could probably be done with Object.fromEntries (https://npmjs.com/object.fromentries) and `.map`?

pcorpet: I didn't think about importing a new package, and the CI stopped me short when I wanted to use `Object.entries`.

As for `Object.fromEntries`, the issue is that it wouldn't tell us about duplicate keys.

What do you think?

ljharb: We can use the object.entries package as well.

I think maybe your original schema was better, since it enforced uniqueness by default.

pcorpet: I've used object.entries and documented.

I like the readability of your suggestion and this is also how most rules options were implemented already.

README.md
```diff
@@ -89,6 +89,22 @@ Add `plugin:jsx-a11y/recommended` or `plugin:jsx-a11y/strict` in `extends`:
 }
 ```
 
+To enable your custom components to be checked as DOM elements, you can set global settings in your
+configuration file by mapping each DOM element type to your component names.
+
+```json
+{
+  "settins": {
```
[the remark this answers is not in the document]
pcorpet: Done. Thanks.

__tests__/src/rules/no-autofocus-test.js
```diff
@@ -28,6 +28,14 @@ const ignoreNonDOMSchema = [
   },
 ];
 
+const componentsSettings = {
+  'jsx-a11y': {
+    components: {
+      button: ['Button'],
```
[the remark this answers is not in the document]
pcorpet: You're the boss, boss! :-)

src/util/getElementType.js
```diff
@@ -0,0 +1,22 @@
...
+
+const getElementType = (context: ESLintContext): ((node: JSXOpeningElement) => string) => {
+  const { settings } = context;
+  const componentMap = settings['jsx-a11y']?.components;
+  if (!componentMap) {
+    return elementType;
+  }
+  return (node: JSXOpeningElement): string => {
+    const rawType = elementType(node);
+    return componentMap[rawType] || rawType;
```
ljharb, line 18: what about a custom component of `toString` etc? i guess we should handle that.
```suggestion
    return has(componentMap, rawType) ? componentMap[rawType] : rawType;
```
(with the `has` package)

ljharb (commented):
Looks great!

src/util/getElementType.js
pcorpet: > what about a custom component of toString etc? 

I'm not sure I understand what is a custom component of `toString`

Note that the settings are coming from a JSON and thus are supposed to be plain. Can you give me a test case that would be handled differently by your suggestion please?

One I can think of is setting an empty string, although I'd say it's an abuse of the settings:
```json
"components": {
  "input": ""
}
```
That would silence all the rules alerting on `input` components.

pcorpet:
I'm done updating every existing rule and helper function that was using the raw `elemType` function and I've added test cases for all of them I believe.

joshuajaco:
We also need something like this, because `control-has-associated-label`'s `controlComponents` option currently does not allow globs, unlike `label-has-associated-control`. I would greatly appreciate this being merged soon.

src/util/getElementType.js
ljharb: What i mean is, there could be jsx like this:
```jsx
<toString />
```
and `Object.prototype.toString` exists.

The settings can be plain, but the actual jsx won't be, nor is this JS code that reads the setting.

pcorpet: OK, got it. I wasn't able to create a test that would actually be harmful, but that might be the case on older versions of Node. Done.

ljharb (approved):
Looks great!

Issue #772, opened 2021-02-12
custom rule - how?


Felicia5:
Hi,
I am trying to add a custom rule using the logic of heading-has-content rule. But instead of headings I am looking for div elements. Also, in the util hasAccessibleChild I am not only looking for any text but a text that is equal to "apply". 

#My questions:
1. I don't know how to handle these errors. Can you suggest some solutions?
    I generated the rule files using the create-rule.md file. I added my changes to the rule and rule-test files. I run npm test and receive these errors: 
    
    ![image](https://user-images.githubusercontent.com/13880509/107732476-580b0480-6cf0-11eb-83f9-a84f7a727335.png)

***Do you have a further guide on these below?***

    2. How do I test if the rule and the rule-test works as I want it to? 
      I have read the [eslint documentation](https://eslint.org/docs/developer-guide/unit-tests) but I couldn't find a solution. 

  3. Without merging my rule to the master branch, how do I add my custom  jsx-eslint /eslint-plugin-jsx-a11y plugin to test my code against the my new custom rule when I develop a new website?

here is my custom rule:
https://github.com/Felicia5/eslint-plugin-jsx-a11y/blob/first-rule-button/src/rules/first-rule-button.js

ljharb:
Hmm - I haven't actually used that script; it was last touched 4 years ago. It's highly likely that it's out of date.

The error looks like it's coming from `flow` itself - what code are you running it on?

Felicia5:
I got this error right after I :

1. run npm run create -- myrule --author="Felicia Kovacs" --description="Enforce use of button elements to display apply text"  - this is the command defined in the : https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/scripts/create-rule.md
2. added code to files: rule, rule-test and hasAccessibleChildApply.js util
3.  run npm test within int-plugin-jsx-a11y folder I forked and cloned on my local desktop. 

Any idea how I can get rid of the error message coming from flow?

ljharb:
Can you make a branch on your fork, and post the link here? I'll clone it locally and try it out.

Felicia5:
here it is: https://github.com/Felicia5/eslint-plugin-jsx-a11y/tree/felicia-rule

Felicia5:
These are the steps I have done after forking:

used git clone https://github.com/Felicia5/eslint-plugin-jsx-a11y.git 
cd eslint-plugin-jsx-a11y
[...]

PR #861, opened 2022-06-23
add `export default` instead of `module.exports`

V2dha:
add `export default` instead of `module.exports` in the boilerplate for `rule.js` to avoid `Cannot use import declarations in modules that export using CommonJS` errors as mentioned in the issue [#772 ](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/772)


ljharb (approved):
thanks, good catch.

Issue #766, opened 2021-01-19
`no-static-element-interactions` cannot be resolved with dynamically-defined roles


bcronin2:
I'm not sure if this behavior is intentional, but in order to satisfy the `no-static-element-interactions` check, a single string literal must be provided for the role:

![image](https://user-images.githubusercontent.com/6960884/105061409-540bff80-5a2e-11eb-9104-c0b676af1665.png)

There are workarounds to this, but it'd be nice if it were possible to specify that the role be one of two (or more) dynamically determined interactive roles (as in the example).

ljharb:
Linters can not analyze dynamic values at all.

bcronin2:
Fair enough, good to know

ljharb:
Although, in this specific case (a ternary with literals on both sides) the rule should be able to handle it. PRs welcome.

V2dha:
Can I work on this?

ljharb:
Go for it! Anything with "help wanted" is up for grabs.

V2dha:
@ljharb for getting the prop value assigned to the `role` attribute, I am using `getProp` and `getPropValue` functions 
`const roleProp = getPropValue(getProp(attributes, 'role'));` 
but while getting its type it is giving me error. `roleProp.type === 'ConditionalExpression'` 
Does the function just returns the value and not the node?

PR #865, opened 2022-07-01
[Fix] `no-static-element-interactions`: allow role assignments using a ternary with literals on both sides

V2dha:
Fixes [766](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/766)
Initial PR for adding a specific case for role assignment in static HTML elements using ternary operator with literals on both sides.

### DIFF
--- a/__tests__/src/rules/no-static-element-interactions-test.js
+++ b/__tests__/src/rules/no-static-element-interactions-test.js
@@ -429,6 +429,11 @@ ruleTester.run(`${ruleName}:recommended`, rule, {
     // Expressions should pass in recommended mode
     { code: '<div role={ROLE_BUTTON} onClick={() => {}} />;' },
     { code: '<div  {...this.props} role={this.props.role} onKeyPress={e => this.handleKeyPress(e)}>{this.props.children}</div>' },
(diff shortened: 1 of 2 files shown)


ljharb:
I rebased this and added a test case, which fails both without your fix and with it.

V2dha:
@ljharb Can you review the code that I added once? I feel like it is not able to access the prop value of the role attribute? Does the `getPropValue` function only returns the value and not the node of the prop?

ljharb:
It kind of seems like `getProp(attributes, 'role')` returns a JSXAttribute node, and `getPropValue` is returning "button", so you can't figure out that it's a conditional expression?

ljharb:
@V2dha i pushed up a commit that i think will fix it; what do you think?

I also filed #867 since the `allowExpressionValues` option is undocumented.

V2dha:
Thanks for the fix! Tests are passing locally now. I tried using `getProp` function earlier as well but it was throwing error while getting its type maybe because it was returning a `JSXAttribute` node.

ljharb (approved):

PR #871, opened 2022-07-08
Test cases passing for form with onSubmit

### LINKED ISSUE #323: form onSubmit trips jsx-a11y/no-noninteractive-element-interactions rule
eg.
```jsx
<form onSubmit={this.handleSubmit.bind(this)} method="POST">
    <button type="submit">
        Save
    </button>
</form>
```

This rule makes sense for a `div` or `span` etc., but not in the above case.

V2dha:
Added test cases that closes #323 of `form` tag with `onSubmit` handler


ljharb (approved):
Thanks!

PR #873, opened 2022-07-11
New rule: ambiguous-anchor-text

### LINKED ISSUE #866: New rule: `anchor-ambiguous-text`
One pattern that I've occasionally caught during code review is

```jsx
<a href="...">click here</a>
```

This is both ambiguous and often inaccessible for screenreader users ([this WebAIM page on links has more](https://webaim.org/techniques/hypertext/)). Similar issues exist for "link", "here", etc.

Do we think adding a rule to catch these errors is reasonable? I'm only familiar with this project as a user (and not a contributor), so I'm not sure if this aligns with the vision for this plugin. I also see an issue where this is more of a subjective rule (rather than a missing attribute), and would be hard to get properly without false positives. If there are other projects that are better suited for something like this, any pointers would be appreciated.

Happy to open a PR for a new rule if we think this is a good idea!

mattxwang:
This is a draft PR to add a new rule `ambiguous-anchor-text`. I've architected the bare minimum rule implementation (i.e. only trivial cases) and some documentation. I'm opening this as a draft as I want to get feedback on various areas (this is my first time contributing to this repo):

- I will likely eventually move `combineChildText` into its own file. before doing so:
  - thoughts on the name?
  - how would I know which AST nodes I should be "combining"? should I be calling this recursively on `JSXElement`s (I think I want to)? What about getting the `aria` text for various variables? this is something I have low visibility in, and am struggling to figure out where to look for more information on the AST.
- what other test cases should I add? One I'm looking to do is skipping screenreader-hidden elements; any other ones?
- how is the documentation? is it the right level of conciseness, clarity, etc.?
- how closely does this repo use flow? I'm not entirely sure what conventions I should be following

Closes #866.

### DIFF
--- a/__tests__/src/rules/anchor-ambiguous-text-test.js
+++ b/__tests__/src/rules/anchor-ambiguous-text-test.js
@@ -1,0 +1,97 @@
+/* eslint-env jest */
+/**
+ * @fileoverview Enforce `<a>` text to not exactly match "click here", "here", "link", or "a link".
+ * @author Matt Wang
+ */
+
+// -----------------------------------------------------------------------------
+// Requirements
+// -----------------------------------------------------------------------------
+
+import { RuleTester } from 'eslint';
+import parserOptionsMapper from '../../__util__/parserOptionsMapper';
+import rule from '../../../src/rules/anchor-ambiguous-text';
+
+// -----------------------------------------------------------------------------
+// Tests
+// -----------------------------------------------------------------------------
+
+const ruleTester = new RuleTester();
+
+const DEFAULT_AMBIGUOUS_WORDS = [
+  'click here',
+  'here',
+  'link',
+  'a link',
+  'learn more',
+];
+
+const expectedErrorGenerator = (words) => ({
+  message: `Ambiguous text within anchor. Screenreader users rely on link text for context; the words "${words.join('", "')}" are ambiguous and do not provide enough context.`,
+  type: 'JSXOpeningElement',
+});
+
+const expectedError = expectedErrorGenerator(DEFAULT_AMBIGUOUS_WORDS);
+
+ruleTester.run('anchor-ambiguous-text', rule, {
+  valid: [
+    { code: '<a>documentation</a>;' },
+    { code: '<a>${here}</a>;' },
+    { code: '<a aria-label="tutorial on using eslint-plugin-jsx-a11y">click here</a>;' },
+    { code: '<a><span aria-label="tutorial on using eslint-plugin-jsx-a11y">click here</span></a>;' },
+    {
+      code: '<a>click here</a>',
+      options: [{
+        words: ['disabling the defaults'],
+      }],
+    },
+    {
+      code: '<Link>documentation</Link>;',
+      settings: { 'jsx-a11y': { components: { Link: 'a' } } },
+    },
+    {
+      code: '<Link>${here}</Link>;',
+      settings: { 'jsx-a11y': { components: { Link: 'a' } } },
+    },
+    {
+      code: '<Link aria-label="tutorial on using eslint-plugin-jsx-a11y">click here</Link>;',
+      settings: { 'jsx-a11y': { components: { Link: 'a' } } },
+    },
+    {
+      code: '<Link>click here</Link>',
+      options: [{
+        words: ['disabling the defaults with components'],
+      }],
+      settings: { 'jsx-a11y': { components: { Link: 'a' } } },
+    },
+  ].map(parserOptionsMapper),
+  invalid: [
+    { code: '<a>here</a>;', errors: [expectedError] },
+    { code: '<a>HERE</a>;', errors: [expectedError] },
+    { code: '<a>click here</a>;', errors: [expectedError] },
+    { code: '<a>learn more</a>;', errors: [expectedError] },
+    { code: '<a>link</a>;', errors: [expectedError] },
+    { code: '<a>a link</a>;', errors: [expectedError] },
+    { code: '<a aria-label="click here">something</a>;', errors: [expectedError] },
+    { code: '<a> a link </a>;', errors: [expectedError] },
+    { code: '<a>a<i></i> link</a>;', errors: [expectedError] },
+    { code: '<a><i></i>a link</a>;', errors: [expectedError] },
+    { code: '<a><span>click</span> here</a>;', errors: [expectedError] },
+    { code: '<a><span> click </span> here</a>;', errors: [expectedError] },
+    { code: '<a><span aria-hidden>more text</span>learn more</a>;', errors: [expectedError] },
+    { code: '<a><span aria-hidden="true">more text</span>learn more</a>;', errors: [expectedError] },
+    { code: '<a><CustomElement>click</CustomElement> here</a>;', errors: [expectedError] },
+    {
+      code: '<Link>here</Link>',
+      errors: [expectedError],
+      settings: { 'jsx-a11y': { components: { Link: 'a' } } },
+    },
+    {
+      code: '<a>a disallowed word</a>',
+      errors: [expectedErrorGenerator(['a disallowed word'])],
+      options: [{
+        words: ['a disallowed word'],
+      }],
+    },
+  ].map(parserOptionsMapper),
+});
--- a/docs/rules/anchor-ambiguous-text.md
+++ b/docs/rules/anchor-ambiguous-text.md
@@ -1,0 +1,70 @@
+# anchor-ambiguous-text
+
+Enforces `<a>` values are not exact matches for the phrases "click here", "here", "link", "a link", or "learn more". Screenreaders announce tags as links/interactive, but rely on values for context. Ambiguous anchor descriptions do not provide sufficient context for users.
+
+## Rule details
+
+This rule takes one optional object argument with the parameter `words`.
+
+```json
+{
+  "rules": {
+    "jsx-a11y/anchor-ambiguous-text": [2, {
+      "words": ["click this"],
+    }],
+  }
+}
+```
+
+The `words` option allows users to modify the strings that can be checked for in the anchor text. Useful for specifying other words in other languages. The default value is set by `DEFAULT_AMBIGUOUS_WORDS`:
+
+```js
+const DEFAULT_AMBIGUOUS_WORDS = ['click here', 'here', 'link', 'a link', 'learn more'];
+```
+
+If an element has the `aria-label` property, its value is used instead of the inner text. Note that the rule still disallows ambiguous `aria-label`s. This rule also skips over elements with `aria-hidden="true"`.
+
+Note that this rule is case-insensitive and trims whitespace. It only looks for **exact matches**.
+
+### Succeed
+```jsx
+<a>read this tutorial</a> // passes since it is not one of the disallowed words
+<a>${here}</a> // this is valid since 'here' is a variable name
+<a aria-label="tutorial on using eslint-plugin-jsx-a11y">click here</a> // the aria-label supersedes the inner text
+```
+
+### Fail
+```jsx
+<a>here</a>
+<a>HERE</a>
+<a>click here</a>
+<a>link</a>
+<a>a link</a>
+<a> a link </a>
+<a><span> click </span> here</a> // goes through element children
+<a>a<i></i> link</a>
+<a><i></i>a link</a>
+<a><span aria-hidden="true">more text</span>learn more</a> // skips over elements with aria-hidden=true
+<a aria-label="click here">something</a> // the aria-label here is inaccessible
+```
+
+## Accessibility guidelines
+
+Ensure anchor tags describe the content of the link, opposed to simply describing them as a link.
+
+Compare
+
+```jsx
+<p><a href="#">click here</a> to read a tutorial by Foo Bar</p>
+```
+
+which can be more concise and accessible with
+
+```jsx
+<p>read <a href="#">a tutorial by Foo Bar</a></p>
+```
+
+### Resources
+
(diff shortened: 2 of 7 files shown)


src/rules/anchor-ambiguous-text.js
```diff
@@ -0,0 +1,83 @@
...
+  words: arraySchema,
+});
+
+const combineChildText = (node: JSXElement): string => node.children.reduce((str: string, child: Node) => {
+  switch (child.type) {
+    case 'Literal':
+      // $FlowFixMe child.value is too vague
+      return str + child.value;
+    // $FlowFixMe JSXText is missing in ast-types-flow
+    case 'JSXText':
```
gaurav5430, line 35: do we need to take care of JSXElement here ?

```
<a ...> Click <span> here </span> </a>
```

docs/rules/anchor-ambiguous-text.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Rule details
+
+This rule takes one optional object argument with the parameter `words`.
+
+```json
+{
+  "rules": {
+    "jsx-a11y/img-redundant-alt": [ 2, {
+      "words": [ "click this" ],
```
ljharb, line 13:
```suggestion
    "jsx-a11y/img-redundant-alt": [2, {
      "words": ["click this"],
```

docs/rules/anchor-ambiguous-text.md
```diff
@@ -0,0 +1,61 @@
...
+    }],
+  }
+}
+```
+
+The `words` option adds additional strings that can be checked for in the anchor text. Useful for specifying other words in other languages.
```
ljharb, line 19: why only additional? what if someone doesn't want the default words?

src/rules/anchor-ambiguous-text.js
```diff
@@ -0,0 +1,83 @@
...
+          words = [],
+        } = options;
+
+        const ambiguousWords = new Set(AMBIGUOUS_WORDS.concat(words));
+
+        if (!ambiguousWords.has(nodeText)) { // check the value
+          return;
+        }
+
+        context.report({
```
gaurav5430, line 76: just checking that the text content matches any of the ambiguous text might not be enough, as the author might have already catered for this by adding an aria-label which explains things more clearly.

src/index.js
```diff
@@ -51,6 +52,7 @@ module.exports = {
       },
       rules: {
         'jsx-a11y/alt-text': 'error',
+        'jsx-a11y/anchor-ambiguous-text': 'error',
```
ljharb, line 55: adding a new rule to a config is a breaking change, so we can't do that here.
```suggestion
        'jsx-a11y/anchor-ambiguous-text': 'off', // TODO: error
```

src/rules/anchor-ambiguous-text.js
```diff
@@ -0,0 +1,83 @@
...
+import getElementType from '../util/getElementType';
+
+const AMBIGUOUS_WORDS = [
+  'click here',
+  'here',
+  'link',
+  'a link',
+];
+
+const errorMessage = `Ambiguous text within anchor. Screenreader users rely on link text for context; the words ${AMBIGUOUS_WORDS.join(',')} are ambiguous and do not provide enough context.`;
```
ljharb, line 23: this error message won't contain the right words if the config specifies its own words

src/rules/anchor-ambiguous-text.js
```diff
@@ -0,0 +1,83 @@
...
+import type { ESLintConfig, ESLintContext } from '../../flow/eslint';
+import { arraySchema, generateObjSchema } from '../util/schemas';
+import getElementType from '../util/getElementType';
+
+const AMBIGUOUS_WORDS = [
+  'click here',
+  'here',
+  'link',
+  'a link',
+];
```
gaurav5430, line 21: we can allow passing this list as an option as well , and fallback to this as default ?

src/rules/anchor-ambiguous-text.js
```diff
@@ -0,0 +1,83 @@
...
+  switch (child.type) {
+    case 'Literal':
+      // $FlowFixMe child.value is too vague
+      return str + child.value;
+    // $FlowFixMe JSXText is missing in ast-types-flow
+    case 'JSXText':
+      return str + child.value;
+    default:
+      return str;
+  }
```
ljharb, line 39:
```suggestion
  if (child.type === 'Literal' || child.type === 'JSXText) {
    return str + child.value;
  }
  return str;
```
switch statements should be avoided :-)

ljharb (changes requested):

src/rules/anchor-ambiguous-text.js
```diff
@@ -0,0 +1,83 @@
...
+  'a link',
+];
+
+const errorMessage = `Ambiguous text within anchor. Screenreader users rely on link text for context; the words ${AMBIGUOUS_WORDS.join(',')} are ambiguous and do not provide enough context.`;
+
+const schema = generateObjSchema({
+  words: arraySchema,
+});
+
+const combineChildText = (node: JSXElement): string => node.children.reduce((str: string, child: Node) => {
```
gaurav5430, line 29: how are you able to use type annotations in js files ?

ljharb: this project uses flow.

src/rules/anchor-ambiguous-text.js
gaurav5430:
```
function getChildText(node) {
  const childText = node.children.reduce((prev, current, index, arr) => {
    switch (current.type) {
      case 'JSXText':
        return prev + current.value;
      case 'JSXElement':
        return prev + getChildText(current);
      default:
        return prev;
    }
  }, '');

  return childText;
}
```

src/index.js
mattxwang: Ah, I see. This portion was auto-generated with the new rule script (i.e. `./scripts/create-rule.js`), so maybe that needs to be updated too?

docs/rules/anchor-ambiguous-text.md
mattxwang: I based this off of `img-redundant-alt`, which has the same interface. I can certainly switch to letting the user override the defaults as well!

src/rules/anchor-ambiguous-text.js
mattxwang: I think we do, yup! The current `switch-case` (or eventual `if` statement given the other cr feedback) is not comprehensive for sure. Do you know if there's a list of all AST elements? I was trying to find one but wasn't particularly successful.

src/rules/anchor-ambiguous-text.js
mattxwang: Agreed, let me see if I can also check for that too!

src/rules/anchor-ambiguous-text.js
mattxwang: Agreed given feedback from @ljharb below as well.

src/rules/anchor-ambiguous-text.js
ljharb: An alternative is to use eslint's message substitution and sub in the words

src/index.js
ljharb: agreed

mattxwang:
Had a chance to also chat offline with some folks at CZI (including Andrew Huth!) about suggested improvements; want to document them here for posterity:

- let `aria-label` supersede internal text
- add `"learn more"` as a default ambiguous word, which in turn would adjust one of the "accepted" test cases to being a failure
- stretch goal: for elements with `aria-labelledby`, traverse children/related nodes to see if the `id`'d element exists and use its text. This may be too much of a performance hit/out of scope for the rule, and has problems if the labelledby element either doesn't exist or is not in the direct tree of the element.

@ljharb @gaurav5430 any thoughts? I'll try to wrap up implementing your suggestions as well sometime today/tomorrow, work schedule permitting.

cc: @dierat @jinlee93 @ahuth @jeremiah-clothier

mattxwang:
I think I've addressed the initial round of feedback (I'm not sure why the rebase action is failing - I've merged `main` into the branch). Haven't yet moved `getAccessibleChildText` into its own file yet, but I'll do that once we finalize the general operational feedback + write some specific tests for it.

Wanted to ask about:

- the `aria-labelledby` idea noted above
- should something like `<a><img alt="click here /></a>` be in scope of this rule?
- the general points of feedback I listed above

mattxwang:
Hi all, just wanted to see if y'all had any more feedback. If not, I'll move `getAccessibleChildText` into its own file, write a couple of tests, and then mark this as ready for review.

src/rules/anchor-ambiguous-text.js
```diff
@@ -0,0 +1,69 @@
...
+
+        if (!ambiguousWords.has(nodeText)) { // check the value
+          return;
+        }
+
+        const errorMessage = `Ambiguous text within anchor. Screenreader users rely on link text for context; the words "${words.join('", "')}" are ambiguous and do not provide enough context.`;
+
+        context.report({
+          node,
+          message: errorMessage,
```
ljharb, line 64:
```suggestion
        context.report({
          node,
          message: 'Ambiguous text within anchor. Screenreader users rely on link text for context; the words "{{wordsList}}" are ambiguous and do not provide enough context.',
          data: {
            wordsList: words.join('", "'),
          },
```

ljharb (commented):
This seems great overall.

Can we ensure that this works with the "components" setting? iow, if someone has mapped `CustomLink` to `a`, it should warn on CustomLink elements also.

mattxwang:
> Can we ensure that this works with the "components" setting? iow, if someone has mapped `CustomLink` to `a`, it should warn on CustomLink elements also.

@ljharb I think I've added a test case for that already, this is [lines 65-69](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/pull/873/files#diff-ac389cac751fa3edfe0e50ee6577beb824be48e326053c0194373fd23e633320R65-R69) in the test code:

```
    {
      code: '<Link>here</Link>',
      errors: [expectedError],
      settings: { 'jsx-a11y': { components: { Link: 'a' } } },
    },
```

Or did you mean something else?

src/rules/anchor-ambiguous-text.js
```diff
@@ -0,0 +1,70 @@
...
+        // Only check anchor elements and custom types.
+        if (typesToValidate.indexOf(nodeType) === -1) {
+          return;
+        }
+
+        const nodeText = getAccessibleChildText(node.parent, elementType);
+
+        const { words = DEFAULT_AMBIGUOUS_WORDS } = options;
+
+        const ambiguousWords = new Set(words);
```
ljharb, line 54: these, and the options extraction, should be done at the top level of `create` to avoid having to do so every time the node is visited.

__tests__/src/rules/anchor-ambiguous-text-test.js
```diff
@@ -0,0 +1,78 @@
...
+    { code: '<a><i></i>a link</a>;', errors: [expectedError] },
+    { code: '<a><span>click</span> here</a>;', errors: [expectedError] },
+    { code: '<a><span> click </span> here</a>;', errors: [expectedError] },
+    { code: '<a><span aria-hidden>more text</span>learn more</a>;', errors: [expectedError] },
+    { code: '<a><span aria-hidden="true">more text</span>learn more</a>;', errors: [expectedError] },
+    { code: '<a><CustomElement>click</CustomElement> here</a>;', errors: [expectedError] },
+    {
+      code: '<Link>here</Link>',
+      errors: [expectedError],
+      settings: { 'jsx-a11y': { components: { Link: 'a' } } },
```
ljharb, line 68: can you add a valid test using `components` as well?

src/util/getAccessibleChildText.js
```diff
@@ -0,0 +1,51 @@
...
+export default function getAccessibleChildText(node: JSXElement, elementType: (JSXOpeningElement) => string): string {
+  const ariaLabel = getLiteralPropValue(getProp(node.openingElement.attributes, 'aria-label'));
+  // early escape-hatch when aria-label is applied
+  if (ariaLabel) return standardizeSpaceAndCase(ariaLabel);
+
+  // skip if aria-hidden is true
+  if (isHiddenFromScreenReader(
+    elementType(node.openingElement),
+    node.openingElement.attributes,
+  )) return '';
```
ljharb, line 36:
```suggestion
  if (
    isHiddenFromScreenReader(
      elementType(node.openingElement),
      node.openingElement.attributes
    )
  ) {
    return '';
  }
```

ljharb (commented):
Almost there :-)

src/util/getAccessibleChildText.js
ljharb: We don't use prettier, thankfully :-) but yes, being able to disallow multiline constructs from not being on their own line would be a great rule that the airbnb config would adopt.

mattxwang:
Thanks for the feedback @ljharb, I think I've addressed all of it in the last two commits!

ljharb (approved):
Thanks!

mattxwang:
Just updated the branch from `main`. I'm not sure why the automatic rebase action is failing.

Re: "stretch" goal features (ex traversing the tree for `aria-labelledby`, `alt` text in images, etc.), should I open another issue to follow up? Certainly something I can do!

ljharb:
Don't worry about the rebase one :-)

Another issue (or PR, if you're ambitious) would be most appreciated.

mattxwang:
> Another issue (or PR, if you're ambitious) would be most appreciated.

Sounds good, and I'd be happy to at the very least write up an issue (I'll try my hand at a PR, but no promises). I'll aim to get one in over the next day or two!

Issue #879, opened 2022-07-22
Improve `ambiguous-anchor-text` false negatives (via extending `getAccessibleChildText`)


mattxwang:
I implemented `ambiguous-anchor-text` in #873, which uses a new util `getAccessibleChildText` to determine what counts as screenreader-accessible text inside an anchor tag. However, I think there are extra cases that would be valuable for the rule to catch.

## Current Approach

The current approach is simple: recursively add children, with base cases for:

- elements with an `aria-label`, returning the label - note that in this case, a label of "click here" **is still an error**
- elements that are hidden from a screenreader (via `isHiddenFromScreenReader`), returning the empty string
- the literal value for literals and text (`Literal` and `JSXText` types)

## Proposed Extensions

However, this misses some nuance in how screenreader text is interpreted, leading to false negatives. I have some specific examples that I think should also be in-scope, all of which seem reasonable to implement:

- `alt` text for `<img />`; this should behave like `aria-label`
- the SVG `<title>` element; this should also behave like `aria-label`

There's also a more challenging (but important) case of `aria-labelledby`. This is tricker since:

- we're not guaranteed that the `id` referenced by the attribute is a child of the anchor tag
- or that it exists at all!
- `aria-labelledby` [supports reference lists](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby#benefits_and_drawbacks), and **order matters**, though needs to ignore repeated IDs; chaining is disallowed

It also seems that [`aria-labelledby` overrides `aria-label`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-labelledby#benefits_and_drawbacks), so we should be sure to implement that.

I imagine that `aria-describedby` is out of scope, since it's a long-form *description* (and may not make sense for a screenreader to read).

What do we think? I don't have great familiarity in this field, so perhaps I am missing some nuance or other motivating examples. I may also split up the PRs so that it's easier to review (i.e. the low hanging fruit first).

mattxwang:
I've thought about this a bit more (and done some local testing) - I don't actually think the `<svg>` override is necessary, since `<title>` is parsed anyways as a child and there *should* not be any other text inside it.

[...]

PR #880, opened 2022-07-27
`anchor-ambiguous-text`: implements check for `alt` tags on `<img />` elements

mattxwang:
This is a (small) PR that extends `getAccessibleChildText` to return the `alt` value of an `` (or bound) component instead of traversing its children. As a result, this lets the utility and `anchor-ambiguous-text` handle anchors that have images as children.

I add a few test cases and update the docs.

Part of (but does not close) #879.


ljharb (approved):
Looks good. I'm a bit nervous this will break someone, and it would then need to be reverted and put behind an option, but we can give it a shot.

mattxwang:
> I'm a bit nervous this will break someone, and it would then need to be reverted and put behind an option, but we can give it a shot.

`anchor-ambiguous-text` wasn't released yet, right? In that case, this would be the "default" behaviour. I also (personally) don't anticipate many use-cases where people have a good reason to have `click here` or something of the sort.

ljharb:
aha, good point :-) yay!

PR #882, opened 2022-08-07
[Fix] expose `prefer-tag-over-role`

JoshuaKGoldberg:
Followup to #833, per https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/pull/833#issuecomment-1207484148.


src/index.js
```diff
@@ -191,6 +192,7 @@ module.exports = {
             ],
           },
         ],
+        'jsx-a11y/prefer-tag-over-role': 'error',
```
ljharb, line 195: actually it's a breaking change to add new rules to the configs, so we can't do that now.

ljharb (changes requested):

ljharb (approved):

PR #884, opened 2022-08-13
[readme] Preventing code repetition in user's eslint config file

Luccasoli:
A dev colleague added this plugin line even after extending the recommended configuration, so this PR is to make more explicit the no need to add the plugin


ljharb (approved):
Sure, why not

PR #886, opened 2022-08-24
[New] `anchor-ambiguous-text`: ignore punctuation

mattxwang:
I had a chance to test `anchor-ambiguous-text` in our internal repos @chanzuckerberg, and a coworker flagged a **false negative**:

```jsx
<a>learn more.</a>
```

The above should be flagged (it's ambiguous anchor text!), but because the rule doesn't ignore punctuation, it doesn't hit the existing dictionary.

So, this is a bite-sized PR that:

- adds failing test cases
- adjusts `getAccessibleChildText` to strip out punctuation
- documents the new behavioural change


src/util/getAccessibleChildText.js
```diff
@@ -10,10 +10,14 @@ import isHiddenFromScreenReader from './isHiddenFromScreenReader';
...
  * and the string is lowercase
  * @param {string} input
- * @returns lowercase, single-spaced, trimmed string
+ * @returns lowercase, single-spaced, punctuation-stripped, trimmed string
  */
 function standardizeSpaceAndCase(input: string): string {
-  return input.trim().replace(/\s\s+/g, ' ').toLowerCase();
+  return input
+    .trim()
+    .replace(/[,.?!;:]/g, '') // strip punctuation
```
ljharb, line 18:
```suggestion
    .replace(/[,.?¿!‽¡;:]/g, '') // strip punctuation
```
we should probably catch a few others

ljharb (changes requested):
This seems like one of those bug fixes that might also be a breaking change :-/ i'd hate to put it behind an option, but that might be the safest approach?

src/util/getAccessibleChildText.js
mattxwang: Definitely agreed with the commit suggestion;

> This seems like one of those bug fixes that might also be a breaking change

We haven't released this rule, so I'm thinking that there wouldn't be a breaking change? Or do you mean that we should add a toggle for user convenience?

ljharb: oh, ha, good point :-) in that case, let's do it

mattxwang:
Thanks for the very quick review!

> This seems like one of those bug fixes that might also be a breaking change

We haven't released this rule, so I'm thinking that there wouldn't be a breaking change? Or do you mean that we should add a toggle for user convenience?

ljharb (approved):
looks good

Issue #881, opened 2022-07-28
New rule: Don't allow `aria-hidden` on focusable elements


khiga8:
Hi team. I'd like to propose a rule that flags when `aria-hidden="true'` is set on focusable elements or on elements with descendants that are focusable.  The rationale is that `aria-hidden` should be reserved for elements that are non-interactive since assistive technology users can still reach interactive elements in keyboard even with `aria-hidden="true"`. Setting `aria-hidden="true"` on focusable elements can lead to unexpected behavior or be confusing.

An exception for this rule could be if a `tabindex="-1"` is set to remove it from keyboard order or if it's an input element with `disabled`. (While this could be an exception, I don't think this is necessarily a "good" approach since it's questionable whether interactive elements should be considered decorative, to begin with)

I would consider the following as focusable elements: `textarea`, `button`, `a`, `input`, `summary`. 

What do you think? Am I missing anything?

I'm happy to start a draft PR.

PR #888, opened 2022-08-26
[New rule] disallow `aria-hidden="true"` on focusable elements

khiga8:
Fixes: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/881

This adds a new rule that flags when `aria-hidden="true"` is set on focusable elements ~~or contain child elements that are focusable~~. People often set `aria-hidden="true"` to hide an element from AT users, but if the element is focusable, AT users can still access it which can be confusing.

I wasn't sure what the contributing process is like so wanted to get a draft PR going to get thoughts. I looked through the list of jsx-a11y rules and I couldn't find a rule for something like this but let me know if I missed it. If maintainers are okay moving forward with this, I will add a doc page. I've ran the lint and added a test.

Related references: 
- [aria-hidden elements do not contain focusable elements](https://dequeuniversity.com/rules/axe/html/4.4/aria-hidden-focus)
- [Element with aria-hidden has no content in sequential focus navigation](https://www.w3.org/WAI/standards-guidelines/act/rules/6cfa84/proposed/)
- [MDN aria-hidden](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden) warning: `Do not use aria-hidden="true" on focusable elements.`

### DIFF
--- a/__tests__/src/rules/no-aria-hidden-on-focusable-test.js
+++ b/__tests__/src/rules/no-aria-hidden-on-focusable-test.js
@@ -1,0 +1,42 @@
+/**
+ * @fileoverview Enforce `aria-hidden="true"` is not used on focusable elements.
+ * @author Kate Higa
+*/
+
+// -----------------------------------------------------------------------------
+// Requirements
+// -----------------------------------------------------------------------------
+
+import { RuleTester } from 'eslint';
+import parserOptionsMapper from '../../__util__/parserOptionsMapper';
+import rule from '../../../src/rules/no-aria-hidden-on-focusable';
+// -----------------------------------------------------------------------------
+// Tests
+// -----------------------------------------------------------------------------
+
+const ruleTester = new RuleTester();
+
+const expectedError = {
+  message: 'aria-hidden="true" must not be set on focusable elements.',
+  type: 'JSXOpeningElement',
+};
+
+ruleTester.run('no-aria-hidden-on-focusable', rule, {
+  valid: [
+    { code: '<div aria-hidden="true" />;' },
+    { code: '<div onClick={() => void 0} aria-hidden="true" />;' },
+    { code: '<img aria-hidden="true" />' },
+    { code: '<a aria-hidden="false" href="#" />' },
+    { code: '<button aria-hidden="true" tabIndex="-1" />' },
+    { code: '<button />' },
+    { code: '<a href="/" />' },
+  ].map(parserOptionsMapper),
+  invalid: [
+    { code: '<div aria-hidden="true" tabIndex="0" />;', errors: [expectedError] },
+    { code: '<input aria-hidden="true" />;', errors: [expectedError] },
+    { code: '<a href="/" aria-hidden="true" />', errors: [expectedError] },
+    { code: '<button aria-hidden="true" />', errors: [expectedError] },
+    { code: '<textarea aria-hidden="true" />', errors: [expectedError] },
+    { code: '<p tabindex="0" aria-hidden="true">text</p>;', errors: [expectedError] },
+  ].map(parserOptionsMapper),
+});
--- a/src/rules/no-aria-hidden-on-focusable.js
+++ b/src/rules/no-aria-hidden-on-focusable.js
@@ -1,0 +1,44 @@
+/**
+ * @fileoverview Enforce aria-hidden is not used on interactive elements or contain interactive elements.
+ * @author Kate Higa
+ */
+
+// ----------------------------------------------------------------------------
+// Rule Definition
+// ----------------------------------------------------------------------------
+
+import { getProp, getPropValue } from 'jsx-ast-utils';
+import getElementType from '../util/getElementType';
+import isFocusable from '../util/isFocusable';
+import { generateObjSchema } from '../util/schemas';
+
+const errorMessage = 'aria-hidden="true" must not be set on focusable elements.';
+const schema = generateObjSchema();
+
+export default {
+  meta: {
+    docs: {
+      url: 'https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/tree/HEAD/docs/rules/no-aria-hidden-on-focusable.md',
+      description: errorMessage,
+    },
+    schema: [schema],
+  },
+
+  create(context) {
+    const elementType = getElementType(context);
+    return {
+      JSXOpeningElement(node) {
+        const { attributes } = node;
+        const type = elementType(node);
+        const isAriaHidden = getPropValue(getProp(attributes, 'aria-hidden')) === true;
+
+        if (isAriaHidden && isFocusable(type, attributes)) {
+          context.report({
+            node,
+            message: errorMessage,
+          });
+        }
+      },
+    };
+  },
+};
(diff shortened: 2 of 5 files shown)


src/rules/no-aria-hidden-on-focusable.js
```diff
@@ -0,0 +1,65 @@
...
+      url: 'https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/tree/HEAD/docs/rules/no-aria-hidden-on-focusable.md',
+      description: 'aria-hidden="true" must not be set on focusable elements or on elements which contain a focusable element.',
+    },
+    schema: [schema],
+  },
+
+  create: (context) => {
+    const elementType = getElementType(context);
+    return {
+      JSXOpeningElement: (node) => {
```
ljharb, line 36:
```suggestion
      JSXOpeningElement(node) {
```

src/rules/no-aria-hidden-on-focusable.js
```diff
@@ -0,0 +1,65 @@
...
+export default {
+  meta: {
+    docs: {
+      url: 'https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/tree/HEAD/docs/rules/no-aria-hidden-on-focusable.md',
+      description: 'aria-hidden="true" must not be set on focusable elements or on elements which contain a focusable element.',
+    },
+    schema: [schema],
+  },
+
+  create: (context) => {
```
ljharb, line 33:
```suggestion
  create(context) {
```

__tests__/src/rules/no-aria-hidden-on-focusable-test.js
```diff
@@ -0,0 +1,41 @@
...
+    // { code: '<button tabIndex="-1" />' },
+    // { code: '<button />' },
+    // { code: '<a href="/" />' },
+  ].map(parserOptionsMapper),
+  invalid: [
+    // { code: '<div aria-hidden="true" tabIndex="0" />;', errors: [expectedError] },
+    // { code: '<input aria-hidden="true" />;', errors: [expectedError] },
+    // { code: '<a href="/" aria-hidden="true" />', errors: [expectedError] },
+    // { code: '<button aria-hidden="true" />', errors: [expectedError] },
+    // { code: '<textarea aria-hidden="true" />', errors: [expectedError] },
```
ljharb, line 38: can these be uncommented?

ljharb (commented):
Seems good overall.

__tests__/src/rules/no-aria-hidden-on-focusable-test.js
khiga8: Ah I was trying to debug a test. sorry about that.

khiga8:
Added more test cases and fixed a bug in [bc220ad](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/pull/888/commits/bc220adf8efe3f594d1375573b84f0c64fcf2b9a).

Added a doc page in [e96c9db](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/pull/888/commits/e96c9db87f52d17151c7718202c256e871b3202d).

src/rules/no-aria-hidden-on-focusable.js
```diff
@@ -0,0 +1,66 @@
...
+import isInteractiveElement from '../util/isInteractiveElement';
+import { generateObjSchema } from '../util/schemas';
+
+const errorMessage = 'aria-hidden="true" must not be set on focusable elements or on elements which contain a focusable element.';
+const schema = generateObjSchema();
+
+/* Returns true when element is interactive and does not have `tabIndex=-1`, or when an element has a tabIndex that isn't -1. */
+function isFocusable(type, attributes) {
+  const tabIndex = getTabIndex(getProp(attributes, 'tabIndex'));
+  return (tabIndex !== -1 && isInteractiveElement(type, attributes)) || (tabIndex !== undefined && tabIndex !== -1);
```
ljharb, line 22:
```suggestion
  return tabIndex !== -1 && (
    tabIndex !== undefined
    || isInteractiveElement(type, attributes)
  );
```

src/rules/no-aria-hidden-on-focusable.js
```diff
@@ -0,0 +1,66 @@
...
+            const childType = elementType(openingChildElement);
+            const childAttributes = openingChildElement.attributes;
+            if (isFocusable(childType, childAttributes)) {
+              context.report({
+                node: openingChildElement,
+                message: errorMessage,
+              });
+            }
+          }
+        });
```
ljharb, line 62:
```suggestion
        children.filter((child) => {
          if (child.type !== 'JSXElement' || !child.openingElement) {
            return false;
          }
          const openingChildElement = child.openingElement;
          const childType = elementType(openingChildElement);
          const childAttributes = openingChildElement.attributes;
          return isFocusable(childType, childAttributes);
        }).forEach((child) => {
          context.report({
            node: child.openingElement,
            message: errorMessage,
          });
        });
```

src/rules/no-aria-hidden-on-focusable.js
```diff
@@ -0,0 +1,70 @@
...
+        const ariaHidden = getPropValue(getProp(attributes, 'aria-hidden')) === true;
+
+        if (!ariaHidden) return;
+        if (isFocusable(type, attributes)) {
+          context.report({
+            node,
+            message: errorMessage,
+          });
+        }
+        const { children } = node.parent;
```
jessebeach, line 52: Generally, you want to use ESLint's visitor pattern to visit nodes, rather than doing your own tree traversal within a rule down into child nodes. What is purpose of creating reports for focusable children of the opening node?

jessebeach (changes requested):
@khiga8 this is a great idea for a rule. You're off to a solid start. I'd love to move your `isFocusable` function to its own utility. And I have a question about the reason behind traversing child elements. 

I think you can get this to a place where we can get it committed. I look forward to the updates!

src/rules/no-aria-hidden-on-focusable.js
khiga8: Thank you for pointing this out. I'm new to ESLint rule creation so uh.. please excuse my non-idiomatic code 🙇‍♀️.

My intention had been to cover cases where `aria-hidden` is set on an element that contains focusable children like:
```.jsx
<div aria-hidden="true">
   <a href="/">Some site</a>
</div>
```

Is this something that can be accomplished with the visitor pattern? I'll have to do a bit more digging into ESLint conventions.

jessebeach: > My intention had been to cover cases where aria-hidden is set on an element that contains focusable children like:

Ok, I understand. For this rule, I'm going to recommend that you keep it narrowly focused to just focusable elements and not their ancestors or descendants. The problem of the cascading effects of aria-hidden on the AX tree is a thorny one and it will complicate a rule that is otherwise straightforward and easy to reason about. Plus, these sorts of tree relationships are hard to lint for, because you're only getting a snippet of the total render tree while linting, so that incidence of positive cases is going to be low with this static linting method. The reason I really like your rule is that it can be narrowly focused to focusable elements and there is good literature about how `aria-hidden` should interact with these.

jessebeach: > please excuse my non-idiomatic code 🙇‍♀️

No need for apologies :)

jessebeach (approved):
This is amazing @khiga8! Thank you so much for proposing this rule and going through the iterations to develop it. Let's get it landed!

jessebeach:
@ljharb ready to land

ljharb (approved):

PR #891, opened 2022-09-13
feat: support new config system

jjangga0214:
Supporting eslint's new config system of eslint.

Note that the legacy config system always has `require()`d plugins and sharable configs, whereas the new system is available with both of ESM and CJS.

Thus [conditional export](https://nodejs.org/api/packages.html#conditional-exports) is introduced to keep compatibility while providing a new config/plugin at the same time.

`plugin:` protocol(e.g. `plugin:react-hooks/recommended`) is not valid any more. Thus the new plugin doesn't have `configs`.

Blocker for #978.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,8 @@ yarn add eslint-plugin-jsx-a11y --dev
 
 **Note:** If you installed ESLint globally (using the `-g` flag in npm, or the `global` prefix in yarn) then you must also install `eslint-plugin-jsx-a11y` globally.
 
-## Usage
+<a id="usage"></a>
+## Usage (legacy: `.eslintrc`)
 
 Add `jsx-a11y` to the plugins section of your `.eslintrc` configuration file. You can omit the `eslint-plugin-` prefix:
 
@@ -107,6 +108,124 @@ configuration file by mapping each custom component name to a DOM element type.
     }
   }
 }
+```
+
+## Usage (new: `eslint.config.js`)
+
+From [`v8.21.0`](https://github.com/eslint/eslint/releases/tag/v8.21.0), eslint announced a new config system.
+In the new system, `.eslintrc*` is no longer used. `eslint.config.js` would be the default config file name.
+
+And from [`v8.23.0`](https://github.com/eslint/eslint/releases/tag/v8.23.0), eslint CLI starts to look up `eslint.config.js`.
+**So, if your eslint is `>=8.23.0`, you're 100% ready to use the new config system.**
+
+You might want to check out the official blog posts,
+
+- <https://eslint.org/blog/2022/08/new-config-system-part-1/>
+- <https://eslint.org/blog/2022/08/new-config-system-part-2/>
+- <https://eslint.org/blog/2022/08/new-config-system-part-3/>
+
+and the [official docs](https://eslint.org/docs/latest/user-guide/configuring/configuration-files-new).
+
+The default export of `eslint-plugin-jsx-a11y` is a plugin object.
+
+```js
+const jsxA11y = require('eslint-plugin-jsx-a11y');
+
+module.exports = [
+  …
+  {
+    files: ['**/*.{js,mjs,cjs,jsx,mjsx,ts,tsx,mtsx}'],
+    plugins: {
+      'jsx-a11y': jsxA11y,
+    },
+    languageOptions: {
+      parserOptions: {
+        ecmaFeatures: {
+          jsx: true,
+        },
+      },
+    },
+    rules: {
+      // ... any rules you want
+      'jsx-a11y/alt-text': 'error',
+    },
+    // ... others are omitted for brevity
+  },
+  …
+];
+```
+
+There're also 2 shareable configs.
+
+- `eslint-plugin-jsx-a11y/strict`
+- `eslint-plugin-jsx-a11y/recommended`
+
+If your eslint.config.js is ESM, include the `.js` extension (e.g. `eslint-plugin-jsx-a11y/recommended.js`). Note that the next semver-major will require omitting the extension for these imports.
+
+**Note**: These configurations will import `eslint-plugin-jsx-a11y` and enable JSX in [`languageOptions.parserOptions`](https://eslint.org/docs/latest/user-guide/configuring/configuration-files-new#configuration-objects).
+
+In the new config system, the `plugin:` protocol(e.g. `plugin:jsx-a11y/recommended`) is no longer valid.
+As eslint does not automatically import the preset config (shareable config), you explicitly do it by yourself.
+
+```js
+const jsxA11yRecommended = require('eslint-plugin-jsx-a11y/recommended');
+
+export default [
+  …
+  jsxA11yRecommended, // This is not a plugin object, but a shareable config object
+  …
+];
+```
+
+You can of course add/override properties.
+
+**Note**: Our shareable configs does not preconfigure `files` or [`languageOptions.globals`](https://eslint.org/docs/latest/user-guide/configuring/configuration-files-new#configuration-objects).
+For most of the cases, you probably want to configure some properties by yourself.
+
+```js
+const jsxA11yRecommended = require('eslint-plugin-jsx-a11y/recommended');
+const globals = require('globals');
+
+module.exports = [
+  …
+  {
+    files: ['**/*.{js,mjs,cjs,jsx,mjsx,ts,tsx,mtsx}'],
+    ...jsxA11yRecommended,
+    languageOptions: {
+      ...jsxA11yRecommended.languageOptions,
+      globals: {
+        ...globals.serviceworker,
+        ...globals.browser,
+      },
+    },
+  },
+  …
+];
+```
+
+The above example is same as the example below, as the new config system is based on chaining.
+
+```js
+const jsxA11yRecommended = require('eslint-plugin-jsx-a11y/recommended');
+const globals = require('globals');
+
+module.exports = [
+  …
+  {
+    files: ['**/*.{js,mjs,cjs,jsx,mjsx,ts,tsx,mtsx}'],
+    ...jsxA11yRecommended,
+  },
+  {
+    files: ['**/*.{js,mjs,cjs,jsx,mjsx,ts,tsx,mtsx}'],
+    languageOptions: {
+      globals: {
+        ...globals.serviceworker,
+        ...globals.browser,
+      },
+    }
+  },
+  …
+];
 ```
 
 ## Supported Rules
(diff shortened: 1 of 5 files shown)


ljharb (commented):
I understand that the shared configs provided by this plugin need to be in their own entrypoints to work with the new system.

However, why can't the existing "main" Just Work with the new config system? What's in `legacy.js` that's different from `index.js`?

package.json
```diff
@@ -16,6 +16,15 @@
...
   "main": "lib/index.js",
+  "exports": {
+    ".": {
+      "import": "./lib/index.js",
+      "require": "./lib/legacy.js"
+    },
+    "./new": "./lib/index.js",
+    "./strict": "./lib/configs/strict.js",
+    "./recommended": "./lib/configs/recommended.js"
+  },
```
ljharb, line 27: adding "exports" is always a breaking change, so this is a nonstarter.

jjangga0214: It becomes non-breaking if we do so. 
What do you think?

```js
{
  ".": {
      "import": "./lib/index.js",
      "require": "./lib/legacy.js"
    },
  "./new": "./lib/index.js",
  "./strict": "./lib/configs/strict.js",
  "./recommended": "./lib/configs/recommended.js"
  "./*": "./*.js",
  "./*.js": "./*.js",
  "./package.json": "./package.json"
}
```

jjangga0214:
> However, why can't the existing "main" Just Work with the new config system? What's in `legacy.js` that's different from `index.js`?

I have never told the existing "main" just doesn't work. It works as the only difference is the existence of `configs` property, and the new config system would ignore it if the plugin object still has it. But my point was that it's not the intended specification, and can confuse users, leading to weird usage.

package.json
ljharb: I'm not convinced that it would be - for example, that breaks `./package`, and `./lib`, and a few others. Enumerating them all isn't really worth it if the "main" already works in the new system.

ljharb:
I'm not sure why a user would be confused, and if they are, I think that's on eslint to fix, not individual plugins in the ecosystem.

package.json
jjangga0214: I'd respect your philosophy, but I think it's quite exhaustive.

> that breaks ./package
You're right.

> and ./lib
Almost all users would `require('eslint-plugin-jsx-a11y')`, which is the same but more standard. 
Isn't `require('eslint-plugin-jsx-a11y/lib')` weird? It's a non-intended way of consuming the package.

> and a few others.
I don't see what you're saying. I think there's no more to include. 
What else? `./docs`? No, who would depend on it..? `./scripts`? Neither..

> Enumerating them all isn't really worth it
I see. I think it's a matter of preference of a maintainer, and I'd respect that.

jjangga0214:
> I'm not sure why a user would be confused, and if they are, I think that's on eslint to fix, not individual plugins in the ecosystem.

Okay. I respect that. I'd patch as long as [this](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/pull/891#discussion_r970420725) is resolved.

package.json
jjangga0214: Apart from [conditional export](https://nodejs.org/api/packages.html#conditional-exports), another purpose of introducing `exports` is for elegance.

For example, `eslint-plugin-jsx-a11y/recommended` is much more elegant and intuitive than `'eslint-plugin-jsx-a11y/lib/configs/recommended.js`.

ljharb: The files field should never ever be used - it’s dangerous. Only npmignore should be used to restrict published files.

I’m quite content to add the new entrypoints for the exported configs - that seems perfectly reasonable. I’m also quite familiar with how exports works, since i helped design the feature - but changing the file structure is only needed when you did it wrong in the first place, so we’re fine here :-)

jjangga0214: Okay, please let me make sure I understood correctly.

> I’m quite content to add the new entrypoints.

- We should add new entrypoints for shareable configs

> adding "exports" is always a breaking change,

- But "exports" should be avoided.

> Enumerating them all isn't really worth it

- Because the `exports` object would be unnecessarily too complex, in order to prevent breaking change.

Therefore, without `exports`, the end-user should do 

```
import jsxA11y from 'eslint-plugin-jsx-a11y/lib/configs/recommended.js'
// Or
import jsxA11y from 'eslint-plugin-jsx-a11y/lib/recommended.js'
```

instead of

```
import jsxA11y from 'eslint-plugin-jsx-a11y/recommended'
```

Would this behavior be the final decision?

ljharb: "exports" shouldn't be avoided in general - it's an awesome thing! But, breaking changes should be avoided. If we ever happen to do a semver-major, we'd add "exports" at that time regardless.

I'd expect to add a top-level `recommended.js` file that requires and re-exports from lib/configs/recommended.js (and the same for the others), so that everyone would import `eslint-plugin-jsx-a11y/recommended`.

jjangga0214:
Pushed a new commit :)

ljharb:
why do you need to bring in your own globals? isn't that what `env` is for?

ljharb (approved):
I fixed up the code samples, added the missing mjs/cjs extensions (native ESM should always be .mjs, not .js, so it's critical to allow that), and converted the CJS files you added in `src/` to ESM.

I'm dismayed that eslint's new config system removes the `env` key; users shouldn't have to import globals themselves, especially if they'd have to depend directly on the `globals` package.

ljharb:
hmm, i'm having second thoughts on the `files` config. @jjangga0214, if `files: ['**/*.{js,mjs,cjs,jsx,mjsx,ts,tsx,mtsx}']` is included by default in the *legacy* config system, will that break anything for them?

jjangga0214:
> if files: ['**/*.{js,mjs,cjs,jsx,mjsx,ts,tsx,mtsx}'] is included by default in the legacy config system, will that break anything for them?

Well, my opinion is yes, ***if*** so. 
But we know the assumption is not true, so what did make you think about the imagination?

ljharb:
I didn't understand your reply.

What I realized is that it's critical for all of these extensions to be included, and highly unlikely devs will remember to include them all, so it's valuable to provide them by default. However, I want the plugin object to be able to be usable in both config systems.

jjangga0214:
It seems there's confusion between us?

> if files: ['**/*.{js,mjs,cjs,jsx,mjsx,ts,tsx,mtsx}'] is included by default in the legacy config system,

To my understanding, `files: ['**/*.{js,mjs,cjs,jsx,mjsx,ts,tsx,mtsx}']` is **NOT** included by default in the legacy config system. Am I wrong?

ljharb:
@jjangga0214 what i mean is, in the legacy config, is the presence of a "files" property problematic? or is it ignored?

jjangga0214:
I don't understand the point. Legacy config should not be used in the new config. They're just different and incompatible. The exception is when the legacy config only has `.rules`.
Other fields like `.plugins`(conflict. Legacy: `string[]`, New: `PluginObject[]`),`.extends`(ignored), `.overrides` (ignored), `.parser` (ignored), `.parserOptions` (ignored), `.env` (ignored) and more, are incompatible. For almost all plugins in the ecosystem, including `eslint-plugin-jsx-a11y`, their presets have at least `.plugins` or `.extends`.

So I still don't grasp what your concern is. 
I think I can understand if you can show me an example of your thoughts. Then I'd be able to easily get the scenario.

ljharb:
"Different and incompatible" is what I'm trying to confirm.

If in fact the same object can't serve in both configs, then that's a massive design mistake on eslint's side - and I'd prefer to impose the cost on the users of the new config, since the legacy config is the most important (and will remain so for a good number of years).

jjangga0214:
> I'd prefer to impose the cost on the users of the new config

Does this mean you're not supposed to support the new config system?

ljharb:
@jjangga0214 oh no, not at all! it just means the main entrypoint of the package will be the legacy config, and flat config users will need to import, eg, `eslint-plugin-jsx-a11y/flat` or something.

However, if the same object can be used in both config systems, then the main entrypoint can support both, which is the ideal.

I've just rebased this, and if tests pass, and we can confirm that the main entrypoint works in both configs, then it'd be good to go :-)

pauliesnug:
@ljharb is this being maintained? the last update was four months ago when you said it would be merged

ljharb:
@pauliesnug of course it is. I'm waiting for confirmation that the plugin works in both config formats.

jjangga0214:
@ljharb @pauliesnug 
> oh no, not at all! it just means the main entrypoint of the package will be the legacy config, and flat config users will need to import, eg, eslint-plugin-jsx-a11y/flat or something.

Okay, I now understand it.

> if the same object can be used in both config systems, 

Unfortunately, no.

https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/1adec3517fc2c9797212ca4d38858deed917e7be/src/index.js#L47-L54

For example, as I've said before,  `plugins` as `string[]` is incompatible, and `parserOption` should be under `languageOptions`.

So in conclusion, we should provide a different object by a different entry point.
Any idea about the name of it?

ljharb:
Just an example, but yes, it'd need to be some other entry point.

anselmbradford:
> Any idea about the name of it?
 
FWIW `'flat/recommended'` is what eslint-plugin-jsdoc names it https://www.npmjs.com/package/eslint-plugin-jsdoc#flat-config

hanselabreu:
+1 on this. This would be useful for importing recommended/strict settings via CLI (e.g. `eslint --config node_modules/eslint-plugin-jsx-a11y/recommended.js`). Which is currently impossible since the current `index.js` contains additional configurations.

ljharb:
@soerenmartius please don't open a new PR; if you want to rebase this PR and keep working on it, please comment a branch link and i'll pull in your changes.

soerenmartius:
> @soerenmartius please don't open a new PR; if you want to rebase this PR and keep working on it, please comment a branch link and i'll pull in your changes.

Thanks - I am on PTO this and next week but will try to contribute once I am back unless you have time to finish this until then! Thanks

michaelfaith:
Is this still being worked on?  I might be able to help out, if nobody's working on it.  Really eager for the v9 & flat config support.

ljharb:
@michaelfaith if you have updates, please comment a link to a branch or sha (not a new PR) and i'll pull in the changes.

michaelfaith:
@ljharb I know you mentioned not creating  new PR, but I went in a totally different direction with this, in order to address some of your concerns, so I opened a new PR: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/pull/993

Totally cool if you'd rather bring those changes into this PR, but figured I'd leave that up to you.  With my changes, we can use the same `main` entry point for both legacy and flat experiences, without the need for a separate entry point.  
I also added a new `examples` folder in the root that has three different (vite-based) React apps using the built plugin, testing usage in both legacy (eslintrc) and flat (esm and cjs) configs.

ljharb:
Closing in favor of #993.

PR #893, opened 2022-09-15
build: add items to .npmignore

jjangga0214:
Currently unnecessary files and directories are included in the published package.


ljharb (changes requested):
These are all intentionally included; `npm explore foo && npm install && npm test` should always work.

We could exclude CONTRIBUTING.md, and .github/workflows, i suppose

ljharb (approved):
I've replaced this with a commit that adds automatic `npmignore` generation, and removes the few extra files we can get away with.

PR #898, opened 2022-10-11
[Tests] Integrate parsers.all function from `eslint-plugin-react`

ROSSROSALES:
This PR is to integrate parsers.all from eslint-plugin-react into each rule

### DIFF
--- a/__tests__/__util__/helpers/parsers.js
+++ b/__tests__/__util__/helpers/parsers.js
@@ -1,0 +1,186 @@
+import path from 'path';
+import semver from 'semver';
+import entries from 'object.entries';
+import { version } from 'eslint/package.json';
+import flatMap from 'array.prototype.flatmap';
+
+let tsParserVersion;
+try {
+  // eslint-disable-next-line import/no-unresolved, global-require
+  tsParserVersion = require('@typescript-eslint/parser/package.json').version;
+} catch (e) { /**/ }
+
+const disableNewTS = semver.satisfies(tsParserVersion, '>= 4.1') // this rule is not useful on v4.1+ of the TS parser
+  ? (x) => ({ ...x, features: [].concat(x.features, 'no-ts-new') })
+  : (x) => x;
+
+function minEcmaVersion(features, parserOptions) {
+  const minEcmaVersionForFeatures = {
+    'class fields': 2022,
+    'optional chaining': 2020,
+    'nullish coalescing': 2020,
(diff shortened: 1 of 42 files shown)


__tests__/__util__/helpers/parsers.js
```diff
@@ -0,0 +1,181 @@
...
+
+const path = require('path');
+const semver = require('semver');
+const entries = require('object.entries');
+const version = require('eslint/package.json').version;
+const flatMap = require('array.prototype.flatmap');
+const tsParserVersion = require('@typescript-eslint/parser/package.json').version;
+
+const disableNewTS = semver.satisfies(tsParserVersion, '>= 4.1') // this rule is not useful on v4.1+ of the TS parser
+  ? (x) => Object.assign({}, x, { features: [].concat(x.features, 'no-ts-new') })
```
ljharb, line 11:
```suggestion
  ? (x) => { ...x, features: [].concat(x.features, 'no-ts-new') }
```
since in this repo we have a build process :-)

ljharb:
LGTM so far!

ROSSROSALES:
`@ROSSROSALES
[integrated parsers.all into majority of Test Cases]`  `37b2d55`

Above excludes test file `anchor-is-valid.js`
Debugging Error in Invalid tests

__tests__/src/rules/alt-text-test.js
```diff
@@ -28,19 +30,29 @@ Use alt="" for presentational images.`,
...
 });
 
-const ariaLabelValueError = 'The aria-label attribute must have a value. The alt attribute is preferred over aria-label for images.';
-const ariaLabelledbyValueError = 'The aria-labelledby attribute must have a value. The alt attribute is preferred over aria-labelledby for images.';
+const ariaLabelValueError = {
+  message: 'The aria-label attribute must have a value. The alt attribute is preferred over aria-label for images.'
+};
+const ariaLabelledbyValueError = {
+  message: 'The aria-labelledby attribute must have a value. The alt attribute is preferred over aria-labelledby for images.'
+};
```
ljharb, line 38: rather than this, we should make sure the helper can handle a string message.

check the react plugin, and the import plugin's parsers branch; i think this case is already handled in at least one of those places.

ljharb:
It's expected that coverage didn't go up, because this project only has one parser at the moment, but it should make it easier to add another.

PR #837, opened 2022-02-28
[Docs] Add infrastructure for auto-generating markdown table and list

uncommon-type:
## Description

This PR adds infrastructure to keep a11y rules in the `/rules` folder in sync with the `README` markdown content. It generates markdown list and table programmatically and makes sure their contents are kept up-to-date.

## Related Issue

Closes #836  

## Acceptance Criteria

- [x] The following has been added as a project dependency: `markdown-magic`
- [x] `README` list and table are generated programmatically
- [x] `README` list and table are kept in sync with a11y rules in the /rules folder. Every time a new a11y rule is added to /rules folder , `README` list and table contents are refreshed programmatically 

## Type of Changes

|     | Type                       |
| --- | -------------------------- |
|    | :bug: Bug fix              |
|    | :sparkles: New feature     |
|    | :hammer: Refactoring       |
|    | :100: Add tests            |
|  ✓  | :link: Update dependencies |
|  ✓  | :scroll: Docs              |

## Testing Steps / QA Criteria
* From your terminal, pull down this branch
* `npm install` to install project dependencies 
* Delete the contents of the `README` table and list but keep in place the `AUTO-GENERATED-CONTENT` markers telling `markdown-magic` to generate table and list contents
* Then `npm generate-list-of-rules` to auto-generate the table and list contents
* Add a new a11y rule to the /rules folder , run `npm generate-list-of-rules` to refresh the table and list contents

PR #901, opened 2022-10-27
Automate docs with eslint-doc-generator

### LINKED ISSUE #903: Repo tooling: add testing to ensure each rule has a corresponding .md
Following the thread from https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/pull/833#issuecomment-1292935945: it'd be great to have testing make sure each rule is documented. In typescript-eslint we make sure rule docs exist at a certain path and follow a certain format ([typescript-eslint > `docs.test.ts`](https://github.com/typescript-eslint/typescript-eslint/blob/344322add846d03c6c9981e486b09e6ba1196555/packages/eslint-plugin/tests/docs.test.ts)).

I'm up for contributing this if you'd like it!

bmish:
I built this CLI tool [eslint-doc-generator](https://github.com/bmish/eslint-doc-generator) for automating the generation of the README rules list and rule doc title/notices for ESLint plugins. It follows common documentation conventions from this and other top ESLint plugins and will help us standardize documentation across ESLint plugins (and generally improve the usability of our custom rules through better documentation and streamline the process of adding new rules).

This builds on some of the documentation work done by @uncommon-type in #837. eslint-doc-generator is more robust compared to the previous documentation generator script which I have now removed. It has additional functionality (for example, the rules list legend is auto-generated, and the rules list will show additional columns of information when applicable) and 100% test coverage. An equivalent PR to eslint-plugin-react was also merged recently: https://github.com/jsx-eslint/eslint-plugin-react/pull/3469.

Closes #903.


src/rules/accessible-emoji.js
```diff
@@ -9,21 +9,18 @@
...
 
-const schema = generateObjSchema();
-
 export default {
   meta: {
     docs: {
       url: 'https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/tree/HEAD/docs/rules/accessible-emoji.md',
     },
     deprecated: true,
-    schema: [schema],
```
bmish, line 26: eslint-doc-generator needs to be able to detect if a rule has options for enforcing that rule docs mention options when present. Passing a no-op variable makes it impossible to statically determine if the rule has options, so I have removed these usages where no options are present.

package.json
```diff
@@ -30,8 +30,8 @@
     "posttest": "aud --production",
     "test:ci": "npm run jest -- --ci --runInBand",
     "jest": "jest --coverage __tests__/**/*",
-    "generate-list-of-rules": "md-magic --path '**/*.md' --ignore 'node_modules'",
-    "generate-list-of-rules:check": "npm run generate-list-of-rules && git diff --exit-code README.md"
+    "generate-list-of-rules": "eslint-doc-generator --rule-doc-title-format name",
```
ljharb, line 33: if this needs the build script, then:
```suggestion
    "pregenerate-list-of-rules": "npm run generate-list-of-rules",
    "generate-list-of-rules": "eslint-doc-generator --rule-doc-title-format name",
```

.github/workflows/readme.yml
```diff
@@ -12,4 +12,5 @@ jobs:
         name: 'nvm install lts/* && npm install'
         with:
           node-version: 'lts/*'
+      - run: npm run build
```
ljharb, line 15: this should no longer be needed with the package.json suggestion

docs/rules/alt-text.md
```diff
@@ -1,9 +1,15 @@
 # alt-text
 
+💼 This rule is enabled in the following configs: ✅ `recommended`, 🔒 `strict`.
```
ljharb, line 3: just like in the react plugin pr, let’s avoid a check for recommended

docs/rules/label-has-for.md
```diff
@@ -1,4 +1,10 @@
-# [Deprecated] label-has-for
+# label-has-for
```
ljharb, line 1: can the rule names include the plugin prefix?

docs/rules/prefer-tag-over-role.md
```diff
@@ -0,0 +1,5 @@
+# prefer-tag-over-role
+
+<!-- end auto-generated rule header -->
+
+TODO
```
ljharb, line 5: woof, we missed this documentation entirely?! I’ll ping on the PR that added it.

src/rules/accessible-emoji.js
ljharb: This, however, is a breaking change for existing rules, no? With the empty object, anything can be passed; otherwise, nothing can be.

docs/rules/prefer-tag-over-role.md
bmish: Yep, it was missing entirely.

package.json
bmish: I think you meant:

```
"pregenerate-list-of-rules": "npm run build",
```

Updated.

.github/workflows/readme.yml
bmish: Removed.

docs/rules/label-has-for.md
bmish: Updated by using `--rule-doc-title-format prefix-name`.

src/rules/accessible-emoji.js
bmish: Correct, the previous schema incorrectly allowed anything (garbage/unused options) to be passed:

```js
schema: [
  {
    type: 'object',
    properties: {},
    required: undefined,
  },
],
```

Switching to `schema: []` correctly prevents non-existent options from being passed.

If you would like to consider this a breaking change, then I'll remove it from the PR. I'll need to implement a feature https://github.com/bmish/eslint-doc-generator/issues/86 to disable the check for options in rule docs as it won't be possible to do this until these incorrect schemas are removed, and will update the PR once the feature is implemented.

docs/rules/prefer-tag-over-role.md
ljharb: This will be fixed in #902

package.json
ljharb: right yes sorry :-)

src/rules/accessible-emoji.js
ljharb: Yes, i think it is a breaking change, even though it’d arguably be a bug fix.

Could you replace it here with an empty object literal? ie, inline the helper

bmish: I could switch to `schema: [{}]` but that won't really help with my issue. While statically-analyzable, such a schema suggests that the rule might have options even though the schema isn't fully-defined, so EDG would still check for an options section in the rule doc. I think it's best to just disable the options section check and not touch these schemas at all.

ljharb: Sounds good.

Perhaps it would be useful to add an option to the docs generator to check schemas but assume empty objects have no config?

docs/rules/alt-text.md
ljharb: Sure, let's stick with the black and white check for now. Thanks!

src/rules/accessible-emoji.js
bmish: I removed the schema changes and added `--rule-doc-section-options false` to avoid this check for now.

> Perhaps it would be useful to add an option to the docs generator to check schemas but assume empty objects have no config?

At this time, I think such an option would be too low-level, so I'd rather users just disable the check until they can switch to `schema: []` which is more correct.

src/rules/no-interactive-element-to-noninteractive-role.js
```diff
@@ -34,7 +34,6 @@ export default ({
     docs: {
       url: 'https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/tree/HEAD/docs/rules/no-interactive-element-to-noninteractive-role.md',
       description: 'Interactive elements should not be assigned non-interactive roles.',
-      errorOptions: true,
```
ljharb, line 37: what does this metadata do, such that it's not needed?

README.md
```diff
@@ -110,95 +111,59 @@ configuration file by mapping each custom component name to a DOM element type.
...
+| :----------------------------------------------------------------------------------------------------------- | :--------------------------------------------------------------------------------------------------------------------------------- | :---- | :-- |
+| [accessible-emoji](docs/rules/accessible-emoji.md)                                                           |                                                                                                                                    |       | ❌  |
+| [alt-text](docs/rules/alt-text.md)                                                                           | Enforce all elements that require alternative text have meaningful information to relay back to end user.                          | ☑️ 🔒 |     |
+| [anchor-ambiguous-text](docs/rules/anchor-ambiguous-text.md)                                                 | Enforce `<a>` text to not exactly match "click here", "here", "link", or "a link".                                                 |       |     |
```
ljharb, line 125: is there an easy way to make these links be footnoted, so this could be:
```suggestion
| [anchor-ambiguous-text]                                                 | Enforce `<a>` text to not exactly match "click here", "here", "link", or "a link".                                                 |       |     |
```

and the raw table would be smaller?

src/rules/no-interactive-element-to-noninteractive-role.js
bmish: This is a non-standard property that was added only for the documentation generation in the previous PR but is no longer used: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/pull/837

README.md
bmish: That would make the table markdown more aesthetic, but I'm not sure that matters much since it's generated and managed by the tool.

Now if we expected or wanted to support the rule names being linked to from elsewhere in the readme, I could see an argument for switching to the reusable footnote link style, or perhaps having an option to switch to footnote links, but it would be good to spell our a use case for that first.

bmish: A downside of switching to the footnote links is doubling the managed markdown height (from the table + the list of rule links).

src/rules/no-interactive-element-to-noninteractive-role.js
ljharb: aha thanks

README.md
ljharb: I feel like that's less of an issue, personally, since it's easier for humans to read the narrower table.

Certainly not a blocker, but I do think it would be an advantage to use footnotes.

ljharb (approved):

README.md
bmish: I'm not sure I'm going to pursue this idea at the moment, but I'll at least track it here: https://github.com/bmish/eslint-doc-generator/issues/189

PR #902, opened 2022-10-27
[Docs] `prefer-tag-over-role`: rename docs file

JoshuaKGoldberg:
Fixes for https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/pull/833#issuecomment-1292935945, that I'd neglected to update the `.md` file name when renaming the rule before merge.


ljharb (approved):
thanks, glad this one was an easy fix :-D

bmish (approved):

Issue #910, opened 2022-12-02
`role-suports-aria-props` is not flagging prohibited accessible name usage on generic elements


khiga8:
I would expect [role-suports-aria-props](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/role-supports-aria-props.md)  to flag [prohibited accessible name usage](https://w3c.github.io/aria/#namefromprohibited) on generic elements like:

```
<div aria-label="Some label">
<span aria-label="Some label">
<div aria-labelledby="some-id">
<span aria-labelledby="some-id">
```

etc.

We see this misuse of accessible name a lot in our projects and we're hoping this is a case we can cover with `eslint-plugin-jsx-a11y`.

I'd hoped that the latest bump of aria-query in https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/pull/814 would start flagging these issues, but when I update the tests in `__tests__/src/rules/role-supports-aria-props-test.js` on the latest `main` to include invalid cases like:

```
    {
      code: '<div aria-label />',
      errors: [errorMessage('aria-label', 'generic', 'div', true)],
    },
```

it doesn't seem like it's being recognized as invalid.

jessebeach:
@khiga8 hello Kate! Directionally, you are on to something here. We don't have a rule that prohibits labels on elements or in combination with roles that do not support a label. I wouldn't do that in this rule, because it is specifically concerned with the combination of ARIA Role and ARIA Prop. In your examples, none of the HTML elements have a role. It's true that the `<div>` element corresponds to the `generic` role, so a label should be disallowed in this case. But what I think we need here is a new rule altogether.

jessebeach:
@khiga8 would you like to update this issue to reflect that you're asking for a new rule?

khiga8:
Hi @jessebeach! I'd noticed in the `role-supports-aria-props-test`, there are [checks for implicit roles](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/__tests__/src/rules/role-supports-aria-props-test.js#L88) and the [doc](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/role-supports-aria-props.md#jsx-a11yrole-supports-aria-props) describes the rule as the following:

> Enforce that elements with explicit or implicit roles defined contain only aria-* properties supported by that role.

[...]

PR #911, opened 2022-12-06
Flag `aria-label` and `aria-labelledby` on non-interactive elements

smockle:
### What does this PR accomplish?

Fixes https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/910

This PR indicates an error in markup like `<span aria-label="Hello, world!">Hello</span>`.

### Why?

`span` and `div` elements implicitly have the `generic` role, and `generic` prohibits `aria-label` and `aria-labelledby`.

From the [`generic` role spec](https://w3c.github.io/aria/#generic):
> The `generic` role is intended for use as the implicit role of generic elements in host languages (such as HTML `div` or `span`)
>
> […]
>
> Prohibited States and Properties: `aria-label`, `aria-labelledby`

### Implementation details

Previously, neither `div` nor `span` had an implicitly role defined in `src/util/implicitRoles/`, so the codepath in [src/rules/role-supports-aria-props.js#L55-L63](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/src/rules/role-supports-aria-props.js#L55-L63) was taken, which _“assume[s] that the element can handle the global set of aria-* properties.”_ This is not a correct assumption, so I added `src/util/implicitRoles/div.js` and `src/util/implicitRoles/span.js` and re-exported them from the `src/util/implicitRoles/index.js` barrel.

Once the [src/rules/role-supports-aria-props.js#L65-L70](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/src/rules/role-supports-aria-props.js#L65-L70) codepath was taken, I discovered that the list of a role’s [`prohibitedProps`](https://github.com/A11yance/aria-query/blob/main/src/etc/roles/literal/genericRole.js#L12-L15) was unused, so I concatenated it to the existing list of [`invalidAriaPropsForRole`](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/src/rules/role-supports-aria-props.js#L69-L70).

Next, I discovered that the list of [`validTests`](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/__tests__/src/rules/role-supports-aria-props-test.js#L74), tests expected to pass without errors, included e.g. `<div role="superscript" aria-label />`. This initially confused me, since `aria-label` is one of [`superscript`’s `prohibitedProps`](https://github.com/A11yance/aria-query/blob/main/src/etc/roles/literal/superscriptRole.js#L12-L15). I traced this to the superclass—[`roletype` includes `aria-label`](https://github.com/A11yance/aria-query/blob/main/src/etc/roles/abstract/roletypeRole.js#L23-L24)—then added a filter to remove them.

Finally, the addition of `src/util/implicitRoles/div.js` broke a few role-related tests that assumed `div` would not have an explicit or implicit role. I replaced `div` with a made-up custom element named `custom-element`, which functions the way `div` did before this PR (i.e. lacking explicit and implicit roles).

### Testing details

I ran `npm test` locally, and all tests pass:

```sh
Test Suites: 65 passed, 65 total
Tests:       16078 passed, 16078 total
Snapshots:   0 total
Time:        40.463s, estimated 43s
```

### DIFF
--- a/__tests__/src/rules/role-supports-aria-props-test.js
+++ b/__tests__/src/rules/role-supports-aria-props-test.js
@@ -579,5 +579,13 @@ ruleTester.run('role-supports-aria-props', rule, {
       code: '<span aria-labelledby />',
       errors: [errorMessage('aria-labelledby', 'generic', 'span', true)],
     },
+    {
+      code: '<div aria-label />',
+      errors: [errorMessage('aria-label', 'generic', 'div', true)],
+    },
+    {
+      code: '<div aria-labelledby />',
+      errors: [errorMessage('aria-labelledby', 'generic', 'div', true)],
+    },
   ].concat(invalidTests).map(parserOptionsMapper),
 });
--- a/__tests__/src/rules/role-supports-aria-props-test.js
+++ b/__tests__/src/rules/role-supports-aria-props-test.js
@@ -50,11 +50,14 @@ const nonAbstractRoles = [...roles.keys()].filter((role) => roles.get(role).abst
 const createTests = (rolesNames) => rolesNames.reduce((tests, role) => {
   const {
     props: propKeyValues,
+    prohibitedProps,
   } = roles.get(role);
-  const validPropsForRole = Object.keys(propKeyValues);
+  const validPropsForRole = Object.keys(propKeyValues)
+    .filter((attribute) => prohibitedProps.indexOf(attribute) === -1);
   const invalidPropsForRole = [...aria.keys()]
     .map((attribute) => attribute.toLowerCase())
-    .filter((attribute) => validPropsForRole.indexOf(attribute) === -1);
+    .filter((attribute) => validPropsForRole.indexOf(attribute) === -1)
+    .concat(prohibitedProps);
   const normalRole = role.toLowerCase();
 
   const allTests = [];
@@ -567,6 +570,14 @@ ruleTester.run('role-supports-aria-props', rule, {
       code: '<Link href="#" aria-checked />',
       errors: [errorMessage('aria-checked', 'link', 'a', true)],
       settings: componentsSettings,
+    },
+    {
+      code: '<span aria-label />',
+      errors: [errorMessage('aria-label', 'generic', 'span', true)],
+    },
+    {
+      code: '<span aria-labelledby />',
+      errors: [errorMessage('aria-labelledby', 'generic', 'span', true)],
     },
   ].concat(invalidTests).map(parserOptionsMapper),
 });
--- a/__tests__/src/util/getComputedRole-test.js
+++ b/__tests__/src/util/getComputedRole-test.js
@@ -24,7 +24,7 @@ describe('getComputedRole', () => {
       describe('lacks implicit', () => {
         it('should return null', () => {
           expect(getComputedRole(
-            'div',
+            'custom-element',
             [JSXAttributeMock('role', 'beeswax')],
           )).toBeNull();
         });
@@ -43,7 +43,7 @@ describe('getComputedRole', () => {
       describe('lacks implicit', () => {
         it('should return null', () => {
           expect(getComputedRole(
-            'div',
+            'custom-element',
             [],
           )).toBeNull();
         });
@@ -62,7 +62,7 @@ describe('getComputedRole', () => {
     describe('lacks implicit', () => {
       it('should return null', () => {
         expect(getComputedRole(
-          'div',
+          'custom-element',
           [],
         )).toBeNull();
       });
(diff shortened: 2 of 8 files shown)


smockle:
I’m stumped about why the pre-Node.js 12 checks are failing ([example failing run](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/actions/runs/3633283276/jobs/6130121504)). I didn’t change the `language-tags` package (or any build/transpilation config), which appears to be the source of the error:

```sh
689 /home/runner/work/eslint-plugin-jsx-a11y/eslint-plugin-jsx-a11y/node_modules/language-tags/lib/Tag.js:17
690 static ERR_DEPRECATED = 1;
691                       ^
692
693 SyntaxError: Unexpected token =
694
695    9 |
696   10 | import { propName, getLiteralPropValue } from 'jsx-ast-utils';
697 > 11 | import tags from 'language-tags';
698      | ^
699   12 | import { generateObjSchema } from '../util/schemas';
700   13 | import getElementType from '../util/getElementType';
701   14 |
702
703   at ScriptTransformer._transformAndBuildScript (node_modules/@jest/transform/build/ScriptTransformer.js:537:17)
704   at ScriptTransformer.transform (node_modules/@jest/transform/build/ScriptTransformer.js:579:25)
705   at Object.<anonymous> (node_modules/language-tags/lib/index.js:11:11)
706   at Object.<anonymous> (src/rules/lang.js:11:1)
707   at Object.<anonymous> (__tests__/src/rules/lang-test.js:12:1)
```

jessebeach:
@smockle thank you for the initiative here.

As a matter of principle, we don't want to store knowledge of WAI-ARIA in this plugin. That knowledge should be stored in `aria-query`: https://github.com/A11yance/aria-query

By doing so, we can leverage a store of all elements that map to specific roles. For example, there are 19 HTML elements that map to the `generic` role: https://github.com/A11yance/aria-query/blob/main/scripts/roles.json#L2961

The big complication here is that there is no constant indicator that determines if a role supports a label. Some interactive and some non-interactive roles will support a label and some roles have it outright prohibited. We get at this inconsistently right now with the `nameFrom` field on the Role definition in `aria-query`. But it hasn't been systematically filled out, so the data is incomplete. Here is the spec: https://github.com/A11yance/aria-query/blob/2d04e2928ed8f1f99950fa5f0075f1f67bfed9c7/scripts/roles.json#LL6C10-L6C10

Once we plumbed that data through `aria-query`, then we can leverage it in this plugin by filtering roles that prohibit a label on the role.

smockle:
> …some non-interactive roles will support a label…

This is a good callout; I’d forgotten about cases like `<div role="dialog" aria-labelledby="dialogHeading">`. 👍

jessebeach:
@smockle 

> Does this mean that the [src/util/implicitRoles/](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/tree/main/src/util/implicitRoles) directory will be removed in a future update? Or were you saying that [src/util/implicitRoles/div.js](https://github.com/smockle/eslint-plugin-jsx-a11y/blob/aria-label-on-non-interactive-elements/src/util/implicitRoles/div.js) and [src/util/implicitRoles/span.js](https://github.com/smockle/eslint-plugin-jsx-a11y/blob/aria-label-on-non-interactive-elements/src/util/implicitRoles/span.js) are categorically different from the other files there?

This code is a vestige of the project before `aria-query` and `axobject-query` were introduced in early 2018. I just haven't had time to remove it :) I've been going through `aria-query` and `axobject-query` to get them both in alignment with ARIA 1.3, which will be going into recommendation soon-ish and then become evergreen.

> One of the 19 generic HTML elements in aria-query is a ([scripts/roles.json#L2986-L2991](https://github.com/A11yance/aria-query/blob/2d04e2928ed8f1f99950fa5f0075f1f67bfed9c7/scripts/roles.json#L2986-L2991)). The a element only has the generic role when it lacks an href attribute, right? How does aria-query capture this—is that what [concept.attributes](https://github.com/A11yance/aria-query/blob/2d04e2928ed8f1f99950fa5f0075f1f67bfed9c7/scripts/roles.json#L3666-L3672) is for?

Exactly, the attributes define the shape of the concept, when attributes are required to match a role. You can see this in the AX Tree when using the Inspector in Chrome.

Screen Shot 2022-12-07 at 10 34 46 AM
Screen Shot 2022-12-07 at 10 35 37 AM

Many HTML elements look "semantic", but if they don't have a mapping to an AX Tree Node, then they have no semantic expression exposed to clients. That's what this tech stack is meant to formalize -- the reality of the semantic mappings from HTML (`aria-query`) and the browser's 
mapping (`axobject-query`), their overlap, and the subsequent map that gets exposed as the AX Tree. It's messy, inconsistent and highly idiosyncratic due to layers of history and calcified standards in decades-old tools :)

PR #915, opened 2023-01-09
[Deps] pin `language-tags` to `v1.0.5`

### LINKED ISSUE #914: CI is failing on old Node versions
Node versions 11 and below are failing in GitHub Actions. They're also past EOL. Can they be removed?

Example: #913 -> https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/actions/runs/3875057562/jobs/6607126082.

JoshuaKGoldberg:
Technically this is just >=12, which can include versions like 13 that are no longer supported (https://github.com/nodejs/release#release-schedule). But I figured I'd keep using the existing action to get Node versions.

Fixes #914.

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "emoji-regex": "^9.2.2",
     "has": "^1.0.3",
     "jsx-ast-utils": "^3.3.3",
-    "language-tags": "^1.0.5",
+    "language-tags": "=1.0.5",
     "minimatch": "^3.1.2",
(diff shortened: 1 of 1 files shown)


ljharb (changes requested):
No thanks; this is a breaking change, and there's no reason to do it. official support of node is irrelevant.

ljharb (approved):
I've updated the PR to pin the misbehaving dependency.

PR #916, opened 2023-01-12
[Fix] `no-aria-hidden-on-focusable` rule's missing export

EvHaus:
I noticed v6.7.0 shipped a new `no-aria-hidden-on-focusable` rule, but the `index.js` file doesn't export it so when you try to use it you get:

```
Error:   1:1  error  Definition for rule 'jsx-a11y/no-aria-hidden-on-focusable' was not found  jsx-a11y/no-aria-hidden-on-focusable
```

This PR fixes that problem.


ljharb (approved):
good catch, thanks!

PR #919, opened 2023-01-25
[Fix] `control-has-associated-label`: don't accept whitespace as an accessible label (Fixes #918)

### LINKED ISSUE #918: `control-has-associated-label` accepts whitespace as a label
## The problem
The `control-has-associated-label` rule currently accepts whitespace as an accessible label. This issue is exaggerated by the use of `prettier` which may introduce line breaks and other whitespace automatically, defeating the rule.

### Examples
All of the below examples lack accessible labels.
```html
<button></button>                <!-- rule correctly flags invalid markup -->
<button> </button>               <!-- rule fails to flag invalid markup -->

<button aria-label=""></button>  <!-- rule correctly flags invalid markup -->
<button aria-label=" "></button> <!-- rule fails to flag invalid markup -->
```

## Suggested fix
Update `mayHaveAccessibleLabel` to `.trim()` the node's value before asserting truthiness.

wobo-nate:
Fixes #918

### DIFF
--- a/__tests__/src/util/mayHaveAccessibleLabel-test.js
+++ b/__tests__/src/util/mayHaveAccessibleLabel-test.js
@@ -41,6 +41,14 @@ describe('mayHaveAccessibleLabel', () => {
         JSXAttributeMock('aria-label', ''),
       ], []))).toBe(false);
     });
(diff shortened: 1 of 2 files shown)


ljharb (approved):
nice work

PR #921, opened 2023-02-03
Enable no-aria-hidden-on-focusable rule in config

khiga8:
I was working on a `eslint-plugin-jsx-a11y` bump and realized this new rule, `no-aria-hidden-on-focusable`, isn't pulled in by the recommended or strict config. 

Not sure which category it should be enabled under, so I enabled it for both.


ljharb:
That's intentional, because doing so is a breaking change.

khiga8:
@ljharb Ohh gotcha. Assuming this is supposed to happen in the next major release. I'll close this.

PR #928, opened 2023-03-29
[New] `mouse-events-have-key-events`: add `hoverInHandlers`/`hoverOutHandlers` config

### LINKED ISSUE #927: Should `mouse-events-have-key-events` check for onMouseEnter/Leave, onPointerOver/Enter/Leave/Out?
As the title says, I'm wondering if the rule `mouse-events-have-key-events.js` should, in addition to only `onMouseOver` and `onMouseOut`, also check for the `onMouseEnter/Leave` and `onPointerOver/Enter/Leave/Out` events?

Some context for this question is that I want to enable `mouse-events-have-key-events` in our codebase, but it turns out we have a separate custom rule that tells you to use `onMouseEnter` instead of `onMouseOver` and `onMouseLeave` instead of `onMouseOut`, so `mouse-events-have-key-events` would effectively not do anything for us I guess...

What do we think? I'm happy to submit a PR, but wanted to ask first in case there are reasons for not supporting those additional events.

Thanks!

willwull:
Fixes #927

As discussed in #927, this PR adds an optional argument to the `mouse-events-have-key-events` rule which allows the user to set any arbitrary events that must be paired with `onFocus` and `onBlur` respectively, instead of only checking `onMouseOver` and `onMouseOut`.

Also added a flow annotation to the rule while I was there, mostly to make it easier for myself while I was writing the code.

This was my first time contributing to this project and working on an ESLint rule in general, so any feedback on best practices is appreciated.

Open questions:
1. I used the names "hoverInHandlers" and "hoverOutHandlers" for the config. Are these understandable? (I'm pretty bad at naming things)
2. IF a config is passed, `onMouseOver` and `onMouseOut` (the default handlers) are NOT included implicitly, the user must remember to include these IF a config is passed in. Is this ok?

### DIFF
--- a/docs/rules/mouse-events-have-key-events.md
+++ b/docs/rules/mouse-events-have-key-events.md
@@ -6,9 +6,35 @@
 
 Enforce onmouseover/onmouseout are accompanied by onfocus/onblur. Coding for the keyboard is important for users with physical disabilities who cannot use a mouse, AT compatibility, and screenreader users.
 
-## Rule details
+## Rule options
 
-This rule takes no arguments.
+By default, this rule checks that `onmouseover` is paired with `onfocus` and that `onmouseout` is paired with `onblur`. This rule takes an optional argument to specify other handlers to check for "hover in" and/or "hover out" events:
+
+```json
+{
+  "rules": {
+    "jsx-a11y/mouse-events-have-key-events": [
+      "error",
+      {
+        "hoverInHandlers": [
+          "onMouseOver",
+          "onMouseEnter",
+          "onPointerOver",
+          "onPointerEnter"
+        ],
+        "hoverOutHandlers": [
+          "onMouseOut",
+          "onMouseLeave",
+          "onPointerOut",
+          "onPointerLeave"
+        ]
+      }
+    ]
+  }
+}
+```
+
+Note that while `onmouseover` and `onmouseout` are checked by default if no arguments are passed in, those are *not* included by default if you *do* provide an argument, so remember to explicitly include them if you want to check them.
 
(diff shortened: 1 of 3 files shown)


willwull:
Seems like the `flow` test is failing because some download is failing? not sure what I can do to fix that

docs/rules/mouse-events-have-key-events.md
```diff
@@ -6,9 +6,35 @@
...
+
+By default, this rule checks that `onmouseover` is paired with `onfocus` and that `onmouseout` is paired with `onblur`. This rule takes an optional argument to specify other handlers to check for "hover in" and/or "hover out" events:
+
+```json
+{
+  "rules": {
+    "jsx-a11y/mouse-events-have-key-events": [
+      "error",
+      {
+        "hoverInHandlers": [
```
ljharb, line 19: Should this also support a boolean true, for those who want to get this entire list without hardcoding it?

jessebeach: We've used enumerated lists in all the other rules. Let's keep the list approach here, rather than introducing a variable type to the rule's options.

jessebeach:
This update looks great!! No notes. Thank you @willwull for the test coverage.

jessebeach (approved):

jessebeach:
@ljharb it looks like the one failing check failed for an infra reason and not something to do with this change. I don't have the permission to re-run it. Do you? 

> Tests: pretest/posttest / flow (pull_request)

ljharb:
Yup, re-ran.

PR #930, opened 2023-04-18
[Deps] Upgrade ast-types-flow to mitigate Docker user namespacing problems

rmehner:
Version 0.0.7 had a problem with a very high user id, which lead to problems when used in Docker with user namespacing. For some details check: https://github.com/kyldvs/ast-types-flow/issues/5

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "@babel/eslint-parser": "^7.19.1",
     "@babel/plugin-transform-flow-strip-types": "^7.19.0",
     "@babel/register": "^7.18.9",
-    "ast-types-flow": "^0.0.7",
+    "ast-types-flow": "^0.0.8",
     "aud": "^2.0.2",
     "auto-changelog": "^2.4.0",
     "babel-jest": "^24.9.0",


ljharb (approved):
Thanks! This is a dev dep tho, so i'm not sure why it would be an issue for anyone who's not actively maintaining this package :-)

rmehner:
> Thanks! This is a dev dep tho, so i'm not sure why it would be an issue for anyone who's not actively maintaining this package :-)

Thanks for the quick merge. Unfortunately in one of my client projects the dev dependencies end up in the container's intermediate layers to allow the build to happen within the Dockerfile, so this fixes it :)

ljharb:
That's surprising since there's no build needed - the build output is part of the npm package already.

rmehner:
> That's surprising since there's no build needed - the build output is part of the npm package already.

Sorry for being unclear. In this case it's another project that depends on `eslint-plugin-jsx-a11y` which depends on `ast-types-flow` and this project is trying to install dev dependency within docker, which leads to the aforementioned problem.

ljharb:
I’m still unclear on how that’s possible unless they’re incorrectly running `npm install --dev` - npm doesn’t install transitive dev deps by default.

rmehner:
Dockerfile:

```Dockerfile
FROM node:18.16.0-slim

COPY package*.json ./

RUN npm ci
```

and the package.json:

```json
{
  "devDependencies": {
    "eslint-plugin-jsx-a11y": "^6.7.1"
  }
}
```

And since I didn't have access to the Dockerfile for reasons, this has been the best option.

ljharb:
I'd suggest making a new lockfile from scratch; dev deps of your dev deps (or deps) should never be installed.

Issue #938, opened 2023-06-21
`role-supports-aria-props` documentation page displays a non working sample of code


n-chardon:
The following code is displayed as the "Good" example of `role-supports-aria-props` rule : 

```
<ul role="radiogroup" aria-required aria-labelledby="foo">
    <li tabIndex="-1" role="radio" aria-checked="false">Rainbow Trout</li>
    <li tabIndex="-1" role="radio" aria-checked="false">Brook Trout</li>
    <li tabIndex="0" role="radio" aria-checked="true">Lake Trout</li>
</ul>
```

However, this code does not work as intended, because the `aria-required` attribute has a "false" default value: it should be written as `aria-required="true"` to work properly with assistive technology.

I would recommend that the documentation be updated with a working example.

ljharb:
In HTML, does `aria-required="aria-required"` not work properly?

PR #939, opened 2023-06-21
Update role-supports-aria-props.md: fix code sample

n-chardon:
fixes https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/938

### DIFF
--- a/docs/rules/role-supports-aria-props.md
+++ b/docs/rules/role-supports-aria-props.md
@@ -13,7 +13,7 @@ This rule takes no arguments.
 ### Succeed
 ```jsx
 <!-- Good: the radiogroup role does support the aria-required property -->
-<ul role="radiogroup" aria-required aria-labelledby="foo">
+<ul role="radiogroup" aria-required="true" aria-labelledby="foo">
     <li tabIndex="-1" role="radio" aria-checked="false">Rainbow Trout</li>
     <li tabIndex="-1" role="radio" aria-checked="false">Brook Trout</li>
     <li tabIndex="0" role="radio" aria-checked="true">Lake Trout</li>


docs/rules/role-supports-aria-props.md
```diff
@@ -13,7 +13,7 @@ This rule takes no arguments.
 ### Succeed
 ```jsx
 <!-- Good: the radiogroup role does support the aria-required property -->
-<ul role="radiogroup" aria-required aria-labelledby="foo">
+<ul role="radiogroup" aria-required="true" aria-labelledby="foo">
```
ljharb, line 16: what needs fixing here exactly?

n-chardon: the aria-required attribute must have a "true" value to work properly with assistive technology. Without the value, the attribute does not work to inform screenreaders of the "required" state of the group. 

Without `="true"` in devtools : 
image

With `aria-required="true"` : 
image

ljharb: Thanks. It seems like we don't have any rule or option that can catch this currently - @jessebeach, does this seem like a good addition? Perhaps an option to `aria-props` or something?

sangikhan29:
@ljharb, do you know why the tests here are failing for this simple update to a `md` file? I was working on a [draft](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/pull/940) and found similar tests failing. I ran `npm run jest` on the master branch itself locally and found the same thing.

ljharb:
Yes, see #937; you can ignore it for now.

docs/rules/role-supports-aria-props.md
jessebeach: @ljharb and @n-chardon, honestly, it strikes me as something that requires its own rule, something like `no-empty-boolean-attributes`. The attributes that such a rule would apply to are:

```
aria-atomic
aria-busy
aria-disabled
aria-modal
aria-multiline
aria-multi selectable
aria-readonly
aria-required

```

And, confusingly, there is a tri-state version ([true/false/undefined](https://www.w3.org/TR/wai-aria-1.2/#propcharacteristic_value)) of these "boolean" attributes that allow "undefined", which indicates that the attribute is not applicable. Such a state shouldn't exist, but it exists in the spec, so it needs to be accounted for here:

```
aria-expanded
aria-grabbed
aria-hidden
aria-selected
```

A good solution is pulling the properties map from https://github.com/A11yance/aria-query/blob/main/src/ariaPropsMap.js and refactoring it into a util. A better solution is referring to `aria-query` to get the information about the allowed props and their value types rather than having this information hard-coded in this lint rule plugin.

@n-chardon would you like to write up this rule?

Issue #932, opened 2023-05-16
[no-noninteractive-element-interactions] should accept the dialog role


sangikhan29:
Dialog has to handle events such as moving or resizing  so dialog should be able to accept event handler.
> If the dialog can be moved or resized, ensure that these actions must be performable by keyboard users as well as mouse users.
[reference](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role)

Also refer to a similar previous [issue](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/741)

jessebeach:
Good points, let me investigate a bit.

jessebeach:
I'm consulting the ARIA oracle ☺️

jessebeach:
So, @mcking65 and I had a chat. Here are his thoughts:

```
An "ARIA window" is interactive.
 
However, there are some nuances related to dialog.
 
We normally expect interactive elements to be focusable. If not directly focusable, referenceable via activedescendant.
 
A significant contingent of people (including me) assert that from a screen reader perspective, making dialogs focusable is problematic. However, we conceded to allow browsers to focus dialogs as a fallback.
 
Browser devs need a place to put focus when an HTML dialog opens. There are several ways that authors can set focus and some options for when they don’t. The issue that held up HTML dialog was what browsers should do when authors don’t set focus and the dialog does not contain a focusable element. In this situation, it would be bad if focus were “lost” (e.g., set to the document. Thus, dialogs are “allowed” to be focus. But, this should only happen when necessary as a recovery mechanism.
 
This is the primary difference between a window and other interactive elements.
```

So, only roles descended from `roletype > widget` are considered interactive: 

https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/src/util/isInteractiveRole.js#L11

The change that we will make is, including roles descended from `roletype > window`, which will include `dialog` and `alertdialog`. 

To be determined how we deal with the focus issue in this plugin, if at all.

How does that all sound?

ckundo:
@jessebeach sounds great! thank you for the background and careful consideration.

sangikhan29:
Thank you @jessebeach for looking into this! It sounds like a reasonable approach for me as well.

jessebeach:
> Thank you @jessebeach for looking into this! It sounds like a reasonable approach for me as well.

[...]

PR #940, opened 2023-06-21
add window to isInteractive

sangikhan29:
Addresses https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/932


__mocks__/genInteractives.js
```diff
@@ -71,7 +72,6 @@ const nonInteractiveElementsMap: {[string]: Array<{[string]: string}>} = {
   dd: [],
   details: [],
   dfn: [],
-  dialog: [],
```
sangikhan29, line 74: Updating this list since it's used in the unit tests to confirm the util returns the right value.

ckundo:
blocked by #941 !

jessebeach:
> blocked by #941 !

I saw this as well over the weekend. Trying to get this train back on the rails.

ljharb:
Rebased, and moved "draft" from the title to the PR status. @sangikhan29, please mark as ready for review whenever it is.

__tests__/src/rules/control-has-associated-label-test.js
```diff
@@ -49,11 +49,14 @@ const alwaysValid = [
...
   { code: '<option>Save</option>' },
   { code: '<th>Save</th>' },
+  { code: '<dialog>Save</dialog>' },
   // Interactive Roles
+  { code: '<div role="alertdialog">Save</div>' },
   { code: '<div role="button">Save</div>' },
   { code: '<div role="checkbox">Save</div>' },
   { code: '<div role="columnheader">Save</div>' },
   { code: '<div role="combobox">Save</div>' },
+  { code: '<div role="dialog">Save</div>' },
```
sangikhan29, line 59: I think it's fair to expect dialog to have a label according to [this](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role#labeling).

__mocks__/genInteractives.js
```diff
@@ -22,6 +22,7 @@ const interactiveElementsMap = {
   button: [],
   canvas: [],
   datalist: [],
+  dialog: [],
```
sangikhan29, line 25: bumping `dialog` into interactive element

src/util/isInteractiveRole.js
```diff
@@ -8,7 +8,7 @@ import flatMap from 'array.prototype.flatmap';
 const roles = [...rolesMap.keys()];
 const interactiveRoles = roles.filter((name) => (
   !rolesMap.get(name).abstract
-  && rolesMap.get(name).superClass.some((klasses) => includes(klasses, 'widget'))
+  && rolesMap.get(name).superClass.some((klasses) => includes(klasses, 'widget') || includes(klasses, 'window'))
```
sangikhan29, line 11: I find this change to be the easiest way to implement [this suggestion](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/932#issuecomment-1594877723). But I realized that this change will also impact  [interactive-supports-focus](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/src/rules/interactive-supports-focus.js), where the linter will throw an error if a `dialog` isn't focusable. But according to [mdn](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role), `dialog` is not required to be focusable although it should have a child that's focusable.

[My case for requesting](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/932) `dialog` to be interactive was for more so that it should be allowed (rather than required) to be interactive. I'd feel more comfortable just treating `dialog` [somewhat similarly to `presentation`](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/ce4d57f853ce7f71bd31edaa524eeb3ff1d27cf1/src/rules/interactive-supports-focus.js#L81) on case-by-base basis rather than have it follow all the expectations attached to an interactive role.

@jessebeach would do you have any concerns with just adding custom checks for `dialog`(+`alertdialog`) in `no-static-element-interactions` and `no-noninteractive-element-interactions` rather than setting `window` to be interactive? cc: @ckundo

jessebeach:
I'm seeing the tasks and commentary come through email. Work-work is a bit bananas lately. As soon as we get through the end-of-the-year push and slide into the holidays, I'll have time to focus on this repo.

ljharb:
Sounds good :-) no pressure.

ckundo:
hi friends, bumping this in case there's still an appetite for it!

Issue #656, opened 2019-12-19
summary should be interactive


sh0ji:
I believe the `summary` element should be considered an interactive element and that there shouldn't be any errors thrown when a `click` listener is attached to it:

```jsx
// currently flags the following:
// jsx-a11y/click-events-have-key-events
// jsx-a11y/no-static-element-interactions
<summary onClick={clickHandler} />
```

[The spec does list `details` element as interactive](https://html.spec.whatwg.org/multipage/dom.html#interactive-content), but this is misleading since [its activation behavior](https://html.spec.whatwg.org/multipage/interactive-elements.html#the-details-element:activation-behaviour) is triggered by its child `summary` OR a UA-provided affordance (all browsers add a shadow `summary` to accomplish this affordance). The spec for `summary` does also describe [the `summary` activation behavior](https://html.spec.whatwg.org/multipage/interactive-elements.html#the-summary-element:activation-behaviour), making it more clear that the `summary` is the element that activates the `details` open state.

This all also makes logical sense since the `details` can _contain_ interactive content, meaning it cannot itself be interactive:

```html
<details>
  <summary>More Information</summary>
  <!-- this link couldn't be interactive if the parent details were interactive -->
  <p>The <a href="#">details element</a> is fun!</p>
</details>
```

Further, both browsers and screen readers treat the `summary` like a `button` element (some screen readers actually call it an expandable button), including the same keyboard interaction model as a `button`: it is focusable without `tabindex` and <kbd>Enter</kbd> triggers `click` repeatedly when held (i.e., on `keydown`); <kbd>Space</kbd> triggers `click` on `keyup` only. I made a simple CodePen to demo/test this: https://codepen.io/sh0ji/pen/OJPppER.

Note that this also means that #582 is not a false positive.

jessebeach:
@sh0ji you're totally right, `summary` should be interactive. Let me figure out where to fix this. It's probably in aria-query or axobject-query. Great sleuthing here!

https://w3c.github.io/html-aria/#el-summary
https://w3c.github.io/html-aria/#el-details

jessebeach:
This was closed erroneously. This issue is still present.

stuartjnelson:
I know this is bad open-source etiquette. We're also experiencing this problem currently. Does anyone have any pointers about what we'd need to do in order to fix? I'd be happy putting together a PR

jessebeach:
[...]

PR #941, opened 2023-06-24
Fix failing tests in the main branch

jessebeach:
There are a few failing tests in the main branch and this fixes them.

This PR also pegs `aria-query` to version 5.1.2.


package.json
```diff
@@ -87,7 +87,7 @@
     "minimatch": "^3.1.2",
     "object.entries": "^1.1.6",
     "object.fromentries": "^2.0.6",
-    "semver": "^6.3.0"
+    "semver": "^7.5.3"
```
ljharb, line 90: pretty sure we can’t ever upgrade from v6 because we support node 4

package.json
```diff
@@ -73,7 +73,7 @@
   "license": "MIT",
   "dependencies": {
     "@babel/runtime": "^7.20.7",
-    "aria-query": "^5.1.3",
+    "aria-query": "5.1.3",
```
ljharb, line 76:
```suggestion
    "aria-query": "=5.1.2",
```

package.json
jessebeach: Hmmm, maybe I misinterpreted this posttest failure to require this update. Let me run posttest locally and see what's causing the failure. I'll try to get this PR to green.

Screenshot 2023-06-26 at 11 22 04 AM

ljharb: You didn't misinterpret it - i added `audit-level=moderate` to npmrc in this PR already to ignore the failure.

jessebeach:
Finally found some time (and focus) on a flight to get back to this.

.npmrc
```diff
@@ -1,3 +1,4 @@
 package-lock=false
 allow-same-version=true
 message=v%s
+audit-level=low
```
jessebeach, line 4: Just testing a theory.

jessebeach: @ljharb I'm going to need help with this one. It seems we're in a bind. We can't update `semver` because of this package's commitment Node 4 support. Here are some options

1. We could turn off the `posttest` checks for now, until the Node 4 is no longer supported in this project and `semver` can be updated.
2. Comment out the usages of `semver` in the project -- which means our test coverage for optional chaining is reduced -- and remove the dependency.
3. Update `semver` and drop Node 4 support.
4. Move the test cases that are wrapped in the `semver` check to their own folder and configure ESLint to only run on them if the version of ESLint is greater than v6.

I'm going to try #4, but I'm on a plane right now and internet is spotty, so please ignore some thrash in the PR as I use it to test out some theories.

ljharb: Don't worry about the semver tests - if that's the only failure, i can take care of that. There's still a bunch of failing node tests tho.

ljharb: Actually, if you rebase on top of #944 instead of main, then that should address the problem :-) that way i can merge both PRs at once once the other tests are passing.

jessebeach: Will do

__tests__/src/rules/no-static-element-interactions-test.js
```diff
@@ -340,6 +339,7 @@ const neverValid = [
   { code: '<style onClick={() => {}} />;', errors: [expectedError] },
   { code: '<sub onClick={() => {}} />;', errors: [expectedError] },
   { code: '<sup onClick={() => {}} />;', errors: [expectedError] },
+  { code: '<summary onClick={() => {}} />;', errors: [expectedError] },
```
dummdidumm, line 342: I'm not sure if this test adjustment is right, this may be an issue in `axobject-query`: https://github.com/A11yance/axobject-query/issues/319

karlnorling:
@jessebeach - update on this? I'm waiting to update due to the semver being in `dependencies` it triggers security warning https://security.snyk.io/vuln/SNYK-JS-SEMVER-3247795 for semver.

I noticed it's in https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/pull/944 also. Maybe we shouldn't combine the PRs and have #944 merged first?

ljharb:
@karlnorling semver v5, v6, and v7 all have patches that fix that issue, so nobody should need to do anything except end users updating their lockfiles.

ljharb:
Looks like axe-core broke us too: https://github.com/dequelabs/axe-core/issues/4127

ljharb:
Last 2 failing tests seem to be about error messages for two tests showing up in the wrong order.

ljharb (approved):

jessebeach:
> @jessebeach - update on this?

@karlnorling, @ljharb merged it yesterday.

__tests__/src/rules/no-static-element-interactions-test.js
jessebeach: @dummdidumm ya, I think you're right. This is a tricky one. More details here: https://html.spec.whatwg.org/multipage/interactive-elements.html#the-summary-element

Given that we need to know the rendered output to make a determination about this element's role and behavior, the best course of action for a static linter is to be permissive of either state - an interactive element or a generic (static) element. I'll get this updated when I do a proper update of aria-query for ARIA 1.3.

jessebeach: @dummdidumm I filed #656 to tracked this.

PR #942, opened 2023-07-02
Update package.json

CaptThomas:
Hey reviewers, thanks for a great tool. This PR is designed to fix CVE-2022-25883. I tested this locally using Windows 11 x64 using node v20.3.1 and npm 9.7.2 and it worked as expected.


ljharb:
It’s not a vulnerability here - like most transitive dep CVEs, it’s a false positive -  and we can’t upgrade because v7 drops support for engines we need to support.

PR #945, opened 2023-07-14
Account for polymorphic components in `getElementType`

kendallgassner:
### What

Currently, `getElementType` does not have a great way for testing the semantics of polymorphic components. This PR is intended to fix that.

By default, the `getElementType` will search for a node's attributes for the prop `as`. If `as` is set `getElementType` will return the value of `as`.

### flexibility 

Some libraries do not use `as` to set the semantics of their polymorphic components (e.g. perhaps they use `html`, or `asChild`). In order to allow for flexibility, I introduced a new setting `polymorphicPropName` that allows users to configure the name of the polymorphic prop.

```json
{
  "settings": {
    "jsx-a11y": {
      "polymorphicPropName": "asChild",
      "components": {
        "Box": "p",
        "Link": "a"
      }
    }
  }
}
```

### Questions

A couple of thoughts came up when making this pr:

1. Should I allow users to shut off the polymorphic prop search?
2. Should I allow multiple polymorphicPropNames to be passed in:
```json
   "polymorphicPropName": ["asChild", "html"],
```

### DIFF
--- a/__tests__/src/rules/accessible-emoji-test.js
+++ b/__tests__/src/rules/accessible-emoji-test.js
@@ -43,6 +43,10 @@ ruleTester.run('accessible-emoji', rule, {
       code: '<CustomInput type="hidden">🐼</CustomInput>',
       settings: { 'jsx-a11y': { components: { CustomInput: 'input' } } },
     },
+    {
+      code: '<Box as="input" type="hidden">🐼</Box>',
+      settings: { 'jsx-a11y': { polymorphicPropName: 'as' } },
+    },
   )).map(parserOptionsMapper),
   invalid: parsers.all([].concat(
     { code: '<span>🐼</span>', errors: [expectedError] },
@@ -53,5 +57,10 @@ ruleTester.run('accessible-emoji', rule, {
     { code: '<Foo>🐼</Foo>', errors: [expectedError] },
     { code: '<span aria-hidden="false">🐼</span>', errors: [expectedError] },
     { code: '<CustomInput type="hidden">🐼</CustomInput>', errors: [expectedError] },
+    {
+      code: '<Box as="span">🐼</Box>',
+      settings: { 'jsx-a11y': { polymorphicPropName: 'as' } },
+      errors: [expectedError],
+    },
   )).map(parserOptionsMapper),
 });
--- a/__tests__/src/rules/lang-test.js
+++ b/__tests__/src/rules/lang-test.js
@@ -25,6 +25,7 @@ const expectedError = {
 
 const componentsSettings = {
   'jsx-a11y': {
+    polymorphicPropName: 'as',
     components: {
       Foo: 'html',
     },
@@ -46,11 +47,13 @@ ruleTester.run('lang', rule, {
     { code: '<HTML lang="foo" />' },
     { code: '<Foo lang={undefined} />' },
     { code: '<Foo lang="en" />', settings: componentsSettings },
+    { code: '<Box as="html" lang="en"  />', settings: componentsSettings },
   )).map(parserOptionsMapper),
   invalid: parsers.all([].concat(
     { code: '<html lang="foo" />', errors: [expectedError] },
     { code: '<html lang="zz-LL" />', errors: [expectedError] },
     { code: '<html lang={undefined} />', errors: [expectedError] },
     { code: '<Foo lang={undefined} />', settings: componentsSettings, errors: [expectedError] },
+    { code: '<Box as="html" lang="foo" />', settings: componentsSettings, errors: [expectedError] },
   )).map(parserOptionsMapper),
 });
--- a/__tests__/src/rules/media-has-caption-test.js
+++ b/__tests__/src/rules/media-has-caption-test.js
@@ -33,6 +33,7 @@ const customSchema = [
 
 const componentsSettings = {
   'jsx-a11y': {
+    polymorphicPropName: 'as',
     components: {
       Audio: 'audio',
       Video: 'video',
@@ -144,6 +145,10 @@ ruleTester.run('media-has-caption', rule, {
       code: '<Audio muted={true}></Audio>',
       settings: componentsSettings,
     },
+    {
+      code: '<Box as="audio" muted={true}></Box>',
+      settings: componentsSettings,
+    },
   )).map(parserOptionsMapper),
   invalid: parsers.all([].concat(
     { code: '<audio><track /></audio>', errors: [expectedError] },
@@ -203,6 +208,11 @@ ruleTester.run('media-has-caption', rule, {
     },
     {
       code: '<Video><Track kind="subtitles" /></Video>',
+      settings: componentsSettings,
+      errors: [expectedError],
+    },
+    {
+      code: '<Box as="audio"><Track kind="subtitles" /></Box>',
       settings: componentsSettings,
       errors: [expectedError],
     },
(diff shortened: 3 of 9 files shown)


ljharb:
I'd be open to being convinced, but currently I think polymorphic components are a bad practice that makes code harder to maintain and understand, and I'm not sure it's a good idea to enable that by supporting it here.

kendallgassner:
I agree that polymorphic components can make code harder to follow but I like that they allow teams to utilize design system components without having to compromise on the semantics for their specific use case. 

If a design system has a component that is almost perfect but needs a small semantic change to improve accessibility, should we make the team re-write the component? Or make the team request/implement a design system change?  -- Requesting a change could take a long time and re-writing the component adds tech debt.

Though, it might be nice to implement the change on the design system side polymorphic components give teams an opportunity to make quick semantic change to a component that could prevent issues for many users.

ljharb:
I guess I'd expect each component in the design system to be for a single semantic use case - and if someone wants a new one, the design system could provide a new wrapping component for it (with a refactor so that both components wrapped an internal one with a prop to control the semantics).

That would allow you to lint for, and clearly convey, that each component has a single semantic use. Is that approach not workable?

kendallgassner:
My other thought...

This pattern is so widely used:
- [Radix UI](https://www.radix-ui.com/docs/primitives/utilities/polymorphic#polymorphic)
- [Styled components](https://styled-components.com/docs/api#as-polymorphic-prop)
- [Carbon Design System AspectRatio.mdx](https://github.com/carbon-design-system/carbon/blob/3be26a7353c2e228c45a863f743ca247eed10246/packages/react/src/components/AspectRatio/AspectRatio.mdx#L44) 
- [Material UI PolymorphicComponent.ts](https://github.com/mui/material-ui/blob/dfc4ffd8527439d6de5fe8356d0916f283597db4/packages/mui-base/src/utils/PolymorphicComponent.ts#L10)

that it feels essential to account for polymorphic components. I think even if the setting was defaulted off, allowing for polymorphic support would benefit a wide range of jsx-a11y users.

kendallgassner:
TODO: 
- Use `getLiteralPropValue` instead of `getPropValue`

ljharb:
Those are indeed widely-used systems - although I remain convinced that polymorphic components are a harmful pattern, being able to lint for their usage reduces harm, so let's go forward with this as an option, defaulted to false, that somehow notes in the readme that polymorphic components are a pattern to be avoided.

kendallgassner:
I tested `getElementType` tests locally. I believe the other failures are due to:
- https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/pull/941

kendallgassner:
@ljharb Do you know when the test fixes will be merged?

ljharb:
They're now merged, and I rebased this PR and cleaned up the read me a bit. Taking another look.

ljharb (commented):
This looks good; let's also add some tests in various rules that guarantee we're exercising getElementType indirectly?

__tests__/src/rules/lang-test.js
```diff
@@ -46,11 +47,14 @@ ruleTester.run('lang', rule, {
     { code: '<HTML lang="foo" />' },
     { code: '<Foo lang={undefined} />' },
     { code: '<Foo lang="en" />', settings: componentsSettings },
+    { code: '<Foo lang="en" />', settings: componentsSettings },
```
kendallgassner, line 50:
```suggestion
```

__tests__/src/rules/lang-test.js
```diff
@@ -46,11 +47,13 @@ ruleTester.run('lang', rule, {
     { code: '<HTML lang="foo" />' },
     { code: '<Foo lang={undefined} />' },
     { code: '<Foo lang="en" />', settings: componentsSettings },
-  )).map(parserOptionsMapper),
+    { code: '<Box as="html" lang="en"  />', settings: componentsSettings },
+    )).map(parserOptionsMapper),
```
kendallgassner, line 51:
```suggestion
   )).map(parserOptionsMapper),
```

__tests__/src/rules/media-has-caption-test.js
```diff
@@ -144,7 +145,11 @@ ruleTester.run('media-has-caption', rule, {
       code: '<Audio muted={true}></Audio>',
       settings: componentsSettings,
     },
-  )).map(parserOptionsMapper),
+    {
+      code: '<Box as="audio" muted={true}></Box>',
+      settings: componentsSettings,
+    },
+    )).map(parserOptionsMapper),
```
kendallgassner, line 152:
```suggestion
   )).map(parserOptionsMapper),
```

__tests__/src/rules/accessible-emoji-test.js
```diff
@@ -44,6 +44,12 @@ ruleTester.run('accessible-emoji', rule, {
       settings: { 'jsx-a11y': { components: { CustomInput: 'input' } } },
     },
   )).map(parserOptionsMapper),
+  invalid: parsers.all([].concat(
+    {
+      code: '<Box as="input" type="hidden">🐼</Box>',
+      settings: { 'jsx-a11y': { polymorphicPropName: 'as' } },
+    },
+    )).map(parserOptionsMapper),
   invalid: parsers.all([].concat(
```
ljharb, line 53:
```suggestion
```

__tests__/src/util/getElementType-test.js
```diff
@@ -17,6 +18,9 @@ describe('getElementType', () => {
     it('should return the exact tag name for names that are in Object.prototype', () => {
       expect(elementType(JSXElementMock('toString').openingElement)).toBe('toString');
     });
+    it('should return the default tag name provided', () => {
```
ljharb, line 21:
```suggestion

    it('should return the default tag name provided', () => {
```

ljharb (commented):
LGTM overall!

__tests__/src/rules/media-has-caption-test.js
```diff
@@ -205,6 +210,11 @@ ruleTester.run('media-has-caption', rule, {
       code: '<Video><Track kind="subtitles" /></Video>',
       settings: componentsSettings,
       errors: [expectedError],
+    },
```
kendallgassner, line 213:
```suggestion
    },
```

__tests__/src/rules/accessible-emoji-test.js
```diff
@@ -45,6 +45,10 @@ ruleTester.run('accessible-emoji', rule, {
     },
   )).map(parserOptionsMapper),
   invalid: parsers.all([].concat(
+    {
+      code: '<Box as="input" type="hidden">🐼</Box>',
+      settings: { 'jsx-a11y': { polymorphicPropName: 'as' } },
+    },
```
ljharb, line 51: maybe this was meant to be a valid test?
```suggestion
    {
      code: '<Box as="input" type="hidden">🐼</Box>',
      settings: { 'jsx-a11y': { polymorphicPropName: 'as' } },
    },
  )).map(parserOptionsMapper),
  invalid: parsers.all([].concat(
```

kendallgassner: Yikes, Sorry I ran these locally but I think the merging messed me up 😞

ljharb: np, it happens :-) this might be ready to land after things pass!

kendallgassner:
@ljharb sorry to re-ping you! I just wanted to check in and see if we can move forward on this PR.

ljharb (approved):
thanks for the ping, sorry for the delay

PR #947, opened 2023-07-28
[Docs] `no-static-element-interactions`: `tabIndex` is written `tabindex` 

andreasnuesslein:
see
https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

linting will also fail if it's `tabIndex` with a capital i


ljharb (approved):
Thanks!

ljharb:
The tests are failing on main right now; once fixed I'll rebase this and land it.

PR #963, opened 2023-11-10
remove object.assign

dulmandakh:
I found that Object.assign is already supported in Node.js 4.0.0, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign. So removed object.assign package and opted to Object.assign. Also I had to turn eslint rule because spread operator it not supported on Node.js 4.

I would like to propose to raise Node.js version requirement, maybe up to 16.13.0, then remove polyfills to reduce install size. I would like to help with this.

Here are polyfills and Node.js versions what supports these features. 

| Dependency |Node.js version|
|-------------------------|------|
| array-includes | 6.0.0 |
| object.entries | 7.0.0 |
| array.prototype.flatmap | 11.0.0 |
| object.fromentries | 12.0.0 |
| hasown | 16.9.0 |

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -15,6 +15,7 @@
   "rules": {
     "max-len": "off",
     "no-template-curly-in-string": "off",
+    "prefer-object-spread": "off"
   },
   "overrides": [
     {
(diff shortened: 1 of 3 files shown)


ljharb:
I will never be raising the node version requirement.

Object.assign has bugs in node 4, so we can't drop the package.

PR #964, opened 2023-11-10
remove array-includes

dulmandakh:
let's remove support for very old version of Node.js, and array-includes polyfill. And after a while, let's release a new major version.


ljharb:
No, let's not, thanks. Same discussion as in the other repo. This would be a breaking change immediately, and removing a polyfill isn't a valuable enough change to be worth the insanely high cost of a semver-major.

Issue #417, opened 2018-04-06
img-redundant-alt has incorrect rules


Steve-O-Cassels:
What a fantastic plugin this is and I am so grateful for it.  

However from what I can see, the semantics of the [img-redundant-alt](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/img-redundant-alt.md#img-redundant-alt) rule seem to overlook providing users with an equal experience as outlined in WCAG.  

The following reasoning for this rule confuses me:

>  There is no need to use words such as image, photo, and/or picture.  

But surely there is a need for these meta words 'photo' and 'picture' because a screen-reader cannot convey what type of img, an img is?

For example, comparing alt text of `a car race`, with a `a colour photo of a car race at the height of summer` or indeed with, `a black and white photo of car race`, all suggest different perceptions of the car race event; the fact that the image is a photo is critical information about what is presented in and suggested by (black and white photo would suggest its old) the image.  A visual user would get all the nuance of a black and white photo, or recognise an artists picture as a stylised impression of something. Omitting that it is a picture or a photo does not give an equal experience to the visually impaired end-user. 

So for me, `photo` and `picture` should not be out of bounds.  

Or am I misunderstanding?

ljharb:
“A color photo” eeems the default; i think many of these cases can be trivially rewritten such as “a car race at the height of summer” - and a photo being black and white or sienna really doesn’t guarantee that it’s old, what with processing techniques and instagram filters.

It seems like while it’s fine to use these words if they do, in fact, convey extra information - the visual description might not be sufficiently conveying the information you’re assuming.

Steve-O-Cassels:
I'll try to be more clear.  I agree with you, its a nuanced thing the use of alt text, and having 'photo' or 'print', or 'cartoon' is as much about the style of the image as anything else, but that is pertinent.  'image' and 'picture' do seem so vague as to not be helpful in conveying information and I understand that a screen reader will call these out as 'image' anyway.  I do think these two words should continue to cause a lint error.  But not 'photo'.

I also agree that alt text is often abused and poorly written with content authors putting minimal effort into describing an image which results in overly simplistic 'photo of' descriptions.    However I think that linting is not the way to fix that.

[...]

PR #970, opened 2024-01-08
[Fix] `img-redundant-alt`: fixed multibyte character support

### LINKED ISSUE #969: [img-redundant-alt] `words` option does not work with double-byte character words.
If a multibyte character is specified in `words` option, the corresponding character is not treated as an error.
This may be due to the fact that word boundary [`\b`](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/1adec3517fc2c9797212ca4d38858deed917e7be/src/rules/img-redundant-alt.js#L66) do not support multibyte characters.

For example, in Japanese, "image" can be written as "イメージ". However, specifying this in option `words` is not treated as an error.
```js
    'jsx-a11y/img-redundant-alt': ['error', {
      words: ['イメージ'],
    }],
```
```jsx
      <img src='path/to/image.jpg' alt="イメージ" />
```

makotot:
Fixes #969.

Is it possible to replace the word boundary `\b` with `\s+`? Although it has passed unit tests.

Also, it seemed more natural not to check word boundaries for multibyte characters, which may be a difficult point to handle also related to https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/417.

`(?!{)` is removed because there is no difference in the unit test results whether it is there or not.

### DIFF
--- a/__tests__/src/rules/img-redundant-alt-test.js
+++ b/__tests__/src/rules/img-redundant-alt-test.js
@@ -74,6 +74,7 @@ ruleTester.run('img-redundant-alt', rule, {
     { code: '<img alt="ImageMagick" />;' },
     { code: '<Image alt="Photo of a friend" />' },
     { code: '<Image alt="Foo" />', settings: componentsSettings },
+    { code: '<img alt="画像" />', options: [{ words: ['イメージ'] }] },
   )).map(parserOptionsMapper),
   invalid: parsers.all([].concat(
     { code: '<img alt="Photo of friend." />;', errors: [expectedError] },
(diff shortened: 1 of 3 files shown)


ljharb (changes requested):
you'll need `import includes from 'array-includes';` rather than `.includes` but otherwise this looks great!

ljharb:
nvm, i think i figured it out

makotot:
Didn't reply due to sleep time. Thanks 🙌 
 I will wait for a new version to be released that includes this fix.

Issue #961, opened 2023-11-04
  jsx-a11y/label-has-associated-control gives error when for is put 


TheElegantCoding:
This is my config

`'jsx-a11y/label-has-associated-control': 'error'`

i am not in react i use "for", this make me error in the code only when i put htmlFor works but i cant do that i dont use react
```tsx
return (
  <>
    <label for="email" />
    <input id="email" name="email" type="text" />
  </>
)
```

ljharb:
Most jsx renderers match react, since react invented jsx. It’d be reasonable to add a setting that can rename htmlFor as “for”, i suppose.

AesthetiCoder:
not all jsx renderes match that, i am in solidjs so yeah it will be nice

felquis:
It's missing aria-label attribute in the label element

TheElegantCoding:
wating for this support

Issue #894, opened 2022-09-16
Support for custom JSX renderer


lionel-bijaoui:
Hello,
I have created a custom JSX renderer that take JSX code and output HTML.
This custom implementation doesn't use that exact same keyword.
For example, I use "class" instead of "className" and "for" instead of "htmlFor".

I noticed that there is a certain level of customization for the rules, can it be opened to support my needs, or do you want to strictly adhere to React implementation ?

I'm open to do the PR if needed.
Thank you very much for your time.

ljharb:
While `eslint-plugin-react` has settings for custom jsx renderers, this plugin does not.

While we could add settings to map react jsx prop names to custom prop names, that's a lot of complexity to add for what feels like a very unwise use case.

Can you elaborate on why you think deviating from the overwhelmingly defacto standard is worth it?

lionel-bijaoui:
Ok, I will try to list some arguments, I hope this is enough to convince you of this need :
1. While JSX was created by React, [its specs](https://facebook.github.io/jsx/#sec-jsx-attributes) does not force to use the same attributes as React. That would limit the use and scope of JSX.
2. By this logic, every JSX system allow for custom renderers (typescript, babel, etc.). If JSX was limited to use with React, that would not be the case.
3. Preact use [slightly different attributes conventions](https://preactjs.com/guide/v8/differences-to-react/#whats-added) than React. While React wished to adapt the HTML to be compatible with JS, Preact choose to more closely follow the HTML. I have chosen the same way, since my team is more accustomed to working with HTML, and our existing template use HTML syntax. Why call `for` as `forHTML`, or `class` as `className` ? How is this a standard compared to HTML ?
4. If this plugin was called `eslint-plugin-react-jsx-a11y` I would agree that my demand made no sense, but the fact that `eslint-plugin-react` support custom jsx renderers, I see little reason why not (if not the complexity, which I understand of course and I not trying to underestimate it).

Again, I understand that this if fairly complex, but I don't think that is is completely out of the scope of this plugin. I willing to help, since I'm very curious about writing custom eslint rules and better understanding AST.

Anyway, thank you for your time and allowing me to state my reasoning. I hope this is enough, but if this is not, I understand and thank you for your work on this plugin raising awareness on a11y.

ljharb:
Because for and class are reserved words, so you can’t destructure into those variable names.

I hear your arguments, and it’s worth thinking about.

PR #977, opened 2024-04-08
Support `for` attribute

edoardocavazza:
This PR introduces support for `for` attribute as equivalent for the `htmlFor` property.

Related issues: #894, #961

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,9 @@ Add `plugin:jsx-a11y/recommended` or `plugin:jsx-a11y/strict` in `extends`:
         "CustomButton": "button",
         "MyButton": "button",
         "RoundButton": "button"
+      },
+      "attributes": {
+        "for": ["htmlFor", "for"]
       }
     }
   }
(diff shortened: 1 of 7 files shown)


ljharb:
This must not be done by default, since that's not how react (the original, default, and most common usage of jsx) works.

I think a global setting should be used to map `htmlFor` to `for`, and rules updated to respect that setting - that way, custom renderers can map whatever attributes they want.

edoardocavazza:
In most of the rendering engine, both `for` (as attribute) and `htmlFor` (as property) are valid. Maybe the global setting should be an array of property names to check?

ljharb:
React's is the primary one that matters since it has almost the entirety of the usage, but it's good to design the schema so it can support all the renderers.

how about an object, which defaults to `{ for: ['htmlFor'], 'class': ['className'] }` etc, and then you can set it to any combination you want?

edoardocavazza:
I think it would fit greatly, but I have some issues creating the test cases for the case :(

ljharb:
I’d prefer to reuse this PR (but now that you’ve opened a second one, please keep both open)

PR #984, opened 2024-05-06
[New] allow polymorphic linting to be restricted to specified components

khiga8:
#### Change

This builds on the setting introduced in https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/pull/945.

This change allows the consumer to restrict the polymorphic setting to a specified list of components via a new optional setting,`polymorphicAllowlist`. 

* When `polymorphicAllowlist` is undefined, the polymorphic prop will be used _for any components_ of the prop name specified by `polymorphicPropName` is present. 
* When `polymorphicAllowlist` is defined, then there is an additional check to make sure that components are part of the `polymorphicAllowlist` before we use the polymorphic prop specified by `polymorphicPropName`.

#### Motivation

Linting components can raise false positives when a component handles behavior that the linter has no way to know. 

For example, an `Avatar` component may render as an `img` by default and automatically render an `alt` based on a username. The linter may raise a false positive if `Avatar` is linted as an `img` for the alt text lint rule. This can be avoided by not adding the `Avatar` to the `jsx-a11y` component map. If`Avatar` (for whatever reason) explicitly has `as="img"` set, since it can also be rendered as an `svg`, it will end up automatically linted via the `polymorphicPropName` setting.

In some projects, polymorphic linting may be useful but it may be safer on utility/basic components that don't do much (e.g. a [generic `Box` element](https://mui.com/material-ui/react-box/)), rather than opening it up to any component that allows the polymorphic prop to be set. 

I acknowledge that polymorphism is not an ideal pattern that can add complexity (as warned in the README).

### DIFF
--- a/src/util/getElementType.js
+++ b/src/util/getElementType.js
@@ -4,6 +4,7 @@
 
 import type { JSXOpeningElement } from 'ast-types-flow';
 import hasOwn from 'hasown';
+import includes from 'array-includes';
 import { elementType, getProp, getLiteralPropValue } from 'jsx-ast-utils';
 
 import type { ESLintContext } from '../../flow/eslint';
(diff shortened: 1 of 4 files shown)


src/util/getElementType.js
```diff
@@ -11,11 +11,21 @@ import type { ESLintContext } from '../../flow/eslint';
...
     const polymorphicProp = polymorphicPropName ? getLiteralPropValue(getProp(node.attributes, polymorphicPropName)) : undefined;
-    const rawType = polymorphicProp ?? elementType(node);
+
+    let rawType = elementType(node);
+    if (polymorphicProp) {
+      if (!polymorphicAllowList) {
+        rawType = polymorphicProp;
+      } else if (polymorphicAllowList.includes(rawType)) {
+        rawType = polymorphicProp;
+      }
```
ljharb, line 27:
```suggestion
      if (!polymorphicAllowList || polymorphicAllowList.includes(rawType)) {
        rawType = polymorphicProp;
      }
```

also this needs to use `import includes from 'array-includes';` instead of `.includes`

ljharb (changes requested):

ljharb:
we'll also need documentation.

khiga8:
This is ready for another review 🙇‍♀️

ljharb (approved):
i still find polymorphic components horrific, and i hope you're able to fix your codebase by removing them :-) but let's land this

Issue #989, opened 2024-06-04
Upgrade `axe-core` for more features (it's pinned right now)


norgeous:
Would it be possible to unpin `axe-core` dep?

https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/package.json#L82

Currently it is pinned to `=4.7.0` in `eslint-plugin-jsx-a11y`

The latest available is `4.9.1` (at time of writing)

https://github.com/dequelabs/axe-core/blob/develop/CHANGELOG.md

Personally, I am interested in any version above `4.7.1`, as that adds support for CSS4 colors such as lch, oklch, lab, oklab

Thanks!

norgeous:
related: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/792

ljharb:
The reason we're pinned to v4.7.0 is due to failing tests; see https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/commit/8d8f0169dbaaa28143cf936cba3046c6e53fa134 and https://github.com/dequelabs/axe-core/issues/4127

If indeed v4.9.0 fixed that, then a PR to unpin it would be great!

ljharb:
You'd need to run the tests with node 4 or 6, yes.

norgeous:
After a bit of dicking about with versions of node, npm and eslint

`npm i -g npm@4; npm i -g npm@6; npm -i eslint@3`

I did some local testing:

| node version | npm version | eslint version | axe-core version | test result | note |
|---------------|----------------|---------------|--------------------|------------|------|
| v6.17.1 | v6.14.16 | 3.19.0 | =4.7.0 | ✅66 | same as `main` branch (just checking) |
| v6.17.1 | v6.14.16 | 3.19.0 | =4.7.1 | ❌8 ✅58| my preference |

The 8 failing tests all fail with the same error (coming from `axe-core` imports):

```
TypeError: Object.values is not a function
```

so yeah, there is a problem with `axe-core` higher than the current pinned version when using node v6! 👍

ljharb:
node versions are what matters, not npm versions, but let's look at the tests on the PR :-)

ljharb:
Fixed in #990.

Issue #792, opened 2021-05-13
axe-core version need to update for latest version to support aria-role = text


ShruthiTP08:
axe-core version need to update for latest version to support aria-role = text

ljharb:
What version is needed?

jackstevenson:
`aria-role="text"` support was added in [axe-core 4.2.0](https://github.com/dequelabs/axe-core/releases/tag/v4.2.0)

ljharb:
Sounds like an issue filed on aria-query would be the first step :-)

PR #990, opened 2024-06-05
[Deps] unpin `axe-core`

norgeous:
previously =4.7.0, now ^4.9.1

see https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/989
and https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/792


ljharb (approved):
Seems good, thanks!

PR #993, opened 2024-06-18
feat: add support for Flat Config

michaelfaith:
This change adds support for ESLint's new Flat config system.  It maintains backwards compatibility with eslintrc
style configs as well.

To achieve this, we're now dynamically creating four configs: two are the original `recommended` and `strict`, and the other two are the new `flat/recommended` and `flat/strict`.  The two `flat` ones are setup with the new config format, while the original two have the same options as before.  

# Usage
## Legacy
```json
{
  "extends": ["plugin:jsx-a11y/recommended"]
}
```

## Flat
```js
import globals from 'globals';
import js from '@eslint/js';
import jsxA11y from 'eslint-plugin-jsx-a11y';

export default [
  js.configs.recommended,
  jsxA11y.flatConfigs.recommended,
  {
    files: ['**/*.{js,mjs,cjs,jsx,mjsx,ts,tsx,mtsx}'],
    languageOptions: {
      ecmaVersion: 'latest',
      sourceType: 'module',
      globals: globals.browser,
    },
    ignores: ['dist', 'eslint.config.js'],
    rules: {
      'no-unused-vars': 'warn',
      'jsx-a11y/anchor-ambiguous-text': 'warn',
      'jsx-a11y/anchor-is-valid': 'warn',
    },
  },
];
```

And just to be clear, this does *not* close the v9 support Issue, as there may be additional work needed on the rules themselves to enable compatibility with v9.  This change is solely focused on providing support for the new Flat Config.


michaelfaith:
Looks like the rule doc generator may need new icons for the two new flat configs.  Not sure if you want to use the same check and lock as the legacy configs.  Since the rulesets are the same, it does seem a bit redundant to have both in the table, but I'm not sure how you'd want to represent that.

michaelfaith:
I went ahead and set the two `flat/` configs to ignore on the doc generator, since they're mirrors of `recommended` and `strict`

.eslintignore
```diff
@@ -0,0 +1 @@
+examples/
```
ljharb, line 1: this should be added in eslintrc's `ignorePatterns` instead of in a separate file

README.md
```diff
@@ -60,7 +60,9 @@ yarn add eslint-plugin-jsx-a11y --dev
 
 **Note:** If you installed ESLint globally (using the `-g` flag in npm, or the `global` prefix in yarn) then you must also install `eslint-plugin-jsx-a11y` globally.
 
-## Usage
+<a id="usage"></a>
+
```
ljharb, line 64:
```suggestion
```

README.md
```diff
@@ -109,6 +111,94 @@ Add `plugin:jsx-a11y/recommended` or `plugin:jsx-a11y/strict` in `extends`:
...
+- `flat/strict`
+- `flat/recommended`
+
+#### CJS
+
+```js
+const jsxA11y = require('eslint-plugin-jsx-a11y');
+
+export default [
+  jsxA11y.configs['flat/recommended'],
```
ljharb, line 158: this might be nicer to use as a `flat` object, with a `recommended` property?

examples/flat-cjs/.gitignore
```diff
@@ -0,0 +1,24 @@
+# Logs
+logs
+*.log
+npm-debug.log*
```
ljharb, line 4: any reason this isn't just part of the root gitignore?

examples/flat-cjs/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "devDependencies": {
+    "@eslint/js": "^9.5.0",
+    "@types/react": "^18.2.66",
+    "@types/react-dom": "^18.2.22",
+    "@vitejs/plugin-react": "^4.2.1",
+    "cross-env": "^7.0.3",
+    "eslint": "^8.57.0",
+    "eslint-plugin-jsx-a11y": "file:../..",
+    "globals": "^15.6.0",
+    "vite": "^5.2.0"
```
ljharb, line 25: do we really need all these deps in the repo to showcase eslint config code?

package.json
```diff
@@ -29,9 +29,13 @@
     "test": "npm run jest",
     "posttest": "aud --production",
     "test:ci": "npm run jest -- --ci --runInBand",
+    "test:examples": "npm run build && concurrently -c \"auto\" \"npm:test-example:*\"",
```
ljharb, line 32: my experience with concurrently and alternatives is that it's too easy for failures to silently pass - i'd rather just run them explicitly and in serial.
```suggestion
    "pretest:examples": "npm run build",
    "test:examples": "concurrently -c \"auto\" \"npm:test-example:*\"",
```

README.md
michaelfaith: I like that.  I was going back and forth with that, and landed on this approach because I saw a few others doing it.  But there's no real strong need to have them added to the legacy `configs` object.  I'll make that update.

examples/flat-cjs/.gitignore
michaelfaith: This is just what the `vite` creation script generates.  So, in essence, the "app" is self-contained in that folder.  I'm happy to move them up to the root, if you prefer.

examples/flat-cjs/package.json
michaelfaith: I guess I was aiming to have it be as "real world" of an example as possible.  I just generated a new React app using `vite` and left everything as is, except for the lint config.

package.json
michaelfaith: That's fair.  I went with `concurrently` for the opposite reason.  If they were run serially, then it fails on the first failure, which may not expose the fact that one of the other is failing too (e.g. you fix one, only to find out the next one failed).  But I'm happy to switch that too.

examples/flat-cjs/.gitignore
ljharb: I'd prefer that examples not necessarily be fully runnable apps - that's a lot of extra complexity in this repo. All we need is enough guidance to teach somebody how to use our plugin in _their_ app.

examples/flat-cjs/package.json
ljharb: we don't need a working app to demonstrate linting, though.

package.json
ljharb: Right - in theory concurrently is great. However, in practice, I've found that one of the parallel tasks can fail, without failing the overarching CI, allowing bugs to be merged, so i'd rather suffer the tradeoff of serial failures than silent ones.

README.md
michaelfaith: What do you think about `flatConfigs` rather than `flat` so that it parallels the `configs` prop?

michaelfaith:
```js
// Create configs to plugin object
const configs = {
  recommended: createConfig(recommendedRules),
  strict: createConfig(strictRules),
};

const flatConfigs = {
  recommended: createConfig(recommendedRules, 'recommended'),
  strict: createConfig(strictRules, 'strict'),
};

module.exports = { ...jsxA11y, configs, flatConfigs };
```

examples/flat-cjs/package.json
michaelfaith: I've removed `vite`, its config, and associated build/test commands (along with the boilerplate README and .gitignore).  I did leave `react`, `react-dom` and their associated types.

examples/flat-cjs/.gitignore
michaelfaith: I've removed the unneeded `vite` boilerplate

README.md
michaelfaith: I went ahead and updated with the above.  Let me know what you think about that.

.eslintignore
michaelfaith: Updated

examples/flat-cjs/package.json
michaelfaith: Went ahead and removed the `@types/*` too.

ljharb (approved):
LGTM - i'll make some tweaks so we can run the examples tests in CI, but otherwise this seems like it'll work.

ljharb:
v6.9.0 has been released.

dubzzz:
Thanks a lot for this awesome addition. Flat config is game changer.

You should probably update the description of the PR and replace `jsxA11y.configs['flat/recommended']` by `jsxA11y.flatConfigs.recommended`.

PR #995, opened 2024-06-24
[readme] remove deprecated travis ci badge

trajan0x:
removes deprecated travis CI badge


ljharb (approved):
thanks! i've tweaked this to replace it with a github actions badge instead of just removing it

PR #1003, opened 2024-08-15
[readme] fix jsxA11y import name

k35o:
Changed so that the imported variable name matches the variable name used.


ljharb (approved):
thanks

PR #1004, opened 2024-08-22
[Fix]: `label-has-associated-control`: ignore undetermined label text

### LINKED ISSUE #966: [label-has-associated-control] regression - rule errors when a label does not directly have text, even if it has htmlFor
```jsx
function Foo() {
  return (
    <>
      <label className={styles.linkLabel} htmlFor="1">
        <CustomText />
      </label>
      <input id="1" />
    </>
  );
}
```

In 6.7.1 this code was fine and did not error.
In 6.8.0 this code results in an error.

BillyLevin:
Attempt to fix #966 

The rule no longer errors if the existence of label text cannot be determined. The criteria for not being able to determine whether the text exists are, for each child of the label:

1. It is a React component
2. It has no children
3. It is not registered as a control component. Unfortunately this means it would still error if the control component itself displays text, but I didn't see a way around that

I don't think this accounts for every case, but hopefully a good starting point

I followed the convention already used by making the new parameters for `mayHaveAccessibleLabel` optional, even though they're not. I assume that was to make it easier to write the unit tests. Happy to change that though if you'd like


ljharb (approved):
Seems good! I'll confirm the tests fail without the fix, and then merge.

PR #1007, opened 2024-08-23
[New] `label-has-associated-control`: add additional error message

### LINKED ISSUE #1005: `label-has-associated-control` - misleading error message
This rule checks two things:

1. A label has an associated control component
2. The label contains accessible text

However, the error message never accounts for the second case. So, if we had this code:

```jsx
function Thing() {
  return (
    <div>
      <label htmlFor="js_id" />
      <input id="js_id" />
    </div>
  );
}
```

We get the error message: `A form label must be associated with a control.`, which is not actually what the issue is.

I think it would be helpful to have a separate error message for this case, something like: `A form label must have accessible text.` to clear up any confusion between the two cases

BillyLevin:
Fixes #1005 by introducing a second error message, used when the label doesn't have accessible text.

### DIFF
--- a/__tests__/src/rules/label-has-associated-control-test.js
+++ b/__tests__/src/rules/label-has-associated-control-test.js
@@ -26,6 +26,11 @@ const expectedError = {
   type: 'JSXOpeningElement',
 };
 
+const expectedErrorNoLabel = {
+  message: 'A form label must have accessible text.',
+  type: 'JSXOpeningElement',
+};
+
 const componentsSettings = {
   'jsx-a11y': {
     components: {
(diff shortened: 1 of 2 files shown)


ljharb (approved):
Seems good, thanks!

amethystens (approved):

__tests__/src/rules/label-has-associated-control-test.js
```diff
@@ -26,6 +26,11 @@ const expectedError = {
   type: 'JSXOpeningElement',
 };
 
+const expectedErrorNoLabel = {
```
amethystens, line 29: 29 const expectedErrorNoLabel ={

amethystens (approved):

amethystens (approved):

amethystens (approved):

amethystens (approved):

PR #1009, opened 2024-08-30
[New] add eslint 9 support

### LINKED ISSUE #978: Add support for ESLint 9
[ESLint 9 is released](https://eslint.org/blog/2024/04/eslint-v9.0.0-released/), but it looks like `eslint-plugin-jsx-a11y` does not yet support it.

There _are_ breaking API changes: https://eslint.org/docs/latest/use/migrate-to-9.0.0#breaking-changes-for-plugin-developers

There is also a new default config format ("flat config") that probably also needs some adjustment in the exported `recommended` rules; though, support for flat config could be considered a separate issue.

michaelfaith:
This change adds support for eslint v9.  All three example projects have been updated to use the latest eslint, and the root package's dev and peer deps have been updated. 

In order to make this work with both v9 and older versions, I had to update the parser options helper to adjust the config passed into the `RuleTester`. It was also necessary update jest to a newer version, in order to resolve eslint's use of `node:fs/promises`. 

image

Note: the removed test cases are exact duplicates of other tests.  v9 fails tests when its determined to be an exact duplicate.

Closes: #978


package.json
```diff
@@ -25,7 +25,7 @@
     "lint": "eslint --ext=js,mjs,cjs,ts,tsx .",
     "prepublish": "not-in-publish || npm run prepublishOnly",
     "prepublishOnly": "safe-publish-latest && npm run lint && npm run flow && npm run jest",
-    "pretest": "npm run lint:fix && npm run flow",
+    "pretest": "npm run flow",
```
michaelfaith, line 28: This was necessary since the root eslint config would need to be overhauled in order to run lint for the whole project with v9, and that didn't seem appropriate to be in scope for this change.  Not sure if that means linting should be done explicitly in CI with the default eslint version.

__tests__/src/rules/no-static-element-interactions-test.js
```diff
@@ -85,7 +85,6 @@ const alwaysValid = [
   { code: '<audio onClick={() => {}} />;' },
   { code: '<form onClick={() => {}} />;' },
   { code: '<form onSubmit={() => {}} />;' },
-  { code: '<link onClick={() => {}} href="#" />;' },
```
ljharb, line 88: none of the removals in this file seem like duplicates

package.json
```diff
@@ -63,7 +63,7 @@
     "flow-bin": "^0.147.0",
     "in-publish": "^2.0.1",
     "jackspeak": "=2.1.1",
-    "jest": "^24.9.0",
+    "jest": "^29.7.0",
```
ljharb, line 66: we can't upgrade jest, because jest 24 is the last version that supports node 6.

ljharb (changes requested):

package.json
ljharb: can the lint script just run `npx eslint@8 --ext=js,mjs,cjs,ts,tsx .`, rather than this change?

michaelfaith: Done

package.json
michaelfaith: Hmm. Ok, but jest 24 isn't able to resolve `node:fs/promises` which eslint uses (so we hit this when importing `RuleTester` from eslint.  I could map node:fs/eslint to something like fs-extra using jest's `moduleNameMapper`?  So the project is never going higher than jest 24?

__tests__/src/rules/no-static-element-interactions-test.js
michaelfaith: Interesting.  You're right.  Looks like it was a rebasing issue.  Those were added in https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/commit/75d5dd722bd67186d97afa7b151fd6fee5885c70#diff-b8a286945cae1102f315e16a85e3de23850073d6532bb969c8363bc98d61aa20 after this branch was originally created.  I had an issue where my vscode formatted this file, so I while trying to roll that back, I must have replace it with an earlier commit. I'll address this

michaelfaith: Fixed.  Though there was one legit removal `<div onClick={() => {}} />`

package.json
ljharb: at some point we'll have a breaking change that drops older eslints/nodes, but i'm more likely to bail on jest entirely and use something less frictiony regardless.

can you map `node:fs/promises` to `fs/promises`, instead? the `node:` prefix is pretty useless.

michaelfaith: After a few iterations, I got it working.  I ended up having to map several built in no modules, as well as a couple of esm entry points of eslint packages, I also had to add a polyfill for `structuredClone`, which eslint uses as part of `RuleTester` and was added in node 17.

michaelfaith: Well, it cleared up everything locally, but it's breaking like hell in CI now.

michaelfaith: Even walking through the steps in CI, locally, still works.  The main difference, though, is I'm on a Windows machine. But that shouldn't come into play here.

pauliesnug: honestly i don't know anyone who uses anything older than node 16 at this point. it's standard to have deprecated support for older versions, and would make things a lot easier with vitest and the sorts

michaelfaith: What if we dynamically installed jest 29 in ci for eslint 9, in the same way we're dynmically installing all of the supported eslint versions?  That way we don't have to set up all of this hacky stuff?

michaelfaith: Actually, I think I got it working.

michaelfaith: Everything's green now, if you want to take another look when you have the time.  It may be worth considering, dynamically installing a newer jest for v9 and removing all of those workarounds (in the same way each eslint version is being installed), but I'll leave that up to you.

package.json
```diff
@@ -51,7 +51,8 @@
     "babel-jest": "^24.9.0",
     "babel-plugin-add-module-exports": "^1.0.4",
     "babel-preset-airbnb": "^5.0.0",
-    "eslint": "^3 || ^4 || ^5 || ^6 || ^7 || ^8",
+    "core-js": "^3.38.1",
```
ljharb, line 54: is there a smaller package we could get a structuredClone polyfill from?

__tests__/__util__/nodeReexports/assert.js
```diff
@@ -0,0 +1,3 @@
+import assert from 'assert';
+
+export default assert;
```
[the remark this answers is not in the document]
michaelfaith: That violates an eslint rule in the project.  I could turn the rule off and change those, if you want.  I was just trying to abide by the project's linting rules. 
image

package.json
michaelfaith: https://www.npmjs.com/package/@ungap/structured-clone would work, if you prefer it.

package.json
```diff
@@ -25,7 +25,7 @@
     "lint": "eslint --ext=js,mjs,cjs,ts,tsx .",
     "prepublish": "not-in-publish || npm run prepublishOnly",
     "prepublishOnly": "safe-publish-latest && npm run lint && npm run flow && npm run jest",
-    "pretest": "npm run lint:fix && npm run flow",
+    "pretest": "npx eslint@8 --ext=js,mjs,cjs,ts,tsx . --fix && npm run flow",
```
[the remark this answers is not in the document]
michaelfaith: Done

__tests__/__util__/nodeReexports/assert.js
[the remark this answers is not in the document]
ljharb: ah, yeah this is the special case that rule should allow :-) not a huge deal either way, since I’m probably going to ditch jest and will be able to delete the mocks.

package.json
michaelfaith: swapped `core-js` for `@ungap/structured-clone`

michaelfaith: Swapped back.  that package uses `globalThis` which isn't available in earlier node versions.  `core-js` is the best option here

ljharb: ok, this is just another reason to drop jest :-)

ljharb (approved):
Seems good!

iamtomcat:
When is this going out? Currently just using it from the repository directly.

ljharb:
@iamtomcat i'm not sure how you'd do that in a practical sense, since we have a build process. It's only been 2 days - I'd suggest not using it at all until it's published.

iamtomcat:
Hey @ljharb I appreciate you getting the change out, I wasn't trying to rush you, just wanted to get a time frame was all.

Thanks again.

ljharb:
it's now released; v6.10.0.

ruiaraujo012:
@ljharb I think something is missing, there's no `v6.10.0` release.

ljharb:
oh thanks, i always forget github releases (which don't matter; only npm matters)

PR #1013, opened 2024-10-06
【Fix】: alt-text string null determination

### LINKED ISSUE #971: `alt-text`: missing warning for empty `alt` for `<input type="image" alt="" />`
`<input type="image" alt="" />` should be considered invalid because `<input>` is an interactive element, it can't be "decorative", thus it has to have a non-empty `alt`.

Note: SonarJS does warn in this case, see "Noncompliant code example" here: https://sonarsource.github.io/rspec/#/rspec/S1077/javascript

taku10101:
closed #971

### DIFF
--- a/src/rules/alt-text.js
+++ b/src/rules/alt-text.js
@@ -39,7 +39,7 @@ const ariaLabelHasValue = (prop) => {
   if (value === undefined) {
     return false;
   }
-  if (typeof value === 'string' && value.length === 0) {
+  if (typeof value === 'string' && value.length === 0 && value === '') {
     return false;
   }
   return true;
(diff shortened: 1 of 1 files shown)


src/rules/alt-text.js
```diff
@@ -39,7 +39,7 @@ const ariaLabelHasValue = (prop) => {
   if (value === undefined) {
     return false;
   }
-  if (typeof value === 'string' && value.length === 0) {
+  if (typeof value === 'string' && value.length === 0 && value === '') {
```
ljharb, line 42: `value === ''` and `value.length === 0`, when it's already a string, are the same condition.
```suggestion
  if (typeof value === 'string' && value.length === 0) {
```

ljharb (changes requested):
This would need some tests to show what bug is being fixed.

PR #1018, opened 2024-10-18
Use consistent spelling of 'screen reader'

lb-:
As per PR #725

All 'screen reader' products use two words, in addition, it's helpful to avoid compound words for screen reader accessibility when possible.

The current code is about 50/50 on this but there were slightly more usages of the two word spelling, plus a previous PR.


ljharb (approved):

lb-:
Thank you @ljharb

Issue #1002, opened 2024-08-07
aria-label should not be required when role="heading"


lifeiscontent:
as described here: https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA12#example-1-simple-headings

lb-:
I have confirmed that `aria-label` is **not** required when using `role="heading"`. See https://www.w3.org/TR/wai-aria/#heading

> heading role
> A heading for a section of the page.
> 
> To ensure elements with a role of heading are organized into a logical outline, authors MUST use the [aria-level](https://www.w3.org/TR/wai-aria/#aria-level) attribute to indicate the proper nesting level.

However, I am unable to reproduce the linting issue as described. This code below **correctly** produces **no** warning.

```jsx
/* eslint
  "jsx-a11y/role-has-required-aria-props": "warn",
*/

const Heading = () => (
  <div role="heading" aria-level={3}>
    Vite + React
  </div>
);

export default Heading;

```

Nonetheless I will prepare a PR that includes validity testing for this scenario.

lb-:
We can probably close this issue now. PR with reinforcement tests have been added.

ljharb:
Closed by #1022.

PR #1022, opened 2024-10-23
[Tests] Add tests to reinforce required attributes for `role="heading"`

lb-:
Add unit tests to ensure that there are no warnings when role="heading" is used when aria-level is provided. For example, aria-label is not required.

See #1002


ljharb (approved):
assuming tests pass

MichaelDeBoey (approved):

Issue #936, opened 2023-05-31
`no-redundant-roles` should allow (SVG) images with `role="img"`


saskaak:
Hello!

As far as I know`<img src="some-image.svg" role="img">` is an actual pattern that can be used to increase browser/screenreader support for SVG images specifically.

Thus, `no-redundant-roles` should allow images with role="img".

See the difference here comparing pattern 1 (no explicit role) and pattern 2 (with explicit role): https://www.smashingmagazine.com/2021/05/accessible-svg-patterns-comparison/

preetsahil:
Is someone working on this issue ?
can you assign it to me.

ljharb:
@preetsahil that's not how things work on github; issues aren't assigned like that. just make a PR if you want to work on it.

preetsahil:
Okay I will try to work on it

lb-:
Looks like there is a Safari bug that's the cause of this recommendation, Safari have said it's fixed but it appears to be an issue still.

* See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#identifying_svg_as_an_image
* See https://bugs.webkit.org/show_bug.cgi?id=216364

I will see if I can prepare a PR that changes the implicit role for `img` if src contains svg. This will not work for variables but may be good enough.

lb-:
PR up #1026

PR #1026, opened 2024-10-23
[patch] `no-redundandant-roles`: allow `<img src="*.svg" role="img" />`

lb-:
Setting role="img" is a valid use case to work around a Safari bug for better accessibility when the source image is an SVG file.

This improvement does not account for variables in the `src` attribute but adds a valid exception for when we can parse the string. In addition, add valid `aria-role` test for `<svg role="img" />` to ensure no future conflicts arise.

Closes [#936](https://github.com/lb-/eslint-plugin-jsx-a11y/issues/936)


ljharb (approved):
This seems good (but is a patch, since it's reducing warnings)

lb-:
Thanks @ljharb

PR #1027, opened 2024-10-23
[Fix] `no-redundant-roles`: Refine implicit role of `select` to include `combobox` scenarios

### LINKED ISSUE #949: implicitRoles/select.js: implicit role is now "combobox"
Hi,

With the Aria 1.2 update, the implicit role of select element is now "combobox".

lb-:
Encode implicit roles for `select` elements based on roles defined in https://www.w3.org/TR/html-aria/#el-select

- `select` (with a multiple attribute or a size attribute having value greater than 1) will have the implicit role 'listbox'
- `select` (with NO multiple attribute and NO size attribute having value greater than 1) will have the implicit role 'combobox'

Fixes #949

### DIFF
--- a/__tests__/src/rules/no-redundant-roles-test.js
+++ b/__tests__/src/rules/no-redundant-roles-test.js
@@ -41,12 +41,31 @@ const alwaysValid = [
   { code: '<MyComponent role="button" />' },
   { code: '<button role={`${foo}button`} />' },
   { code: '<Button role={`${foo}button`} />', settings: componentsSettings },
+  { code: '<select role="menu"><option>1</option><option>2</option></select>' },
+  { code: '<select role="menu" size={2}><option>1</option><option>2</option></select>' },
+  { code: '<select role="menu" multiple><option>1</option><option>2</option></select>' },
 ];
 
 const neverValid = [
-  { code: '<button role="button" />', errors: [expectedError('button', 'button')] },
   { code: '<body role="DOCUMENT" />', errors: [expectedError('body', 'document')] },
+  // button - treated as button by default
+  { code: '<button role="button" />', errors: [expectedError('button', 'button')] },
   { code: '<Button role="button" />', settings: componentsSettings, errors: [expectedError('button', 'button')] },
+  // select - treated as combobox by default
+  { code: '<select role="combobox"><option>1</option><option>2</option></select>', errors: [expectedError('select', 'combobox')] },
+  { code: '<select role="combobox" size="" />', errors: [expectedError('select', 'combobox')] },
+  { code: '<select role="combobox" size={1} />', errors: [expectedError('select', 'combobox')] },
+  { code: '<select role="combobox" size="1" />', errors: [expectedError('select', 'combobox')] },
+  { code: '<select role="combobox" size={null}></select>', errors: [expectedError('select', 'combobox')] },
+  { code: '<select role="combobox" size={undefined}></select>', errors: [expectedError('select', 'combobox')] },
+  { code: '<select role="combobox" multiple={undefined}></select>', errors: [expectedError('select', 'combobox')] },
+  { code: '<select role="combobox" multiple={false}></select>', errors: [expectedError('select', 'combobox')] },
+  { code: '<select role="combobox" multiple=""></select>', errors: [expectedError('select', 'combobox')] },
+  // select - treated as listbox when multiple OR size > 1
+  { code: '<select role="listbox" size="3" />', errors: [expectedError('select', 'listbox')] },
+  { code: '<select role="listbox" size={2} />', errors: [expectedError('select', 'listbox')] },
+  { code: '<select role="listbox" multiple><option>1</option><option>2</option></select>', errors: [expectedError('select', 'listbox')] },
+  { code: '<select role="listbox" multiple={true}></select>', errors: [expectedError('select', 'listbox')] },
 ];
 
 ruleTester.run(`${ruleName}:recommended`, rule, {
--- a/src/util/implicitRoles/select.js
+++ b/src/util/implicitRoles/select.js
@@ -1,6 +1,18 @@
+import { getProp, getLiteralPropValue } from 'jsx-ast-utils';
+
 /**
- * Returns the implicit role for a select tag.
+ * Returns the implicit role for a select tag depending on attributes.
+ *
+ * @see https://www.w3.org/TR/html-aria/#el-select
  */
-export default function getImplicitRoleForSelect() {
-  return 'listbox';
+export default function getImplicitRoleForSelect(attributes) {
+  const multiple = getProp(attributes, 'multiple');
+  if (multiple && getLiteralPropValue(multiple)) {
+    return 'listbox';
+  }
+
+  const size = getProp(attributes, 'size');
+  const sizeValue = size && getLiteralPropValue(size);
+
+  return sizeValue > 1 ? 'listbox' : 'combobox';
 }
--- a/__tests__/src/util/implicitRoles/select-test.js
+++ b/__tests__/src/util/implicitRoles/select-test.js
@@ -1,0 +1,158 @@
+import test from 'tape';
+
+import JSXAttributeMock from '../../../../__mocks__/JSXAttributeMock';
+import getImplicitRoleForSelect from '../../../../src/util/implicitRoles/select';
+
+test('isAbstractRole', (t) => {
+  t.test('works for combobox', (st) => {
+    st.equal(
+      getImplicitRoleForSelect([]),
+      'combobox',
+      'defaults to combobox',
+    );
+
+    st.equal(
+      getImplicitRoleForSelect([JSXAttributeMock('multiple', null)]),
+      'combobox',
+      'is combobox when multiple attribute is set to not be present',
+    );
+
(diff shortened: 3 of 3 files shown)


MichaelDeBoey (approved):

__tests__/src/rules/no-redundant-roles-test.js
```diff
@@ -41,12 +41,19 @@ const alwaysValid = [
...
 const neverValid = [
   { code: '<button role="button" />', errors: [expectedError('button', 'button')] },
   { code: '<body role="DOCUMENT" />', errors: [expectedError('body', 'document')] },
   { code: '<Button role="button" />', settings: componentsSettings, errors: [expectedError('button', 'button')] },
+  { code: '<select role="combobox" size />', errors: [expectedError('select', 'combobox')] },
+  { code: '<select role="combobox" size={1} />', errors: [expectedError('select', 'combobox')] },
+  { code: '<select role="combobox"><option>1</option><option>2</option></select>', errors: [expectedError('select', 'combobox')] },
+  { code: '<select role="listbox" size="3" />', errors: [expectedError('select', 'listbox')] },
+  { code: '<select role="listbox" size={2} />', errors: [expectedError('select', 'listbox')] },
+  { code: '<select role="listbox" multiple><option>1</option><option>2</option></select>', errors: [expectedError('select', 'listbox')] },
```
ljharb, line 56: can we add a test case for `multiple={false}` and `multiple={undefined}`?

lb-: Good find, I had not thought about where `undefined` or `null` will not set the attribute in React.

I have updated the unit tests & added explicit tests for the implicit role resolver as the code tests will treat `multiple={null}` as the string `'null'` not the value `null`.

Ready for another review.

ljharb: What about false? It’s a boolean attribute, so setting it to false should count as “not multiple”, i think?

lb-: For this spec, from my reading & manual testing, the value of the attribute makes no difference.

```html
<select id="shakes" name="shakes" multiple>
  <option>Vanilla Shake</option>
  <option>Strawberry Shake</option>
  <option>Chocolate Shake</option>
</select>
```

is treated the same as

```html
<select id="shakes" name="shakes" multiple="false">
  <option>Vanilla Shake</option>
  <option>Strawberry Shake</option>
  <option>Chocolate Shake</option>
</select>
```

## Firefox

Treats the same `listbox`, irrespective of the value set to `multiple` as long as the attribute is set.

Screenshot 2024-10-26 at 9 50 04 PM
Screenshot 2024-10-26 at 9 50 32 PM

## Chrome

Treats the same `listbox`, irrespective of the value set to `multiple` as long as the attribute is set.

Screenshot 2024-10-26 at 9 58 04 PM
Screenshot 2024-10-26 at 9 58 28 PM

## Safari

Treats the same `listbox`, irrespective of the value set to `multiple` as long as the attribute is set.

Screenshot 2024-10-26 at 9 54 36 PM

Screenshot 2024-10-26 at 10 00 31 PM

lb-: This also aligns with my understanding of how a Boolean attribute should behave. 

https://developer.mozilla.org/en-US/docs/Glossary/Boolean/HTML

> If an HTML tag contains a boolean attribute — no matter the value of that attribute — the attribute is set to true on that element. If an HTML tag does not contain the attribute, the attribute is set to false.

ljharb: `multiple={false}` isn't the same as `multiple="false"` in react - i believe the former renders identical to `multiple={undefined}`.

lb-: I also did manual testing and I'm not sure I agree. 

- `multiple={false}`  -> `multiple="false"`
- `multiple="false"`  -> `multiple="false"`
- `multiple={null}`  -> (multiple is not present)
- `multiple={undefined}`  -> (multiple is not present)

https://react-cn.github.io/react/tips/false-in-jsx.html

ljharb: That link doesn’t talk about boolean attributes, which are special - but if you tested manually then I’m clearly wrong here :-)

MichaelDeBoey: `multiple={false}` and `multiple={undefined}` are not by definition the same as it's dependent on the actual implementation

But since it's a boolean value _and_ this is built-in JSX, it's indeed the same in this case

MichaelDeBoey: I would even say that `multiple="false"` or `multiple={null}` isn't allowed in (typed) React.

So only values that are **valid** would be no value (just `multiple`, which is the same as `multiple={true}`), `multiple={true}`, `multiple={false}` & `multiple={undefined}`

ljharb: right - but that means `multiple={false}` should have the same implicit role as `multiple={undefined}`?

MichaelDeBoey: @ljharb I would say so, yes

lb-: I guess we also need to work out if we want to write tests for things that 'should not' happen, I think it's better to write tests for cases like this.

However, I agree with the general sentiment that a lot of these cases being discussed are not valid TypeScript JSX and would be bad practice in normal JSX.

---

But, `multiple={false}` would **not** have the same same implicit role as `multiple={undefined}` @ljharb  @MichaelDeBoey 

* `multiple={false}` will render as `multiple="false"` which is treated in the browser as `multiple` (the boolean HTML attribute) being present and hence true.
* `multiple={undefined}` will render as ... nothing, the attribute never gets put into the DOM by React in this case, which means it will be treated by the browser as the `multiple` attribute not being there, hence false.

Unless different version of React work differently, `undefined` has always meant "do not render this prop as an attribute in the DOM", similar to how `null` is treated.

A reminder though that the testing framework used will treat `null` as a string `'null'` so it's harder to test for this outside of the lower level unit tests.

Screenshot 2024-10-27 at 10 54 52 AM

Screenshot 2024-10-27 at 10 56 08 AM

lb-: Ohhh K, I am wrong. My apologies.

I thought I had tested this but must have been testing other attributes, not `multiple`.

you are both right, @ljharb  @MichaelDeBoey  - `multiple={false}` will be like `multiple={undefined}` or `multiple={null}` for this attribute (and must be the same for other HTML Boolean attributes. 

Sorry for my confusion, I will update the unit tests accordingly.

Screenshot 2024-10-27 at 12 05 54 PM
Screenshot 2024-10-27 at 12 05 27 PM

MichaelDeBoey: > * `multiple={false}` will render as `multiple="false"` which is treated in the browser as `multiple` (the boolean HTML attribute) being present and hence true

This seems incorrect to me as that would mean that you can't conditionally make a `select` a multi-select if both passing `false` & `true` will result in the same thing 🤔

> `undefined` has always meant "do not render this prop as an attribute in the DOM"

Passing `undefined` is indeed the same as not passing the prop at all, but doesn't mean by definition it's not rendered in the DOM.  
Again: it's dependent on the implementation of the component.
For built-in JSX, it is indeed the same

> similar to how `null` is treated

`null` is a totally different story as that's seen as an actual value in React world, where `undefined` is the same as not passing the prop at all.  
The underlying implementation _could_ give you the exact same result though

Again strictly speaking you should not pass `null` for `multiple`, but in pure JS world (so not TS world) it's technically possible to pass whatever you want 🤷‍♂️

lb-: Thanks for your patience @ljharb  & @MichaelDeBoey  - I have learned some nuance that I thought I understood but was in fact more nuanced than before.

I have now simplified the code, updated unit tests and have hopefully got the correct behaviour tested now.

`multiple` being a HTML boolean attribute will only be output into the DOM as `multiple=""` in the cases where it's either explicitly set to `true` or simply set `multiple` which is the same as `true`. In almost all other cases (e.g. set to `false`, `""` empty string, `undefined` or `null` it will not add the `multiple` attribute to the output DOM.

Please take another review and let me know if this matches expectations on your end.

Again, thanks for helping me understand this better. Sorry it took a few goes.

src/util/implicitRoles/select.js
```diff
@@ -1,6 +1,21 @@
...
-  return 'listbox';
+export default function getImplicitRoleForSelect(attributes) {
+  const multiple = getProp(attributes, 'multiple');
+  if (multiple && getLiteralPropValue(multiple)) {
+    return 'listbox';
+  }
+
+  const size = getProp(attributes, 'size');
+  const sizeValue = size && getLiteralPropValue(size);
+  if (sizeValue && (Number(sizeValue) > 1)) {
```
ljharb, line 16:
```suggestion
  if (sizeValue > 1) {
```
fwiw i think this is equivalent and simpler

ljharb (approved):

PR #1029, opened 2024-10-25
[Breaking] update eslint version

MichaelDeBoey:
This way it's in line with what `typescript-eslint` is doing

https://typescript-eslint.io/users/dependency-versions/#eslint

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "auto-changelog": "^2.5.0",
     "babel-plugin-add-module-exports": "^1.0.4",
     "babel-preset-airbnb": "^5.0.0",
-    "eslint": "^8.8 || ^9.13",
+    "eslint": "^8.57.0 || ^9.13.0",
     "eslint-config-airbnb-base": "^15.0.0",
     "eslint-doc-generator": "^1.7.1",
     "eslint-plugin-eslint-plugin": "^4.3.0",
@@ -88,7 +88,7 @@
     "safe-regex-test": "^1.0.3"
   },
   "peerDependencies": {
-    "eslint": "^8.8 || ^9.13"
+    "eslint": "^8.57.0 || ^9.13.0"
   },
   "auto-changelog": {
     "output": "CHANGELOG.md",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -49,7 +49,7 @@
     "auto-changelog": "^2.5.0",
     "babel-plugin-add-module-exports": "^1.0.4",
     "babel-preset-airbnb": "^5.0.0",
-    "eslint": "^8.8 || ^9.13",
+    "eslint": "^8.57.0 || ^9.0.0",
```
ljharb, line 52: ah, the reason i stuck with 8.8 here is that my personal eslint config remains stuck on that eslint version, due to a "bug fix" that broke my primary use case.

Is there a reason we need to support eslint < 9.12?

MichaelDeBoey: Just wanted to keep this in line with latest `typescript-eslint`

ESLint v8.57.0 is the latest v8 that has all the flat config goodies, hence why they chose that one.
Nothing special happened between ESLint v9.0.0 and v9.13, so I don't see why we should keep v9.13 tbh

ljharb: i guess i see the opposite - if there's no reason to include 9.0 - 9.12, why bother?

My intention is to only support the latest version of 9 that's out at release time.

I suppose we could stick to 8.57, since long term I need to find a solution to the 8.8 thing anyways.

MichaelDeBoey: Supporting 9.0 comes at **no cost** _at all_, which would benefit more people updating this plugin imo

But if you want to keep 9.13, I'm happy to update the PR

ljharb: The cost of 9.0 is that we're not forcing anyone else to upgrade to 9.13 :-)

MichaelDeBoey: You're the one that's always preaching to have as much of BC as possible, but I'm fine with 9.13 as well 🤷‍♂️

ljharb: I preach *preserving* back compat, yes. but at breaking change time - which should be exceedingly rare - things should be as aggressive as possible, to leverage the opportunity.

ljharb (approved):

PR #1020, opened 2024-10-20
remove polyfills

v1rtl:
TODO:

- [x] array-includes
- [x] array.prototype.flatmap
- [x] object.fromentries
- [x] string.prototype.includes
- [x] hasown 

~this PR also bumps minimum Node version to 18 and ESLint to 8 (both in GHA and package.json)~

PR #1030, opened 2024-10-25
refactor: remove `es-iterator-helpers` dependency

MichaelDeBoey:


ljharb (changes requested):
This dep is still needed until node 20 is dropped.

MichaelDeBoey:
@ljharb Can you explain why?

`aria-query` returns normal arrays as of v5 for all `*.keys()` & `*.values()`, so we don't need iterator helpers and can just use built-in array functionality?
https://github.com/A11yance/aria-query/blob/main/CHANGELOG.md#500

ljharb:
aha, you're right. in that case, this PR should be able to go to `main` instead of `breaking`, yes?

MichaelDeBoey:
@ljharb Those array functions didn't exist on Node v4 yet, so we can't merge this PR to `main`

ljharb:
True. but on `main` we could replace the iterator helpers with array method packages?

MichaelDeBoey:
That would indeed be possible, but that would mean that we need to do 2 PRs: 1 on `main` to use array method packages + 1 on `breaking` to remove those packages as well and do what I did in this PR

ljharb:
True. It'd be nice to have both tho :-) also likely the next time i rebase `breaking` i'll be removing all the relevant array dep usages you add anyways

MichaelDeBoey:
@ljharb #1020 is already removing all these array function package usages, so I'd rather keep working on `breaking` to move forward asap tbh

ljharb (approved):

PR #1032, opened 2024-10-25
[Refactor] use `messageId` instead of `message`

MichaelDeBoey:


ljharb (approved):
split it up into 2 commits, and also updated the tests

PR #1033, opened 2024-10-25
[Refactor] don't destructure builtins

MichaelDeBoey:
Introduced in a3c0ffa5, but these are the only places we do this.  
In all other places we use `Object.*`

### DIFF
--- a/__mocks__/genInteractives.js
+++ b/__mocks__/genInteractives.js
@@ -10,8 +10,6 @@ import JSXElementMock from './JSXElementMock';
 import type { JSXAttributeMockType } from './JSXAttributeMock';
 import type { JSXElementMockType } from './JSXElementMock';
 
-const { fromEntries } = Object;
-
 const domElements = dom.keys();
 const roleNames = roles.keys();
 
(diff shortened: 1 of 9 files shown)


ljharb:
I actually prefer to do it everywhere - iow, I think this is going the wrong direction.

MichaelDeBoey:
@ljharb This is always going to come from `Object` from now on, so it's unnecessary to destruct `Object` globally imo

ljharb:
It's not globally, it's at the top level module scope.

MichaelDeBoey:
@ljharb Yeah that's what I meant, sorry for the miswording.

My opinion is just: from this point onwards, these methods will always come from `Object`, so I don't see any point to destruct these methods.
People just seeing `entries(abc)` will ask themselves where `entries` is coming from and why `Object.entries` isn't used (they'll think it's probably a special implementation for some reason), only to find out it actually is used.
When making it explicit directly, people will need less cognitive load to understand where everything is coming from

ljharb:
That's true, and I suppose I've accepted the risk of environment mutation, so that's not really an argument that applies.

ljharb (approved):
Rebased to clean up remaining examples of this pattern.

PR #1034, opened 2024-10-25
refactor: don't spread arrays

MichaelDeBoey:
Same reason as mentioned in #1030: `*.keys()` & `*.entries()` is an array, so no need to spread them anymore

### DIFF
--- a/__mocks__/genInteractives.js
+++ b/__mocks__/genInteractives.js
@@ -12,8 +12,8 @@ import JSXElementMock from './JSXElementMock';
 import type { JSXAttributeMockType } from './JSXAttributeMock';
 import type { JSXElementMockType } from './JSXElementMock';
 
-const domElements = [...dom.keys()];
-const roleNames = [...roles.keys()];
+const domElements = dom.keys();
+const roleNames = roles.keys();
 
 const interactiveElementsMap = {
   a: [{ prop: 'href', value: '#' }],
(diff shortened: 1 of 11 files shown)


__mocks__/genInteractives.js
```diff
@@ -12,9 +12,6 @@ import type { JSXElementMockType } from './JSXElementMock';
 
 const { fromEntries } = Object;
 
-const domElements = [...dom.keys()];
-const roleNames = [...roles.keys()];
```
ljharb, line 16: let's still cache the result at module level in all non-test files, otherwise every call has to regenerate the array.

ljharb (changes requested):

ljharb:
I pulled this into the branch as 0298c54 - i think this could probably be closed?

(although what would be much better is rebasing it onto `main` so it can be landed there - if you don't feel like doing it, though, i'll do so)

MichaelDeBoey:
> although what would be much better is rebasing it onto `main` so it can be landed there

That could cause errors as `Object.*` is probably not fully supported in Node 4 yet 🤔

ljharb:
This PR is about not spreading *-query arrays, though, not about destructuring builtins.

MichaelDeBoey:
Oops, I mixed them up as I was rebasing both PRs 🙈

Rebasing this to `main` would be a pain to resolve conflicts, so I'd rather keep this at `breaking` tbh

ljharb:
No worries, I've done it :-) thanks!

ljharb (approved):

PR #1038, opened 2024-11-15
[Fix] fix unstable plugin reference with flat configs

bradzacher:
The previous use of spread meant that
```ts
import jsxA11y from 'eslint-plugin-jsx-a11y';

jsxA11y !== jsxA11y.flatConfigs.recommended.plugins['jsx-a11y']
```

This is a problem because if someone does something like this 
```js
import jsxA11y from 'eslint-plugin-jsx-a11y';

export default [
  { plugins: { 'jsx-a11y': jsxA11y } },
  jsxA11y.flatConfigs.recommended,
];
```
then ESLint will crash with the error `Config "jsx-a11y/recommended": Key "plugins": Cannot redefine plugin "jsx-a11y".`

This PR fixes that by using `Object.assign` to mutate the `jsxA11y` object and maintain referential equality.

### DIFF
--- a/__tests__/index-test.js
+++ b/__tests__/index-test.js
@@ -38,3 +38,12 @@ test('schemas', (t) => {
 
   t.end();
 });
+
+test('plugin referentially equal to prevent flat config issues', (t) => {
+  const keys = Object.keys(plugin.flatConfigs);
+  for (let i = 0; i < keys.length; i += 1) {
+    const config = plugin.flatConfigs[keys[i]];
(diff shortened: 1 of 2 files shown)


ljharb:
Can we add a test that covers this?

ljharb (approved):

sergio-milu:
hey @ljharb @bradzacher this was merged but never released? cant find it in published versions

thanks

ljharb:
yes, that's correct, it hasn't been released yet

PR #1042, opened 2024-12-24
feat(label-has-associated-control): add option for enforcing label's htmlFor matches control's id

### LINKED ISSUE #956: label-has-associated-control not checking an actual valid configuration
Apologies if I've misunderstood the rule or configured this wrong.

I want to ensure that every (custom) input element has an associated label, and it's configured correctly. So the following works out the box:

Given the following configuration:
```json
    'jsx-a11y/label-has-associated-control': [
      'warn',
      {
        labelComponents: ['FormLabel'],
        controlComponents: ['Input', 'Select']
      },
    ],
```

I expect this to fail, as no HtmlFor, which it fails.
```tsx
<>
   <FormLabel>Foo</FormLabel>
   <Input name="foo" />
</>
```

I expect this to fail, as the htmlFor doesn't link to anything, but it doesn't fail
```tsx
<>
   <FormLabel htmlFor="foo">Foo</FormLabel>
   <Input name="foo" />
</>
```

Then expect this to pass
```tsx
<>
   <FormLabel htmlFor="foo">Foo</FormLabel>
   <Input id="foo" name="foo" />
</>
```

Given that:
> This rule checks that any label tag (or an indicated custom component that will output a label tag) either (1) wraps an input element (or an indicated custom component that will output an input tag) or (2) has an htmlFor attribute and that the label tag has text content.

I feel like my expected behaviour is not the reality, but I wasn't quite sure. 

Should my expectation work, or am I just misunderstanding the library/rule? Thanks.

michaelfaith:
This change adds an option to the `label-has-associated-control` rule, enforcing that the label's htmlFor attribute matches the associated control's id attribute.  Previously, the only validation done on htmlFor was that it was on the label component and had text.  There was no attempt to cross-check that value against any attribute on the associated control.  Not, when the option is enabled, cases where they don't match will report.

I also took the opportunity to update the error messages so that each assert type gets an error message with verbiage specific to the assertion. (not sure if this should be called out as a separate feature in the changelog?).

Note: the current implementation only checks the first instance it finds of a child component that matches each control component type.  It assumes that there won't be any acceptable cases where a label would have multiple inputs nested beneath it.  Let me know if that assumption doesn't hold and i can refine that logic more.

Closes: #956

### DIFF
--- a/flow/eslint-jsx.js
+++ b/flow/eslint-jsx.js
@@ -3,9 +3,24 @@
  */
 import type {
   JSXAttribute,
+  JSXElement,
   JSXOpeningElement,
 } from 'ast-types-flow';
 
 export type ESLintJSXAttribute = {
   parent: JSXOpeningElement
 } & JSXAttribute;
+
+type JSXOpeningFragment = {
+  type: 'JSXOpeningFragment',
+};
+
+type JSXClosingFragment = {
+  type: 'JSXClosingFragment',
+};
+
+export type JSXFragment = JSXElement & {
+  type: 'JSXFragment',
+  openingFragment: JSXOpeningFragment,
+  closingFragment: JSXClosingFragment,
+};
--- a/__mocks__/JSXFragmentMock.js
+++ b/__mocks__/JSXFragmentMock.js
@@ -1,0 +1,17 @@
+/**
+ * @flow
+ */
+
+export type JSXFragmentMockType = {
+  type: 'JSXFragment',
+  children: Array<Node>,
+};
+
+export default function JSXFragmentMock(
+  children?: Array<Node> = [],
+): JSXFragmentMockType {
+  return {
+    type: 'JSXFragment',
+    children,
+  };
+}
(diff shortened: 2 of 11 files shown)


flow/eslint-jsx.js
```diff
@@ -3,9 +3,24 @@
...
+
+type JSXOpeningFragment = {
+  type: 'JSXOpeningFragment',
+};
+
+type JSXClosingFragment = {
+  type: 'JSXClosingFragment',
+};
+
+export type JSXFragment = JSXElement & {
```
michaelfaith, line 22: It would have been better to use `Omit` to remove `openingElement` and `closingElement` from `JSXElement`, but this version of flow doesn't have `Omit` (that was introduced in 0.211.0).

michaelfaith, line 22: The `ast-types-flow` doesn't include a type for `JSXFragment`

michaelfaith:
@ljharb I realized this is still pending review.  Is this feature still something you want to move forward with?

ljharb:
Yes, thanks for your patience.

PR #1043, opened 2024-12-25
[patch] `no-autofocus`: don't report error if autoFocus is set to false

### LINKED ISSUE #1014: `jsx-a11y/no-autofocus` should not warn for false
I know the docs https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/no-autofocus.md show fail as

```
<div autoFocus />
<div autoFocus="true" />
<div autoFocus="false" />
<div autoFocus={undefined} />
```

but I feel like this is bad linting for components outside of developer control. We use the prop on chakra ui components to force it to be false --which is proper a11y. 

Ideally a framework would have set it up properly but they didn't. The workaround to set no-autofocus to warning or to throw in comments to ignore each case. However, I think a better way forward would be if the plugin allowed for falsy values (maybe as an option) so it would not fail on `<Component autoFocus="false" />` or `<Component autoFocus={false} />`.

Comments work fine so completely understandable if this isn't the right mentality.

michaelfaith:
This change adjusts the error condition for `no-autofocus` to allow for manually disabling `autoFocus`.  Now, any usage of `autoFocus="false"` or `autoFocus={false}` will not report.

Closes: #1014


ljharb (approved):
Seems good

PR #1048, opened 2025-04-18
refactor: replace emoji-regex with emoji-regex-xs

Namchee:
## Overview

This pull request replaces `emoji-regex` with [`emoji-regex-xs`](https://github.com/slevithan/emoji-regex-xs) package that has 97% smaller size while maintaining its API and features.

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "axe-core": "^4.10.2",
     "axobject-query": "^4.1.0",
     "damerau-levenshtein": "^1.0.8",
-    "emoji-regex": "^9.2.2",
+    "emoji-regex-xs": "^1.0.0",
     "hasown": "^2.0.2",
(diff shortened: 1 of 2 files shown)


ljharb (changes requested):
It looks good at a glance, but using it is blocked on https://github.com/slevithan/emoji-regex-xs/issues/4

Namchee:
After reading through `emoji-regex-xs` again, I don't think it can replace `emoji-regex` in this repository. My bad.

slevithan:
@ljharb Can you confirm that you do not want to use it if it requires (7 years old) Node.js 10?

ljharb:
Unfortunately yes, currently the place I'd use it requires node 4 support.

slevithan:
No worries. Thanks for confirming.

Issue #1051, opened 2025-07-08
`no-static-element-interactions` doesn't allow custom attribute mapping


sarahvharris:
**Problem:**

Our project uses React Router which utilizes `to` instead of `href`. Because of this, `no-static-element-interactions` will fail on multiple jsx elements

```
import { Link } from 'react-router-dom';

// currently fails although `to` will render to `href`
<Link onClick={onClick} to="some/path">some content</Link>
```

This goes a step further when we have additional layers that eventually render down to a React Router link + `to` utilization:

```
import { MenuLink } from 'some/path'; // uses react router Link within

<MenuLink onClick={onClick} to="some/path">some content</MenuLink>
```

**Potential solution:**

Add custom attribute mapping to the `no-static-element-interactions` rule. For now I was thinking in the following format (this mimics options available for `jsx-a11y/no-redundant-roles`):

```
  'jsx-a11y/no-static-element-interactions': [
    'error',
    {
      a: {
        attributes: {
          href: ['to', 'href'],
        },
      },
    }
  ]
```

From some testing locally, I believe this solution will solve this restriction as well as open the door to allow custom options for other rules in the `checkIsInteractiveElement` utility. Please lmk your thoughts, hoping to contribute back unless you see issues with this solution 🙏

ljharb:
I think this is a duplicate of #409?

sarahvharris:
> I think this is a duplicate of [#409](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/409)?

Sorry I think this is a separate issue, I am able to not have `jsx-a11y/anchor-is-valid` fail with this configuration but the other rule will still fail

```
    'jsx-a11y/anchor-is-valid': [
      'error',
      {
        components: ['Link'],
        specialLink: ['to'],
      },
    ],
```

PR #1052, opened 2025-07-09
`no-static-element-interactions` custom attribute mapping

sarahvharris:
based on this issue: https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/1051

### Problem
The `no-static-element-interactions` does not support custom attribute mapping which limits elements that utilize things like React Router that uses `to` instead of `href` (for example)

### Changes made
- Adds support for to defined attributes per `components` global setting
- Adds new util, `getSettingsAttributes` that will look at the global setting to see if theres any mapping of custom attributes, otherwise return the attributes set on the native element

### API structure

Within the `settings` config we have three ways to implement `components`

1. custom component with a simple mapping (current state)

```
`jsx-a11y`: {
	components: {
		Button: 'button'
	}
}
```

2. custom component with attributes

```
`jsx-a11y`: {
	components: {
		Button: {
			component: 'button',
			attributes: {
				onClick: ['handleClick'] // Button custom component has a prop `handleClick` instead of `onClick`
			}
		}
	}
}
```


sarahvharris:
> I think rather than adding it to just this rule, we should consider adding it as a plugin-wide setting, that all rules use when available?

Yes, @myasonik and I were noodling on how we could utilize the global `attributes` for this use case (and others). We found this rule to be a wide scope compared to other rules but anticipate wanting to contribute more when we can. Do you think creating an API proposal would be a good next step so we can discuss?

ljharb:
That'd be great! I'm happy to discuss it on the linked issue, or here.

myasonik:
We were thinking that using the `attributes` config as is could be dangerous because wouldn't that mean that defining something like `to: href` would mean you can't have a `to` prop on any other component? 

For example, if you had something like `<ArrowIcon to="left" />` would get interpreted as `<ArrowIcon href="left" />` which would fail because it's not on an anchor (presumably). 

So now we're also changing the global attributes config which felt like a big change...

ljharb:
ah. I don't think it should be "this prop on any component", but "this prop on these specific non-HTML elements only".

myasonik:
So are you thinking this would supersede or complement the existing global `attributes` config?

I think we could come up with a better API but that becomes a (future) breaking change (can be just a deprecation for now).

Currently: 
```
{
  "settings": {
    "jsx-a11y": {
      "attributes": {
        "for": ["htmlFor", "for"]
      }
    }
  }
}
```

Could become: 
```
{
  "settings": {
    "jsx-a11y": {
      "customAttributeMappings": [
        { // new config for label-has-associated-control; attributes would be deprecated
          "elements": ["label"], // which elements this applies to
          "attributes": { // which attributes have alternative names
            "for": ["htmlFor", "for"]
          }
        },
      {
        "elements": ["a"],
        "attributes": {
          "href": ["href", "to"]
        }
      },
      { // hypothetical example of what this unlocks
        "elements": ["a", "button"], // multiple elements
        "attributes": { // with multiple attributes
          "onClick": ["onClick", "handleClick"],
          "onKeyDown": ["onKeyDown", "handleKeyDown"]
        }
      }]
    }
  }
}
```

That feels better IMO

myasonik:
@ljharb Wanted to bump this thread - any thoughts on what an API could look like? 

Our first proposal was too narrowly scoped to the rule, the second proposal was too broad and we're struggling to see a good middle ground. Happy to contribute the change with a little direction!

A hail marry idea: to the previous proposed API we add a `components` key to complement `elements` so that things could be scoped only to `<Link />` and not to `<a />`. 

{
  "settings": {
    "jsx-a11y": {
      "customAttributeMappings": [
        { // new config for label-has-associated-control; attributes would be deprecated
          "elements": ["label"], // which elements this applies to
          "attributes": { // which attributes have alternative names
            "for": ["htmlFor", "for"]
          }
        },
      {
        "components": ["Link"],
        "attributes": {
          "href": ["href", "to"]
        }
      }] // hypothetically you could mix and match these
    }
  }
}

sarahvharris:
@ljharb I really like your second solution. I can take some time looking into the lift and report back here with progress/ code updates

docs/rules/no-static-element-interactions.md
```diff
@@ -21,24 +21,25 @@ Indicate the element's role with the `role` attribute:
...
+1. `link`
+1. `checkbox`
+1. `menuitem`
+1. `menuitemcheckbox`
+1. `menuitemradio`
+1. `option`
+1. `radio`
+1. `searchbox`
+1. `switch`
+1. `textbox`
```
ljharb, line 42: let's revert these unrelated formatting changes (i smell prettier)

docs/rules/no-static-element-interactions.md
```diff
@@ -47,12 +48,16 @@ Note: Adding a role to your element does **not** add behavior. When a semantic H
...
+}
+{
+  /* eslint-disable-next-line jsx-a11y/no-static-element-interactions */
+}
 <div onClick={this.handleButtonClick}>
   <button>Save</button>
   <button>Cancel</button>
-</div>
+</div>;
 ```
```
ljharb, line 61: these too

src/util/getAttributes.js
```diff
@@ -0,0 +1,50 @@
...
+  const { settings } = context;
+  const { attributes } = node;
+  const components = settings['jsx-a11y']?.components;
+
+  if (!components || !type) {
+    return attributes;
+  }
+
+  const componentConfig = Object.entries(components).find(([, config]) => (
+    config && typeof config === 'object' && config.component === type
```
ljharb, line 13:
```suggestion
    config
    && typeof config === 'object'
    && config.component === type
```

src/util/getAttributes.js
```diff
@@ -0,0 +1,50 @@
...
+
+  if (!attributeMap || typeof attributeMap !== 'object') {
+    return attributes;
+  }
+
+  const mappedAttributes = [...attributes];
+
+  Object.entries(attributeMap).forEach(([originalAttr, mappedAttrs]) => {
+    if (Array.isArray(mappedAttrs)) {
+      mappedAttrs.forEach((mappedAttr) => {
```
ljharb, line 31: instead of nested forEach's and mutation, can we use a more functional approach here? perhaps with flatMap?

ljharb (commented):
I'm not sure if we have existing tests for an invalid `components` setting, but either way it'd be good to add them to cover this PR :-)

__tests__/src/rules/no-static-element-interactions-test.js
```diff
@@ -34,6 +34,22 @@ const componentsSettings = {
     components: {
```
sarahvharris, line 34: FYI need to add tests for `getAttributes`

docs/rules/no-static-element-interactions.md
sarahvharris: reverted!

src/util/getSettingsAttributes.js
```diff
@@ -0,0 +1,68 @@
...
+  const jsxElementName = elementType(node);
+  const componentConfig = components[jsxElementName];
+
+  const { attributes: settingsAttributes } = typeof componentConfig === 'object' ? componentConfig : {};
+
+  if (!settingsAttributes || typeof settingsAttributes !== 'object') return rawAttributes;
+
+  const mappedRawAttrNames = new Set();
+
+  const mappedAttributes = Object.entries(settingsAttributes).flatMap(([originalAttr, mappedAttrs]) => {
```
sarahvharris, line 42: Object.entires and flatMap failing on older node versions - should I be pulling from `object.*`?

sarahvharris:
Hi @ljharb just bumping this for a review and hopefully assistance on the `Object.entires` and `flatMap` failures for older major versions 🙏

PR #1080, opened 2026-02-24
fix: update minimatch version to 3.1.5

nilzona:
The npm dependency `minimatch` [has a reported](https://security.snyk.io/package/npm/minimatch) "high vulnerability" which has been fixed in version 10.2.1 (and 10.2.2 of course).

This PR updates `minimatch` version to the latest (10.2.2)

UPDATE!

minimatch@3.1.5 has fixed the vulnerability, so updating to that one instead


nilzona:
ok, so obviously the new version of minimatch is incompatible with node versions <= 13 

Will this library support these old node versions forever?

UPDATE!

This PR has been updated to only update to the patched version within semver range (3.1.5), so the above comment is not relevant for this PR.

ljharb:
Updating intermediate dependencies to fix vulnerabilities is never necessary. In this specific case, the CVE fixes have been backported to minimatch v3, so all you have to do is update your own lockfiles. No changes are needed here or anywhere else.

PR #1089, opened 2026-05-04
[patch] lang: add htmlHasLang option

### LINKED ISSUE #1087: The document of `lang` differs from the implementation.
[`lang`](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/lang.md) rule checks that a `lang` attribute on the `<html>` is a valid IETF's BCP 47 language tag and doesn't check whether `<html>` has `lang` attribute or not, but the document says `<html>` fails for this rule.

[`html-has-lang`](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/main/docs/rules/html-has-lang.md) document says:
>This rule is largely superseded by the `lang` rule

so should we fix the implementation of `lang` rule to check `html-has-lang` rule in addition to `lang` validation? or should we remove the example of `<html>` from `lang` rule document?

mehm8128:
close https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/issues/1087

~~I worked two things, and this is a breaking change because `lang` rule will do additional checks once this is released.
The latter is not mentioned on the above issue, but I want to introduce this change (if needed, divide changes into another PR).~~

- add option to whether to check `lang` prop for `<html>` element
- ~~check other HTML elements other than `<html>` whether its `lang` prop is valid or not~~

### DIFF
--- a/__tests__/src/rules/lang-test.js
+++ b/__tests__/src/rules/lang-test.js
@@ -52,6 +52,7 @@ ruleTester.run('lang', rule, {
     { code: '<HTML lang="foo" />' },
     { code: '<Foo lang={undefined} />' },
     { code: '<html lang={undefined} />' },
+    { code: '<html lang={foo} />', options: [{ htmlHasLang: true }] },
     { code: '<Foo lang={undefined} />', settings: componentsSettings },
     { code: '<Foo lang="en" />', settings: componentsSettings },
     { code: '<Box as="html" lang="en"  />', settings: componentsSettings },
@@ -63,6 +64,10 @@ ruleTester.run('lang', rule, {
     { code: '<html lang="zz-LL" />', errors: [expectedError] },
     { code: '<Box as="html" lang="foo" />', settings: componentsSettings, errors: [expectedError] },
     { code: '<html />', options: [{ htmlHasLang: true }], errors: [expectedHtmlHasLangError] },
+    { code: '<html lang={undefined} />', options: [{ htmlHasLang: true }], errors: [expectedHtmlHasLangError] },
+    {
+      code: '<Foo lang={undefined} />', settings: componentsSettings, options: [{ htmlHasLang: true }], errors: [expectedHtmlHasLangError],
+    },
     {
       code: '<Foo />', settings: componentsSettings, options: [{ htmlHasLang: true }], errors: [expectedHtmlHasLangError],
     },
--- a/__tests__/src/rules/lang-test.js
+++ b/__tests__/src/rules/lang-test.js
@@ -7,10 +7,10 @@
 // Requirements
 // -----------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


ljharb:
We definitely want to avoid any and all breaking changes; only the option should alter behavior.

__tests__/src/rules/lang-test.js
```diff
@@ -46,14 +51,21 @@ ruleTester.run('lang', rule, {
     { code: '<html lang={foo} />' },
     { code: '<HTML lang="foo" />' },
     { code: '<Foo lang={undefined} />' },
+    { code: '<html lang={undefined} />' },
+    { code: '<Foo lang={undefined} />', settings: componentsSettings },
```
mehm8128, line 56: `lang={undefined}` is dealt with as `lang` is not provided, because in these both cases the inner value of `lang` is `undefined`.

mehm8128:
@ljharb 
Sorry, I fixed to include only changes related to `htmlHasLang` option.

__tests__/src/rules/lang-test.js
ljharb: let's be sure to duplicate the exact source code test cases added to valid, in invalid (and vice versa)

docs/rules/lang.md
```diff
@@ -3,23 +3,39 @@
...
 This rule takes no arguments.
 
+## Rule options
+
+This rule takes one optional boolean that determines whether to check for the `lang` prop on `<html>` elements. It defaults to `false`:
+
+```json
+{
+    "rules": {
+        "jsx-a11y/lang": [ 2, {
```
ljharb:
```suggestion
        "jsx-a11y/lang": [2, {
```

src/rules/lang.js
```diff
@@ -25,45 +34,44 @@ export default {
...
     return {
-      JSXAttribute: (node) => {
-        const name = propName(node);
-        if (name && name.toUpperCase() !== 'LANG') {
-          return;
-        }
+      JSXOpeningElement: (node: JSXOpeningElement) => {
+        const checkHtmlHasLang = context.options[0]?.htmlHasLang ?? false;
+        const langProp = getProp(node.attributes, 'lang');
+        const langPropVal = getLiteralPropValue(langProp);
```
ljharb:
```suggestion
        const value = getLiteralPropValue(langProp);
```
to minimize the diff

ljharb (changes requested):

__tests__/src/rules/lang-test.js
mehm8128: Did you mean "add invalid test cases of `<html lang={undefined} />` and `<Foo lang={undefined} />` with `htmlHasLang: true`", right?

